PR #150, opened 2017-01-19
Adding attention layers

matt-gardner:
This PR does two things:

1. It moves `common.tensors` to its own top-level module, `tensors`, and adds some structure to the module.

2. It adds a `tensors.similarity_functions` module, which allows for configurable similarity computations, which form the core of an `Attention` layer.  Along with this, it adds a `layers.attention` module, with a couple of `Attention` layers, and merges what was `layers.softmaxes` with it.


src/main/python/deep_qa/layers/attention/attention.py
```diff
@@ -49,6 +49,7 @@ def call(self, inputs, mask=None):
         vector, matrix = inputs
         matrix_mask = mask[1]
         num_rows = K.int_shape(matrix)[1]
+        print("Matrix shape:", K.int_shape(matrix))
```
nelson-liu: you forgot to remove this after :)

nelson-liu: nevermind, i fixed it in https://github.com/allenai/deep_qa/pull/153/commits/e6666a2f5d44e96c9ddd8fe51020584818081098

PR #153, opened 2017-01-24
Add experiment for AS Reader + more tests / numerical stability improvements

nelson-liu:
Don't merge this yet, at least not until I confirm i can reproduce the paper results for the attention sum reader on Who Did What. In addition, this PR depends on #152 so look at that first :)

This PR:
- adds an example experiment for the attention sum reader
- adds the slew of tests that I wrote while trying to debug the `NaN` values
- includes various (purported) numerical stability improvements
- lets the user set the encoder / seq2seq encoder output dimension as a parameter in the config file. if no value is specified, it falls back to the embedding dimension (which was previously the unconditional default).


src/test/python/deep_qa_test/layers/attention/attention_test.py
```diff
@@ -1,13 +1,92 @@
...
+
+        result = model.predict([vector, matrix])
+        assert_almost_equal(result, np.array([[0.5, 0.5]]))
+
+    def test_masked(self):
+        # Testing general masked non-batched case.
+        vector = K.variable(np.array([[0.3, 0.1, 0.5]]))
+        matrix = K.variable(np.array([[[0.6, 0.8, 0.1], [0.15, 0.5, 0.2], [0.1, 0.4, 0.3]]]))
+        mask = K.variable(np.array([[1.0, 0.0, 1.0]]))
+        result = K.eval(Attention().call([vector, matrix], mask=["_", mask]))
```
nelson-liu: im not sure this is a good idea actually, maybe keras does some stuff internally that we miss out on when we just test __call__ directly. At least it was the case for me (with the whole not a keras tensor thing) that these tests would pass, but the others that actually used a model and the full framework would fail which seems like they do something abstracted away from the user. Besides, having them in the whole model scaffolding is more accurate of their actual use, right? The reason i did this is because the situation with masking layer only masking if the element at the final dimension is the same value, and since we have arrays and they take floats it isn't really compatible...

matt-gardner: Hmm, ok.  If there are failures that doing it this way won't catch, that's a good argument in favor of doing it the other way.

src/main/python/deep_qa/layers/entailment_models/word_alignment.py
```diff
@@ -55,6 +55,7 @@ def _align(source_embedding, target_embedding, source_mask, target_mask, normali
                 # (batch_size, source_length, target_length)
                 s2t_mask = K.expand_dims(float_source_mask, dim=-1) * K.expand_dims(float_target_mask, dim=1)
                 flattened_s2t_mask = last_dim_flatten(s2t_mask)
+                # pylint: disable=redefined-variable-type
```
nelson-liu: this is possible, but it's not transparent to me which pylint is used since it uses a cached version...

src/main/python/deep_qa/layers/encoders/shareable_gru.py
```diff
@@ -24,5 +24,9 @@ def call(self, x, mask=None, **kwargs):
             result_shape = K.int_shape(res)
             if input_shape[1] is not None and result_shape[1] is None:
                 shape = [d if d is not None else -1 for d in input_shape]
+                if self.output_dim != input_shape[-1]:
```
nelson-liu: i think we need the check on line 25 still because result shape isn't guaranteed to be 3 dimensional (`return_sequence=False`)

nelson-liu: oops was looking at the wrong conditional, yeah that seems unnecessary now.

src/test/python/deep_qa_test/layers/attention/attention_test.py
```diff
@@ -1,13 +1,92 @@
 # pylint: disable=no-self-use,invalid-name
 
-import numpy
+import numpy as np
```
nelson-liu: haha that's fine, i definitely understand. i should probably go through and retroactively fix the places i've used `np`, should be a quick grep away.

src/test/python/deep_qa_test/layers/attention/attention_test.py
nelson-liu: a masking layer that lets us mask inputs that aren't just single values would be very useful, we should consider writing one...

PR #155, opened 2017-01-26
Added buildable sphinx docs

nelson-liu:
This PR adds HTML docs to the project. They have to be tweaked a bit after #150 goes in, but I'll be doing that shortly. ping me on slack if you want to see a preview, or just pull down this branch, install the new requirements, navigate to `deep_qa/src/main/python/deep_qa/doc/` and then run `make html`. Then, youll find the built html files in `doc/built_`.

todo in future PRs: add more narrative, user-guide-esque docs and fix the mangled docstrings and add more docstrings.

PR #156, opened 2017-01-26
Added buildable sphinx docs

nelson-liu:
This is the same as #155 , but with updates for the changes in #150 

This PR adds HTML docs to the project. They have to be tweaked a bit after #150 goes in, but I'll be doing that shortly. ping me on slack if you want to see a preview, or just pull down this branch, install the new requirements, navigate to deep_qa/src/main/python/deep_qa/doc/ and then run make html. Then, youll find the built html files in doc/built_.

todo in future PRs: add more narrative, user-guide-esque docs and fix the mangled docstrings and add more docstrings.


.gitignore
```diff
@@ -17,3 +17,5 @@ tmp*/
 *.make
 *.temp
 .coverage
+_build
+.DS_Store
```
matt-gardner: The `.DS_Store` line should go in your global `.gitignore`, not in here.  It's not specific to our repo, it's specific to your dev environment.

nelson-liu: good point, fixed.

nelson-liu:
> One question, though - what do you think of storing the files under a top-level doc/ directory, instead of under src/main/python/deep_qa/doc/? Or src/doc/. src/main/ is where you should put code for running whatever's in this, src/test/ is test code for what's in src/main/, and neither one seems like a great fit for documentation.

So the reason i put it in here was because this is pretty much documentation for the python package, so i thought that it deserves to go within the package itself... i like the idea of putting it in the toplevel doc directory, but i wanted to show that these docs were for the python module. Though i guess i do want this to also have a user guide, etc which is more general than the python module....

nelson-liu:
merging this, will update when i get it online.

PR #157, opened 2017-01-26
Add option for setting "preferred" backend

nelson-liu:
This PR lets you specify a preferred backend in the json, and a warning will be displayed if the experiment is run with a different backend (but it'll still run).

this is motivated by my recent experience with the attention sum reader --- the theano model trains quite well and gets around 0.55 keras val accuracy (comparable to published result), but the tensorflow backend doesn't seem to learn anything and just stagnates at 0.38 accuracy. Thus, as the writer of the experiment i'd like to strongly using theano to run it, and i would set this parameter.

### DIFF
--- a/src/main/python/deep_qa/training/trainer.py
+++ b/src/main/python/deep_qa/training/trainer.py
@@ -45,9 +45,23 @@ class Trainer:
         # Preferred backend to use for training. If a different backend is detected, we still
         # train but we also warn the user.
         self.preferred_backend = params.pop('preferred_backend', None)
-        if self.preferred_backend and self.preferred_backend != K.backend():
-            logger.warning("Preferred backend is %s, but current backend is %s.",
-                           self.preferred_backend, K.backend())
+        if self.preferred_backend and self.preferred_backend.lower() != K.backend():
+            warning_info = ("@ Preferred backend is %s, but "
+                            "current backend is %s. @" % (self.preferred_backend.lower(),
+                                                          K.backend()))
+            end_row = "@" * len(warning_info)
+            warning_row_spaces = len(warning_info) - len("@ WARNING: @")
+            left_warning_row_spaces = right_warning_row_spaces = warning_row_spaces // 2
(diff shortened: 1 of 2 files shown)


src/main/python/deep_qa/training/trainer.py
```diff
@@ -42,6 +42,13 @@ def __init__(self, params: Dict[str, Any]):
             parent_directory = os.path.dirname(self.model_prefix)
             os.makedirs(parent_directory, exist_ok=True)
 
+        # Preferred backend to use for training. If a different backend is detected, we still
+        # train but we also warn the user.
+        self.preferred_backend = params.pop('preferred_backend', None)
+        if self.preferred_backend and self.preferred_backend != K.backend():
+            logger.warning("Preferred backend is %s, but current backend is %s.",
```
matt-gardner: Is it worth waiting here for user input?  If not I would probably take several lines to display this warning, making it really bold and obvious.

Also, should you update your example experiment with the ASReader to include this parameter?

nelson-liu:
@matt-gardner done, here's an image of what the new warning looks like (thanks, ssh!)

screen shot 2017-01-26 at 5 33 26 pm

src/main/python/deep_qa/training/trainer.py
```diff
@@ -42,6 +42,27 @@ def __init__(self, params: Dict[str, Any]):
             parent_directory = os.path.dirname(self.model_prefix)
             os.makedirs(parent_directory, exist_ok=True)
 
+        # Preferred backend to use for training. If a different backend is detected, we still
+        # train but we also warn the user.
+        self.preferred_backend = params.pop('preferred_backend', None)
+        if self.preferred_backend and self.preferred_backend.lower() != K.backend():
+            warning_info = ("@ Preferred backend is %s, but "
```
matt-gardner: One last nitpick - this is a lot of logic to be sitting in `__init__`.  Can you move it to a function like `_print_backend_warning()`?

PR #159, opened 2017-01-26
Added WeightedSum layer

matt-gardner:
I'm not sure what needs to change in `doc/` to account for this new layer - @nelson-liu, can you point me to that so that I can add it to the PR?


nelson-liu:
so to update the docs to reflect your new layer, you're going to broadly have to do the following:
1. update the API docs
2. update the sidebar to include this layer

for 1, just edit [doc/api_doc/deep_qa.layers.attention.rst](https://github.com/allenai/deep_qa/blob/master/doc/api_doc/deep_qa.layers.attention.rst). if you follow the same format, you should be good. The ordering doesn't matter within the document.

For 2, you're going to want to edit [doc/layers/attention.rst](https://github.com/allenai/deep_qa/blob/master/doc/layers/attention.rst) in the same fashion. note that the formatting between the two is slightly different.

to build the docs locally, you're going to want to get the new requirements in `requirements.txt` and then navigate to `doc/` and then run `make html`. the output should be in `_build`, and you can go in there and start a python webserver to view it locally.

doc/models/text_classification.rst
```diff
@@ -9,7 +9,7 @@ Input: Text sequence
 Output: Class label
 
 
-tree_lstm_model
+TreeLstmModel
```
nelson-liu: based on your changes https://github.com/allenai/deep_qa/pull/159/files#diff-a095eea3e0a97386695f6b99083d0bbbR22, maybe keep it consistent and have this be `TreeLSTMModel`?

src/main/python/deep_qa/layers/attention/weighted_sum.py
```diff
@@ -0,0 +1,91 @@
...
+    @staticmethod
+    def _expand_matrix_if_necessary(matrix, attention_vector):
+        try:
+            matrix_shape = K.int_shape(matrix)[:-1]  # taking off the embedding dimension here.
+        except Exception:  # pylint: disable=broad-except
+            if K.backend() == 'theano':
+                raise RuntimeError("Theano backend doesn't support K.int_shape very well - use "
+                                   "tensorflow instead for this model.")
+            else:
+                raise
```
nelson-liu: perhaps you meant to have something else here? What does the singular `raise` do in this context?

src/main/python/deep_qa/layers/attention/weighted_sum.py
```diff
@@ -0,0 +1,91 @@
...
+        attention_shape = K.int_shape(attention_vector)
+        if matrix_shape != attention_shape:
+            # We'll take care of the batch size first.  After this, the matrix_shape should match
+            # the end of the attention_shape exactly.
+            assert matrix_shape[0] == attention_shape[0], "somehow batch sizes don't match"
```
nelson-liu: maybe have this message me a bit more descriptive?

src/main/python/deep_qa/layers/attention/weighted_sum.py
```diff
@@ -0,0 +1,91 @@
...
+            matrix_mask = K.expand_dims(matrix_mask, dim=-1)
+            matrix_mask = self._expand_matrix_if_necessary(matrix_mask, attention_vector)
+            matrix = switch(matrix_mask, matrix, K.zeros_like(matrix))
+        return K.sum(K.expand_dims(attention_vector, dim=-1) * matrix, -2)
+
+    @staticmethod
+    def _expand_matrix_if_necessary(matrix, attention_vector):
+        try:
+            matrix_shape = K.int_shape(matrix)[:-1]  # taking off the embedding dimension here.
+        except Exception:  # pylint: disable=broad-except
```
nelson-liu: i'm actually surprised that this doesn't work for theano --- what's the issue like?

nelson-liu: this seems similar to https://github.com/allenai/deep_qa/blob/52a950685bae2673d05272b544d647dce0ded8ce/src/main/python/deep_qa/layers/option_attention_sum.py#L90

matt-gardner: It's because this gets called on the mask, and the mask isn't made into a keras tensor.  Maybe we could fix this by doing whatever keras does to make things into keras tensors in the first place, and we could probably find out by tracing what happens to the output of `call()`.

src/main/python/deep_qa/layers/attention/weighted_sum.py
matt-gardner: A bare `raise` in python just means "whatever exception you had previously, raise it again".

src/main/python/deep_qa/layers/attention/weighted_sum.py
matt-gardner: What else would you say?

src/main/python/deep_qa/layers/attention/weighted_sum.py
matt-gardner: I spent a bit of time trying to figure out if we could make the mask into a keras tensor, and it did not look feasible.

doc/models/text_classification.rst
matt-gardner: Hmm, you're right.  Really, it should just be `Lstm` everywhere, but I'll change this one to match for now, instead of changing everywhere else (including the code).

matt-gardner:
Does this look ok now?

src/main/python/deep_qa/layers/attention/weighted_sum.py
nelson-liu: i'm actually not too sure, it frankly looked a bit like a debug message that you used in testing to me when I first saw it.

nelson-liu:
yes, thanks!

PR #161, opened 2017-01-26
Fixed the label in CharacterSpanInstance

matt-gardner:


matt-gardner:
The span indices never got converted to one-hot vectors previously, but they need to be in order to use them to train.  That's all this is doing.

PR #164, opened 2017-01-26
modify argument ordering in bi_gru encoder

nelson-liu:
So readthedocs runs python 3.4, which doesn't like having kwargs before positional args. This is "fixed" in 3.5. This PR makes the ordering backwards compatible

example (on 3.4, both work in 3.5):
```python
In [1]: dict(y=2, **{'x': 1})
Out[1]: {'x': 1, 'y': 2}

In [2]: dict(**{'x': 1}, y=2)
  File "<ipython-input-2-5b28e96ee1a8>", line 1
    dict(**{'x': 1}, y=2)
                   ^
SyntaxError: invalid syntax
```


nelson-liu:
merging this while i can ;)

PR #165, opened 2017-01-26
Adding .travis.yml

matt-gardner:


.travis.yml
```diff
@@ -0,0 +1,25 @@
+sudo: required
```
nelson-liu: i don't think we need `sudo: required`? Note [the benefits of using `sudo: false`](https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments)

nelson-liu: ah but that would require using `dist: precise`....

matt-gardner: I was just resurrecting the old config to see if this would work at all.  As we've seen, the answer looks like it's "no".

PR #167, opened 2017-01-27
Set up TravisCI

nelson-liu:
Here's a set of edits to `.travis.yml` to make it work for testing our python code. not sure if we should switch, but i figured i'd make this PR just so it's available if we wanted to do it.


build_tools/travis/install.sh
```diff
@@ -0,0 +1,37 @@
...
+# Deactivate the travis-provided virtual environment and setup a
+# conda-based environment instead
+deactivate
+
+# Use the miniconda installer for setup of conda itself
+pushd .
+cd
+mkdir -p download
+cd download
+if [[ ! -d /home/travis/miniconda3 ]]
```
matt-gardner: Given that this is now just a python project, do we even need to worry about this?  Should travis already know how to give us the right python environment for python 3.5?

nelson-liu: I'd prefer to explicitly have it set up conda, so we have a unified  environment across dev + CI.

PR #169, opened 2017-01-27
Added some simple Layers

matt-gardner:
Need to add documentation for these, still.


src/test/python/deep_qa_test/layers/complex_concat_test.py
```diff
@@ -48,4 +48,4 @@ def test_call_handles_complex_combinations(self):
                 input_4_tensor + input_3_tensor,
                 input_3_tensor
                 ], axis=1)
-        numpy.testing.assert_almost_equal(concat_tensor, expected_tensor, decimal=4)
+        numpy.testing.assert_almost_equal(concat_tensor, expected_tensor, decimal=3)
```
nelson-liu: why is there such low precision?

matt-gardner: I'm pretty sure the reason is the division by the output of `numpy.random.rand()`, which could be really small.

src/main/python/deep_qa/layers/backend/max.py
```diff
@@ -0,0 +1,41 @@
+from keras import backend as K
+from keras.layers import Layer
+from overrides import overrides
+
+from ...tensors.backend import switch, very_negative_like
+
+
+class Max(Layer):
+    '''
```
nelson-liu: i'm going to start being a lot more hard-nosed about docstring formatting since we actually have docs to maintain now --- [PEP257](https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring) specifies that these should be triple quotes.

src/main/python/deep_qa/layers/backend/max.py
```diff
@@ -0,0 +1,41 @@
...
+    This Layer performs a max over some dimension.  Keras has a similar layer called
+    GlobalMaxPooling1D, but it is not as configurable as this one, and it does not support masking.
```
nelson-liu: double backticks around `GlobalMaxPooling1D`

src/main/python/deep_qa/layers/backend/max.py
```diff
@@ -0,0 +1,41 @@
+from keras import backend as K
+from keras.layers import Layer
+from overrides import overrides
+
+from ...tensors.backend import switch, very_negative_like
+
+
+class Max(Layer):
+    '''
+    This Layer performs a max over some dimension.  Keras has a similar layer called
```
nelson-liu: if "Layer" is capitalized, i feel like you're referring to the object --- in that case, I think this should be in double backticks. If you're just referring to the concept of a layer in general, i think lowercase would be better.

src/main/python/deep_qa/layers/backend/max.py
```diff
@@ -0,0 +1,41 @@
...
+    GlobalMaxPooling1D, but it is not as configurable as this one, and it does not support masking.
+
+    If the mask is not None, it must be the same shape as the input.
```
nelson-liu: double backticks around none

src/main/python/deep_qa/layers/backend/permute.py
```diff
@@ -0,0 +1,35 @@
+from typing import Tuple
+
+from keras import backend as K
+from keras.layers import Layer
+from overrides import overrides
+
+
+class Permute(Layer):
+    '''
+    This Layer calls K.permute_dimensions on both the input and the mask.
```
nelson-liu: double backticks on `Layer` (if applicable, see comments above) and `K.permute_dimensions`

src/main/python/deep_qa/layers/backend/permute.py
```diff
@@ -0,0 +1,35 @@
...
+
+    If the mask is not None, it must have the same shape as the input.
```
nelson-liu: backticks around none

src/main/python/deep_qa/layers/backend/repeat.py
```diff
@@ -0,0 +1,36 @@
+from keras import backend as K
+from keras.layers import Layer
+from overrides import overrides
+
+
+class Repeat(Layer):
+    '''
+    This Layer calls K.repeat_elements on both the input and the mask, after calling K.expand_dims.
```
nelson-liu: backticks on Layer and the functions

src/main/python/deep_qa/layers/backend/repeat.py
```diff
@@ -0,0 +1,36 @@
...
+
+    If the mask is not None, we must be able to call K.expand_dims using the same axis parameter as
```
nelson-liu: here as well

src/main/python/deep_qa/layers/backend/repeat.py
```diff
@@ -0,0 +1,36 @@
...
+    we do for the input.
+
+    Input: A tensor of arbitrary shape.
+    Output The input tensor repeated along one of the dimensions.
```
nelson-liu: `:` after `Output`

src/main/python/deep_qa/layers/highway.py
```diff
@@ -0,0 +1,11 @@
+from keras.layers import Highway as KerasHighway
+from overrides import overrides
+
+class Highway(KerasHighway):
+    """
+    Keras's Highway layer does not support masking, but it easily could, just by returning the
```
nelson-liu: `Keras's` -> `Keras'`? I don't really mind either way. Backticks around Highway.

src/main/python/deep_qa/layers/highway.py
```diff
@@ -0,0 +1,10 @@
+from keras.layers import Highway as KerasHighway
+
+class Highway(KerasHighway):
+    """
+    Keras's Highway layer does not support masking, but it easily could, just by returning the
```
nelson-liu: `Keras's` -> `Keras'`? I don't quite mind either way. Backticks on `Highway`.
What do you think about naming this `MaskedHighway?`

src/main/python/deep_qa/layers/backend/repeat.py
nelson-liu: This also occurs in other files, missed it the first time sorry.

src/main/python/deep_qa/layers/backend/max.py
matt-gardner: Thanks.  I'm not very familiar with how docstrings should be formatted in order to get the best results in our docs, so pointers like this are really helpful.

matt-gardner: Done.

src/main/python/deep_qa/layers/backend/max.py
matt-gardner: Done.

src/main/python/deep_qa/layers/backend/max.py
matt-gardner: Done.

src/main/python/deep_qa/layers/backend/max.py
matt-gardner: Done.

src/main/python/deep_qa/layers/backend/permute.py
matt-gardner: Done.

src/main/python/deep_qa/layers/backend/permute.py
matt-gardner: Done.

src/main/python/deep_qa/layers/backend/repeat.py
matt-gardner: Done.

src/main/python/deep_qa/layers/backend/repeat.py
matt-gardner: Done.

src/main/python/deep_qa/layers/backend/repeat.py
matt-gardner: Done.

src/main/python/deep_qa/layers/highway.py
matt-gardner: I'm actually thinking of going the other way, and renaming `MaskedSoftmax` to `Softmax`.  I guess my thought is, if we're fixing a bug in Keras, we'll use the same names, and maybe submit a PR to Keras, so that when if/when it gets fixed in Keras, we just change an import, and we're done.

matt-gardner:
Ok, this should be good to take another look.

src/main/python/deep_qa/layers/backend/repeat.py
```diff
@@ -0,0 +1,39 @@
...
+class Repeat(Layer):
+    """
+    This `Layer` calls `K.repeat_elements` on both the input and the mask, after calling
+    `K.expand_dims`.
+
+    If the mask is not `None`, we must be able to call `K.expand_dims` using the same axis
+    parameter as we do for the input.
+
+    Input:
+        A tensor of arbitrary shape.
```
nelson-liu: Maybe add bullets for each item? http://docutils.sourceforge.net/docs/user/rst/quickref.html#bullet-lists

PR #174, opened 2017-01-30
Add train logs for asreader_who_did_what experiment

nelson-liu:


matt-gardner:
Hmm, I would probably put this in a condensed form somehow on a new page under `doc/`, instead of committing the actual log file.  I might put it here: http://deep-qa.readthedocs.io/en/latest/models/about_models.html, though I'm not sure yet.

nelson-liu:
yeah, i'm not too sure where or how to present it as well --- we can leave this open until we think of a good solution.

matt-gardner:
I'd probably add a section under `About Models`  listing all of the models we can reproduce, including references to papers, links to the experiment json file / scala code to run the experiment, the result we got, and the running time (including on what machine).  My only question is whether to have this in the `About Models` page, or somewhere else (e.g., one section like this under each type of model).

nelson-liu:
still unsure where this should go...

nelson-liu:
i cant think of a centralized place to put these things, so i'm going to close this for now; the files will remain on my branch / these results are documented so I don't think we're losing much.

matt-gardner:
Maybe you could add a section to [this page](http://deep-qa.readthedocs.io/en/latest/models/about_models.html) that lists some of the more popular models that we have implemented (with links to the docs for the actual implementation), and gives results from training those models with our code.

doc/models/about_models.rst
```diff
@@ -20,3 +20,76 @@ We also define a few general Pretrainers in a submodule here. The Pretrainers in
...
+Attention Sum Reader
+--------------------
+
+The `Attention Sum Reader
+<https://www.semanticscholar.org/paper/Text-Understanding-with-the-Attention-Sum-Reader-Kadlec-Schmid/1023b20d226bd0af9fdf0fd1847accefbfa5ec84>`_
+Network is implemented in
+:mod:`~deep_qa.models.reading_comprehension.attention_sum_reader`.
+
+Here are the train logs::
+
```
matt-gardner: Is it possible to have a collapsible block in RST?  Looks like this might work: http://stackoverflow.com/a/25543713/1467533.

nelson-liu: hmm, I'm not sure we can do that since we use RTD's theme...

matt-gardner: Ok, don't bother spending time in your last two days worrying about this.  But can you add an issue to look into whether we can do this?

nelson-liu: yes, i also raised an issue on the RTD repo as a feature request

PR #175, opened 2017-01-30
Enable creating multiple encoders of one type

nelson-liu:
Before this PR, it was not possible to make two `*_encoders` (more concretely, it would not be possible to make two separate `sentence_encoders` that had different parameters). This PR attempts to change that by adopting a similar interface to how embedding layers are handled --- storing them in a dictionary where the key is the `"type"` of encoder.

@matt-gardner I have no idea if I did this properly, it's quite likely that the changes aren't far-reaching enough (i.e. I forgot to change something somewhere or consider some case).

This needs to work in order to implement the Gated Attention Reader.


src/main/python/deep_qa/training/text_trainer.py
```diff
@@ -94,9 +93,9 @@ def __init__(self, params: Dict[str, Any]):
         # don't want to set them now, because they use max length information that only gets set
         # after reading the training data.
         self.embedding_layers = {}
-        self.sentence_encoder_layer = None
-        self.word_encoder_layer = None
-        self.seq2seq_encoder_layer = None
+        self.sentence_encoder_layers = {}
+        self.word_encoder_layers = {}
```
matt-gardner: I would remove this field entirely, and change `self.sentence_encoder_layers` to `self.encoder_layers`.

src/main/python/deep_qa/training/text_trainer.py
```diff
@@ -183,15 +182,15 @@ def _load_layers(self):
...
-            elif layer.name == "timedist_sentence_encoder":
-                logger.info("  Found sentence encoder")
-                if self.sentence_encoder_layer is None:
-                    self.sentence_encoder_layer = layer
-                else:
-                    logger.warning("  FOUND DUPLICATE SENTENCE ENCODER LAYER!  NOT SURE WHAT TO DO!")
+            if '_sentence_encoder' in layer.name:
+                sentence_encoder_type = layer.name.replace("_sentence_encoder", "")
+                self.sentence_encoder_layers[sentence_encoder_type] = layer
+            if '_word_encoder' in layer.name:
```
matt-gardner: These lines can be removed.

src/main/python/deep_qa/training/text_trainer.py
```diff
@@ -183,15 +182,15 @@ def _load_layers(self):
...
-            elif layer.name == "sentence_encoder":
-                logger.info("  Found sentence encoder")
-                self.sentence_encoder_layer = layer
-            elif layer.name == "timedist_sentence_encoder":
-                logger.info("  Found sentence encoder")
-                if self.sentence_encoder_layer is None:
-                    self.sentence_encoder_layer = layer
-                else:
-                    logger.warning("  FOUND DUPLICATE SENTENCE ENCODER LAYER!  NOT SURE WHAT TO DO!")
+            if '_sentence_encoder' in layer.name:
```
matt-gardner: `s/sentence_encoder/encoder/`.

src/main/python/deep_qa/training/text_trainer.py
```diff
@@ -360,34 +359,36 @@ def _get_new_embedding(self, name: str, embedding_size: int, vocab_name: str='wo
                                                name=name + '_projection')
         return embedding_layer, projection_layer
 
-    def _get_sentence_encoder(self):
+    def _get_sentence_encoder(self, encoder_type="default"):
```
matt-gardner: `s/_get_sentence_encoder/_get_encoder/`

matt-gardner: And I would probably just use `name` instead of `encoder_type`.

src/main/python/deep_qa/training/text_trainer.py
```diff
@@ -360,34 +359,36 @@ def _get_new_embedding(self, name: str, embedding_size: int, vocab_name: str='wo
...
-    def _get_new_sentence_encoder(self, name="sentence_encoder"):
-        # The code that follows would be destructive to self.encoder_params (lots of calls to
-        # params.pop()), but we may need to create several encoders.  So we'll make a copy and use
-        # that instead of self.encoder_params.
-        return self._get_new_encoder(deepcopy(self.encoder_params), name)
+    def _get_new_sentence_encoder(self, encoder_type="default", name="sentence_encoder"):
+        return self._get_new_encoder(deepcopy(self.encoder_params[encoder_type]), name)
 
-    def _get_word_encoder(self):
+    def _get_word_encoder(self, encoder_type="default"):
```
matt-gardner: These `_word_encoder` methods can just be removed.

src/main/python/deep_qa/training/text_trainer.py
```diff
@@ -360,34 +359,36 @@ def _get_new_embedding(self, name: str, embedding_size: int, vocab_name: str='wo
...
         A sentence encoder takes as input a sequence of word embeddings, and returns as output a
         single vector encoding the sentence.  This is typically either a simple RNN or an LSTM, but
         could be more complex, if the "sentence" is actually a logical form.
         """
-        if self.sentence_encoder_layer is None:
-            self.sentence_encoder_layer = self._get_new_sentence_encoder()
-        return self.sentence_encoder_layer
+        if encoder_type not in self.sentence_encoder_layers:
+            encoder_layer_name = encoder_type + "_sentence_encoder"
+            new_sentence_encoder = self._get_new_sentence_encoder(encoder_type, encoder_layer_name)
```
matt-gardner: You can remove `_get_new_sentence_encoder` method, instead just using `_get_new_encoder`.

src/main/python/deep_qa/training/text_trainer.py
```diff
@@ -397,16 +398,18 @@ def _get_new_encoder(self, params: Dict[str, Any], name: str):
         set_regularization_params(encoder_type, params)
         return encoders[encoder_type](**params)
 
-    def _get_seq2seq_encoder(self, input_shape):
+    def _get_seq2seq_encoder(self, input_shape, encoder_type="default"):
```
matt-gardner: I would use `name` here, too, instead of `encoder_type`.

src/main/python/deep_qa/training/text_trainer.py
nelson-liu: yeah, i guess i was trying to cater to the idea of having multiple instantiations of one encoder, and have "types" that are essentially templates of params that you specify that you can pick from in instantiating.

I also tried as much as possible to not conflate the "name" of the layer as keras takes it, and the "name" of the layer in the sense of which parameters to assign it. i suppose it would be better if they were one and the same, though...

src/main/python/deep_qa/training/text_trainer.py
nelson-liu: the seq2seq as well?

src/main/python/deep_qa/training/text_trainer.py
nelson-liu: having just `encoder` would also pick up `seq2seq_encoder`, so i moved this check after `seq2seq_encoder` and made it an `elif`.

src/main/python/deep_qa/models/multiple_choice_qa/question_answer_memory_network.py
```diff
@@ -53,6 +53,6 @@ def _get_entailment_output(self, combined_input):
         # for a closer re-implementation of prior memory networks.
         answer_input = Input(shape=(self.num_options, self.max_answer_length), dtype='int32', name="answer_input")
         answer_embedding = self._embed_input(answer_input)
-        answer_encoder = EncoderWrapper(self._get_new_sentence_encoder(), name="answer_encoder")
+        answer_encoder = EncoderWrapper(self._get_encoder(), name="answer_encoder")
```
nelson-liu: not sure if this was the right fix.

matt-gardner: Almost.  I would do `self._get_encoder(name="answer")`, or even `self._get_encoder(name=self.answer_encoder_name)`, where you add a parameter to the class that allows specifying the key to use in the dictionary for the answer encoder.  That would let you either use a new encoder or share the encoder, depending on what you want to do.

src/main/python/deep_qa/training/text_trainer.py
```diff
@@ -74,12 +74,11 @@ def __init__(self, params: Dict[str, Any]):
         # one of the keys in `encoders`) to determine the type of the encoder, then pass the
         # remaining args to the encoder constructor.
         # Hint: Use lstm or cnn for sentences, treelstm for logical forms, and bow for either.
-        self.encoder_params = params.pop('encoder', {'default': {}})
+        self.encoder_params = params.pop('encoder', {'default': {}, 'word': {}})
```
matt-gardner: I don't think this is necessary.  Instead, I would implement some logic in `_get_encoder` around what to do if the requested key (like `'word'`) is not present.

src/test/python/deep_qa_test/common/models.py
```diff
@@ -19,7 +19,7 @@ def get_model(cls, additional_arguments=None):
     params['train_files'] = [TRAIN_FILE]
     params['validation_files'] = [VALIDATION_FILE]
     params['embedding_size'] = 5
-    params['encoder'] = {"default": {'type': 'bow'}}
+    params['encoder'] = {"default": {'type': 'bow'}, "word": {'type': 'bow'}}
```
matt-gardner: This shouldn't be necessary.

doc/training/trainers.rst
```diff
@@ -1,37 +1,37 @@
...
+.. automodule:: deep_qa.training.trainer
     :members:
     :noindex:
     :undoc-members:
     :show-inheritance:
 
-optimizers
-----------
+TextTrainer
+------------
```
nelson-liu: looks like this subtitle is a bit too long

src/main/python/deep_qa/data/tokenizers/word_and_character_tokenizer.py
```diff
@@ -13,9 +13,18 @@
 
 class WordAndCharacterTokenizer(Tokenizer):
     """
-    A WordAndCharacterTokenizer first splits strings into words, then splits those words into
-    characters, and returns a representation that contains _both_ a word index and a sequence of
+    A ``WordAndCharacterTokenizer`` first splits strings into words, then splits those words into
+    characters, and returns a representation that contains `both` a word index and a sequence of
```
nelson-liu: \`both\` to \*both\*

doc/training/trainers.rst
matt-gardner: Done.

src/main/python/deep_qa/data/tokenizers/word_and_character_tokenizer.py
matt-gardner: I wanted italics, not bold.

PR #177, opened 2017-01-30
Boosted query

bsharpataz:
added the question and answer boosted option for querying.  not sure if formatting/commenting is in line with what you want - please let me know what you'd like me to fix


src/main/scala/org/allenai/deep_qa/data/BackgroundCorpusSearcher.scala
```diff
@@ -82,19 +82,39 @@ class LuceneBackgroundCorpusSearcher(params: JValue) extends BackgroundCorpusSea
   lazy val settings = Settings.builder().put("cluster.name", esClusterName).build()
   lazy val esClient = TransportClient.builder().settings(settings).build().addTransportAddress(address)
 
-  override def getBackground(query: String): Seq[String] = {
+  override def getBackground(query: Query): Seq[String] = {
+    // Make the bool query and the more-or-less canonical query string for the
+    // consolidation step (below)
+    val queryBuilder = QueryBuilders.boolQuery()
+    var queryString = ""
```
matt-gardner: In scala, avoid `var` wherever possible.  Using `val` removes potential sources of error and makes it easier to parallelize code.  So, you could write this as:

```scala
val queryString = query match {
  case stringQuery: StringQuery => {
    queryBuilder.should(...)
    stringQuery.query
  }
  case boostedQuery: BoostedQuery => {
    ...
    boostedQuery.boostedQueries.map(...)
  }
}

File: src/main/scala/org/allenai/deep_qa/data/BackgroundCorpusSearcher.scala [hunk 2]
```diff
@@ -82,19 +82,39 @@ class LuceneBackgroundCorpusSearcher(params: JValue) extends BackgroundCorpusSea
...
+    query match {
+      case sq: StringQuery => {
```
matt-gardner: Avoid abbreviations like this.  It's better to call this `stringQuery`, so that I immediately know what it is when I see it used below.  Code readability is way more important than short lines: https://google.github.io/styleguide/cppguide.html#General_Naming_Rules.

src/main/scala/org/allenai/deep_qa/data/BackgroundCorpusSearcher.scala
```diff
@@ -82,19 +82,39 @@ class LuceneBackgroundCorpusSearcher(params: JValue) extends BackgroundCorpusSea
...
+        queryBuilder.should(QueryBuilders.matchQuery("text", sq.query))
+        queryString = sq.query
+      }
+      case bq: BoostedQuery => {
+        for ((qString, boostAmt) <- bq.boostedQueries) {
+          queryBuilder.should(QueryBuilders.matchQuery("text", qString).boost(boostAmt))
+          queryString = bq.boostedQueries.map(q => q._1).mkString(" ")
```
matt-gardner: The `q` inside a very short call to `map()` like this is one place where I make exceptions to the rule against abbreviations.  However, in this case you don't even need it at all, as you can just do `boostedQueries.map(_._1).mkString(" ")`.

src/main/scala/org/allenai/deep_qa/data/BackgroundCorpusSearcher.scala
```diff
@@ -126,3 +146,33 @@ class LuceneBackgroundCorpusSearcher(params: JValue) extends BackgroundCorpusSea
...
   }
 }
+
+
+// trait used for handling various types of queries
+trait Query {}
+
+
+// Query type for basic string queries
+class StringQuery(var query: String = "") extends Query {
```
matt-gardner: There's no need for this class to be mutable.  Mutability should be avoided where it is not necessary.  So you can just do `case class StringQuery(query: String) extends Query`, with no additional class body necessary.

src/main/scala/org/allenai/deep_qa/data/BackgroundCorpusSearcher.scala
```diff
@@ -126,3 +146,33 @@ class LuceneBackgroundCorpusSearcher(params: JValue) extends BackgroundCorpusSea
...
+
+  def set(queryIn: String): Unit = {
+    query = queryIn
+  }
+}
+
+
+// Query type for boosting different portions of the query
+class BoostedQuery(boostedIn: Seq[(String, Float)]) extends Query () {
```
matt-gardner: Again here, I don't think you actually need mutability.  You can just do `case class BoostedQuery(boostedQueries: Seq[(String, Float)]) extends Query`.

src/main/scala/org/allenai/deep_qa/pipeline/BackgroundCorpusSearcherStep.scala
```diff
@@ -97,14 +95,39 @@ class DefaultBackgroundCorpusSearcherStep(
...
     // TODO(matt): might want to make this streaming, for large input files
     val lines = fileUtil.readLinesFromFile(sentencesFile)
-    val indexedSentences = lines.map(line => {
+    val indexedSentences: Seq[(Int, Seq[Query])] = lines.map(line => {
       val fields = line.split("\t")
       sentenceFormat match {
-        case "plain sentence" => (fields(0).toInt, Seq(fields(1)))
+        case "plain sentence" => (fields(0).toInt, Seq(new StringQuery(fields(1))))
         case "question and answer" => {
+          val qText = fields(1)
```
matt-gardner: `s/qText/questionText/`

src/main/scala/org/allenai/deep_qa/pipeline/BackgroundCorpusSearcherStep.scala
```diff
@@ -97,14 +95,39 @@ class DefaultBackgroundCorpusSearcherStep(
...
           val answers = fields(2).split("###")
-          val queries = answers.map(fields(1) + " " + _).toSeq
-          (fields(0).toInt, Seq(fields(1)) ++ queries)
+          val stringQueries = answers
+              .map(a => new StringQuery(qText + " " + a))
```
matt-gardner: This is on its own line, so there's plenty of horizontal space.  I would avoid the abbreviation here, using `answer` instead of `a`.

src/main/scala/org/allenai/deep_qa/pipeline/BackgroundCorpusSearcherStep.scala
```diff
@@ -97,14 +95,39 @@ class DefaultBackgroundCorpusSearcherStep(
...
+              .toSeq
+          (fields(0).toInt, Seq(new StringQuery(qText)) ++ stringQueries)
+        }
+        case "question and answer boosted" => {
+          // Get the boosting values for question and answer
+          val questionBoost = JsonHelper.extractWithDefault[Float](
```
matt-gardner: You don't need to have the type parameter here, because it's implied by the third argument.  Also, I would use `1.0` instead of `1f`.

src/main/scala/org/allenai/deep_qa/pipeline/BackgroundCorpusSearcherStep.scala
```diff
@@ -97,14 +95,39 @@ class DefaultBackgroundCorpusSearcherStep(
...
+            params,
+            "question boost",
+            1f
+          )
+          val answerBoost = JsonHelper.extractWithDefault[Float](
+            params,
+            "answer  boost",
```
matt-gardner: There's an extra space here, which seems like it's not doing what you want.

src/main/scala/org/allenai/deep_qa/pipeline/BackgroundCorpusSearcherStep.scala
```diff
@@ -97,14 +95,39 @@ class DefaultBackgroundCorpusSearcherStep(
...
+            1f
+          )
+          val qText = fields(1)
```
matt-gardner: `s/qText/questionText/`

bsharpataz:
I dont' know if you see this comment -- the only one I didn't fix was the 1f to 1.0 because 1.0 is  Double and I need it to be a float for elasticSearch, so I changed it to 1.0f if that's ok?

PR #178, opened 2017-01-30
Fixed typos in equations

nelson-liu:
Sloppily left a bunch of RST typos in the math equations for knowledge selectors --- this should fix that.


src/main/python/deep_qa/layers/knowledge_selectors.py
```diff
@@ -104,14 +104,21 @@ class ParameterizedKnowledgeSelector(Layer):
     This layer concatenates the memory vector with each background sentence, passes them through a
     non-linearity, then does a softmax to get attention weights.
 
-    Equations:
-    Inputs: :math:`u` is the sentence encoding, :math:`z_t` are the background sentence encodings
-    Weights: :math:`W_1` (called ``self.dense_weights``), :math`v` (called ``self.dot_bias``)
-    Output: :math:``a_t``
+    Inputs:
+        - :math:`u` is the sentence encoding
```
matt-gardner: Now that it's formatted like this, I would probably write it as:

```
Inputs:
    - A sentence encoding :math:`u`, with shape ``(batch_size, encoding_dim)``
    - Background sentence encodings :math:`z_t` with shape ``(batch_size, knowledge_length, encoding_dim)``
```

PR #180, opened 2017-02-02
Add Gated Attention Reader

nelson-liu:
This PR implements the [Gated Attention Reader](https://www.semanticscholar.org/paper/Gated-Attention-Readers-for-Text-Comprehension-Dhingra-Liu/200594f44c5618fa4121be7197c115f78e6e110f).

This still needs (roughly in order):
- [ ] a rebase on master, followed by fixes for the new encoder format (#175)
- [ ] tests
- [ ] calculation of the final question encoding from the cloze token index in the biGRU, not the entire output.
- [ ] pretrained wordvectors
- [ ] character level embeddings
- [ ] docs


matt-gardner:
Should I wait until all of the checkboxes are done before reviewing this?

nelson-liu:
Yes, thanks!

nelson-liu:
lastly this PR trains the GAReader on sciQ / validates on sciQ, and then runs evaluation on the SciQ test set.

nelson-liu:
so i fixed the bug that made it impossible to use character embeddings + pretrained word vectors together due to conflicting layer names, but I'm finding it hard to add a non-regression test. Using an actual set of glove vectors would probably be way too slow, I guess perhaps we could write a "vector file" (and then gzip it, since the method takes gzipped files) like we do to test the models and read that?

matt-gardner:
Yeah, gzipping the file in the test seems reasonable.

nelson-liu:
So this PR as-is is ready for merging, but it's a bit abstruse and probably too large to review effectively. I've split it up into a bunch of mini PRs (#183, #184, #185, #186, #187). I'll leave this one open for archival purposes (/ a central branch to pull down if you actually want to run the code before it's all checked into master), and will close when its components are merged.

PR #181, opened 2017-02-03
Enable evaluation of models on test set in Python code

nelson-liu:
This PR lets you pass in a `test_file` to evaluate on after training is completed.

TODO:
- [ ] Add code to the scala evaluation classes to take advantage of this parameter.
- [ ] Enable test set evaluation on the validation set epoch with the maximum validation accuracy

### DIFF
--- a/build_tools/travis/install.sh
+++ b/build_tools/travis/install.sh
@@ -47,6 +47,3 @@ pip install -r requirements.txt
 
 # Install punkt tokenizer
 python -m nltk.downloader punkt
-
-# Install spacy data
-python -m spacy.en.download all
--- a/build_tools/travis/install.sh
+++ b/build_tools/travis/install.sh
@@ -47,3 +47,6 @@ pip install -r requirements.txt
 
 # Install punkt tokenizer
 python -m nltk.downloader punkt
+
+# Install spacy data
+python -m spacy.en.download all
--- a/build_tools/travis/install.sh
+++ b/build_tools/travis/install.sh
@@ -43,9 +43,13 @@ else
 fi
 
 # Install requirements via pip in our conda environment
-pip install -r requirements.txt
+pip install -U -r requirements.txt
 
+# List the packages to get their versions for debugging
 pip list
 
 # Install punkt tokenizer
 python -m nltk.downloader punkt
+
+# Install spacy data
+python -m spacy.en.download all
--- a/build_tools/travis/install.sh
+++ b/build_tools/travis/install.sh
@@ -45,5 +45,7 @@ fi
 # Install requirements via pip in our conda environment
 pip install -r requirements.txt
 
+pip list
+
 # Install punkt tokenizer
 python -m nltk.downloader punkt
--- a/build_tools/travis/test_script.sh
+++ b/build_tools/travis/test_script.sh
@@ -14,7 +14,7 @@ run_tests() {
     if [[ "$KERAS_BACKEND" == "tensorflow" ]]; then
         KERAS_BACKEND=tensorflow py.test -v --cov=deep_qa --durations=20
     else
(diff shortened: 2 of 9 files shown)


nelson-liu:
This PR runs the evaluation of training and validating the attention sum reader on sciQ, and then evaluating on a test set.

nelson-liu:
This PR is ready to be merged as is, but it'd be nice to be able to do test set evaluation on the _best_ epoch, as well as add this to the scala experiment code.

nelson-liu:
i think this can be merged as is. Travis pylint is failing, but it looks to be spurious and I can't figure out why it's complaining about that file.

matt-gardner:
I was wondering why I didn't get notified of this, and why it didn't get merged earlier - I guess you never added me as a reviewer!  I'll look over it now.

build_tools/travis/install.sh
```diff
@@ -47,3 +47,6 @@ pip install -r requirements.txt
 
 # Install punkt tokenizer
 python -m nltk.downloader punkt
+
+# Install spacy data
+python -m spacy.en.download all
```
matt-gardner: Do we actually need this?

build_tools/travis/test_script.sh
```diff
@@ -14,7 +14,7 @@ run_tests() {
     if [[ "$KERAS_BACKEND" == "tensorflow" ]]; then
         KERAS_BACKEND=tensorflow py.test -v --cov=deep_qa --durations=20
     else
-        KERAS_BACKEND=theano OMP_NUM_THREADS=4 THEANO_FLAGS='optimizer=fast_compile,openmp=True' py.test -v --cov=deep_qa --durations=20
+        KERAS_BACKEND=theano THEANO_FLAGS='optimizer=fast_compile' py.test -v --cov=deep_qa --durations=20
```
matt-gardner: Is this what fixed the theano CNN crash?

deep_qa/training/trainer.py
```diff
@@ -77,6 +77,10 @@ def __init__(self, params: Dict[str, Any]):
         # the `keras_validation_split` parameter to split the training data for validation.
         self.validation_files = params.pop('validation_files', None)
 
+        # The files containing the data that should be used for evaluation.
+        # The default of None mean to just not perform test set evaluation.
```
matt-gardner: `s/mean/means/`

build_tools/travis/test_script.sh
nelson-liu: yup

build_tools/travis/install.sh
nelson-liu: the tests complained that we didn't have this, but i'm not sure. I was frankly surprised as well, since it was a non-issue before but spacy recently had a new release so it could be related to that. It's a small download anyway (50 MB)...

nelson-liu:
[looks like it still complains](https://travis-ci.org/allenai/deep_qa/jobs/214843819)

I think the reason this didn't show up before was because we cache the conda environment, so we weren't pulling this new version. I don't think this behavior is desirable, so we can:

1. remove the conda environments from the cache (this is annoying since we need to reinstall packages every time)
2. cache the conda environment, but do `pip install -U`. I'm not sure how this interacts with the `requirements.txt` file, as I'd like to pull the _latest_ version of the packages that we haven't pinned but not upgrade the packages that we have pinned.

Or we could say nevermind about the new spacy version, pin it to the old version and don't install data. I'm personally partial to 2, thoughts?

nelson-liu:
looks like `-U` upgrades only if the requirement in `requirements.txt` isn't fulfilled --- i'll go ahead and do that

matt-gardner:
This looks good to me.  This does let us evaluate on the best model, by running training without test files, then by running test without training files, which will load the best model.  Feel free to merge.  I'll merge this soon if you haven't, unless there's something you still want to change.

nelson-liu:
That's true, but it'd be nice to do it all without manual intervention :)

PR #182, opened 2017-02-04
[MRG] Add experiment for omnibus8/omnibus4 with and without SciQ dataset

nelson-liu:
This dataset adds a file for running various experiments on SciQ. In addition, it has JValues to create combined datasets with Omnibus 4 + SciQ / Omnibus 8 + SciQ.


nelson-liu:
just going to leave some notes in these PRs to keep track of what branch runs what: this branch is used for the ASReader / GAReader Omnibus vs Omnibus + SciQ tests (run from scala).

src/main/scala/org/allenai/deep_qa/experiments/datasets/Science.scala
```diff
@@ -60,17 +75,55 @@ object ScienceDatasets {
...
+  )
+
+  val omnibusQaGradeFourTestBuscBackgroundFile: JValue = JohannesDatasets.makePassageBackgroundFile(
+    ScienceFiles.omnibusGradeFourTestSentences_questionAndAnswer,
+    "question and answer",
+    ScienceCorpora.buscElasticSearchIndex(3)
+  )
+
+  /*
+   * Fourth Grade Questions with BUSC Background (no multiple choice options)
```
nelson-liu: @matt-gardner I never quite figured out what these classes do (and correspondingly the `Eighth Grade Questions with BUSC Background` section), would you happen to know?

src/main/scala/org/allenai/deep_qa/experiments/SciQDatasetExperiments.scala
```diff
@@ -0,0 +1,302 @@
...
+    validationDataset: JValue
+  ): JValue = {
+    ("name" -> name) ~
+      ("model params" -> modelParams) ~
+      ("dataset" -> trainingDataset) ~
+      ("validation dataset" -> validationDataset)
+  }
+
+  // Turn SciQ reading comprehension train file with BUSC background to a dataset.
+  val sciQTrainDataset: JValue =
```
matt-gardner: I would put this in `SciQDatasets`.

src/main/scala/org/allenai/deep_qa/experiments/SciQDatasetExperiments.scala
```diff
@@ -0,0 +1,302 @@
...
+    ("dataset type" -> "from sentence producers") ~
+      ("data files" -> Seq(SciQDatasets.readingComprehensionTrainWithBuscBackgroundFile))
+
+  /*
+   * Turn Omnibus-8 reading comprehension train, test, and dev files
+   * with BUSC background to datasets.
+   */
+
+  val omnibusEightTrainReadingComprehensionDataset: JValue =
```
matt-gardner: These should probably also just be in `ScienceDatasets`.

src/main/scala/org/allenai/deep_qa/experiments/SciQDatasetExperiments.scala
```diff
@@ -0,0 +1,302 @@
...
+  val omnibusFourTestReadingComprehensionDataset: JValue =
+    ("dataset type" -> "from sentence producers") ~
+      ("data files" -> Seq(ScienceDatasets.readingComprehensionOmnibusQaGradeFourTestQuestionsWithBuscBackground))
+
+  /*
+   * Create combined datasets with Omnibus4Train + SciQ Train and
+   * Omnibus8Train + SciQ Train.
+   */
+
+  val combinedSciQTrainOmnibusEightTrainDataset: JValue =
```
matt-gardner: This one stays here, though.

src/main/scala/org/allenai/deep_qa/experiments/SciQDatasetExperiments.scala
```diff
@@ -0,0 +1,302 @@
...
+      ("embedding_dropout" -> 0.0) ~
+      ("patience" -> 0) ~
+      ("embedding_size" -> 256) ~
+      ("num_epochs" -> 5)
+    )
+  }
+
+  val gatedAttentionReader: JValue = {
+    Models.gatedAttentionReader merge
+    (("embedding_size" -> 100) ~
```
matt-gardner: I would put all of these default settings into `Models.gatedAttentionReader`, or something similar inside `Models`.

matt-gardner: Same with the AS Reader above.

src/main/scala/org/allenai/deep_qa/experiments/SciQDatasetExperiments.scala
```diff
@@ -0,0 +1,302 @@
...
+      ) ~
+      ("embedding_dropout" -> 0.0) ~
+      ("patience" -> 0) ~
+      ("num_epochs" -> 5)
+    )
+  }
+
+
+  def main(args: Array[String]) {
+    runASReaderOmnibusEightExperiment
```
matt-gardner: I think the convention is that if calling a method has side effects, you should include the parens, even if they're not strictly necessary.

http://docs.scala-lang.org/style/method-invocation.html
http://docs.scala-lang.org/style/naming-conventions

src/main/scala/org/allenai/deep_qa/experiments/SciQDatasetExperiments.scala
```diff
@@ -0,0 +1,302 @@
...
+  ): JValue = {
+    experiment(
+      name,
+      modelParams,
+      trainingDataset,
+      omnibusEightDevReadingComprehensionDataset
+    )
+  }
+
+  def combinedSciQOmnibusGradeFourExperiment(
```
matt-gardner: I don't think these ones are necessary.

src/main/scala/org/allenai/deep_qa/experiments/SciQDatasetExperiments.scala
```diff
@@ -0,0 +1,302 @@
...
+import org.allenai.deep_qa.pipeline.Evaluator
+
+import com.mattg.util.FileUtil
+import org.json4s.JsonDSL._
+import org.json4s._
+
+object SciQDatasetExperiments {
+  val fileUtil = new FileUtil
+
+  def experiment(
```
matt-gardner: It'd be nice to have this be in some general place somewhere, instead of duplicated like this.  Not sure where to put it, though.

src/main/scala/org/allenai/deep_qa/experiments/SciQDatasetExperiments.scala
```diff
@@ -0,0 +1,302 @@
...
+    val gaReaderOmnibusEightWithSciQDataset = omnibusGradeEightExperiment(
+      "GAReader omnibus eight plus SciQ Dataset",
+      gatedAttentionReader,
+      combinedSciQTrainOmnibusEightTrainDataset
+    )
+
+    val models = Seq(gaReaderOmnibusEightDefault, gaReaderOmnibusEightWithSciQDataset)
+    new Evaluator(Some("GAReader_omnibus_eight_plus_sciq_dataset"), models, fileUtil).runPipeline()
+  }
+
```
matt-gardner: Minus some nitpicks about where code should go, this experiment was set up quite nicely. Now that you've done this, do you have any feedback on the experiment creation process?  Good?  Bad?  What can be improved?

src/main/scala/org/allenai/deep_qa/experiments/datasets/Science.scala
```diff
@@ -60,17 +75,55 @@ object ScienceDatasets {
...
 
+  /////////////////////////////////////////////////////////////////////
+  // Omnibus QA
+  /////////////////////////////////////////////////////////////////////
+
+  /*
+   * Fourth Grade BUSC Background Files
+   */
+
+  val omnibusQaGradeFourTrainBuscBackgroundFile: JValue = SciQDatasets.makePassageBackgroundFile(
```
matt-gardner: The `SciQDatasets` reference made me worried there was a bug here for a minute, but you're just using a static reference to some implemented code.  It'd probably be a good idea to have a `DatasetUtil` object, or something similar, that groups all of the `makeFile` and `makeDataset` methods.

src/main/scala/org/allenai/deep_qa/experiments/datasets/Science.scala
matt-gardner: What classes are you asking about?

The difference between these and the "multiple true false" options is just how the question is encoded.  In the MTF case, you get four lines per question, with each line containing a question/answer pair.  In the QA version, you get a single line per question, with a column for the question and a column for the answer options.  I really dislike the naming convention we have in here, but I'm not sure how to make it better.  If you have suggestions, I'm happy to hear them.

src/main/scala/org/allenai/deep_qa/experiments/datasets/Science.scala
```diff
@@ -83,6 +136,51 @@ object ScienceDatasets {
     ScienceCorpora.buscElasticSearchIndex(3)
   )
 
+  /*
+   * Fourth Grade Questions in RC Format (BUSC-fetched passage,
+   * question, answer options)
+   */
+  val readingComprehensionOmnibusQaGradeFourTrainQuestionsWithBuscBackground: JValue =
```
matt-gardner: To follow the same convention as above, this should be named `omnibusRcGradeFourTrainQuestionsWithBuscBackground`.

src/main/scala/org/allenai/deep_qa/experiments/datasets/Science.scala
```diff
@@ -83,6 +136,51 @@ object ScienceDatasets {
...
+    ("sentence producer type" -> "qa and background to mc") ~
```
matt-gardner: I'm just noticing that you named this type "qa and background to mc".  It should be "qa and background to rc".

src/main/scala/org/allenai/deep_qa/experiments/datasets/Science.scala
```diff
@@ -135,18 +261,18 @@ object ScienceDatasets {
 object ScienceFiles {
   def makeMultipleTrueFalseQuestionAnswerFile(questionFile: String, outputFile: String): JValue = {
     ("sentence producer type" -> "question interpreter") ~
-    ("create sentence indices" -> true) ~
-    ("question file" -> questionFile) ~
-    ("output file" -> outputFile) ~
-    ("interpreter" -> ("type" -> "append answer"))
+      ("create sentence indices" -> true) ~
```
matt-gardner: Yes, this is definitely more readable.  My editor doesn't do this by default, so that's why it was how it was.  I'll have to see if I can fix that...

src/main/scala/org/allenai/deep_qa/experiments/datasets/Science.scala
```diff
@@ -60,17 +75,55 @@ object ScienceDatasets {
...
+   * Fourth Grade BUSC Background Files
+   */
+
+  val omnibusQaGradeFourTrainBuscBackgroundFile: JValue = SciQDatasets.makePassageBackgroundFile(
+    ScienceFiles.omnibusGradeFourTrainSentences_questionAndAnswer,
+    "question and answer",
+    ScienceCorpora.buscElasticSearchIndex(3)
+  )
+
+  val omnibusQaGradeFourDevBuscBackgroundFile: JValue = SciQDatasets.makePassageBackgroundFile(
```
matt-gardner: I had `ScienceDatasets` and `ScienceFiles` as two separate classes, and these would technically go under `ScienceFiles`.  I'm not thrilled by that organization, though; if you think there's a better way to organize all of this, feel free to make it better.

src/main/scala/org/allenai/deep_qa/experiments/datasets/Science.scala
```diff
@@ -95,12 +193,40 @@ object ScienceDatasets {
     ScienceCorpora.buscElasticSearchIndex(3)
   )
 
+  /*
+   * Eighth Grade Questions in RC Format (BUSC-fetched passage,
+   * question, answer options)
+   */
+
+  val readingComprehensionOmnibusQaGradeEightTrainQuestionsWithBuscBackground: JValue =
```
matt-gardner: Same issue here with the naming convention.

src/main/scala/org/allenai/deep_qa/experiments/datasets/Science.scala
nelson-liu: ah i was more wondering the difference between `omnibusQaGradeFourTrainQuestionsWithBuscBackground` and `readingComprehensionOmnibusQaGradeFourTrainQuestionsWithBuscBackground`.

matt-gardner: Does it make sense now?  It's just in how the data is formatted.

src/main/scala/org/allenai/deep_qa/experiments/SciQDatasetExperiments.scala
nelson-liu: i think the experiment creation process is kind of a pain for running small things, but for more methodical things like this it's pretty nice. would use again :)

src/main/scala/org/allenai/deep_qa/experiments/datasets/Science.scala
nelson-liu: sure, perhaps work for a future PR?

matt-gardner: Yeah, that's fine, just leave a TODO.

nelson-liu:
thanks for the review @matt-gardner , addressed your comments.

nelson-liu:
@matt-gardner did you want to look at this again before I merge?

src/main/scala/org/allenai/deep_qa/experiments/Models.scala
```diff
@@ -46,10 +46,94 @@ object Models {
...
+    ("patience" -> 0) ~
+    ("preferred_backend" -> "theano") ~
+    ("encoder" ->
+      ("question_final" ->
+        ("type" -> "bi_gru")~
+        ("output_dim" -> 128)
+      )
+    ) ~
+    ("seq2seq_encoder" ->
+      ("question_0" ->
```
matt-gardner: This looks fine; feel free to merge.  The only thing you might think about changing is using the `default` key here, so you don't have to repeat yourself with all of these encoders.  Note that this requires setting the `fallback_behavior` in [these lines](https://github.com/allenai/deep_qa/pull/187/files#diff-58cd40374724399ec279f21d52c0bec5R109) (which I think you should do, anyway).

PR #184, opened 2017-02-07
[MRG] add gated attention layer, tests, and docs

nelson-liu:
Added the gated attention layer described in [Dhingra et. al 2016](https://www.semanticscholar.org/paper/Gated-Attention-Readers-for-Text-Comprehension-Dhingra-Liu/200594f44c5618fa4121be7197c115f78e6e110f)

Built docs: 
API Summary for layers.attention: https://133-47793051-gh.circle-artifacts.com/0/home/ubuntu/deep_qa/doc/_build/html/layers/attention.html#gatedattention
Full API Reference for layers.attention: https://133-47793051-gh.circle-artifacts.com/0/home/ubuntu/deep_qa/doc/_build/html/api_doc/deep_qa.layers.attention.html#module-deep_qa.layers.attention.gated_attention


src/main/python/deep_qa/layers/attention/gated_attention.py
```diff
@@ -0,0 +1,99 @@
...
+        document_matrix, question_matrix, normalized_qd_attention = inputs
+
+        # question update is of shape (batch, document length, bigru hidden)
+        question_update = K.batch_dot(normalized_qd_attention, question_matrix, axes=[2, 1])
+
+        # use the gating function to calculate the new document representation
+        # which is of shape (batch, document length, biGRU hidden length)
+        if self.gating_function == "*":
+            return question_update * document_matrix
+        if self.gating_function == "+":
```
DeNeutoy: Here, I think we need to utilise the mask from the document matrix. In the multiplication case, it's fine as the question update already has zeros where they should be, but for the concatenation and addition cases, if we don't mask here and don't pass a mask forward from this layer, I don't think this is going to be correctly masked upstream.

src/test/python/deep_qa_test/layers/attention/gated_attention_test.py
```diff
@@ -0,0 +1,102 @@
+# pylint: disable=no-self-use,invalid-name
```
DeNeutoy: Might be good to have a test for the masked case as well, given I'm not 100% sure it would pass in the addition and concatenation cases.

src/main/python/deep_qa/layers/attention/gated_attention.py
```diff
@@ -0,0 +1,99 @@
...
+            raise ConfigurationError("Invalid gating function "
+                                     "{}, expected one of {}".format(self.gating_function,
+                                                                     GATING_FUNCTIONS))
+
+        super(GatedAttention, self).__init__(**kwargs)
+
+    def compute_mask(self, inputs, mask=None):
+        # pylint: disable=unused-argument
+        # We do not need a mask beyond this layer.
+        return None
```
matt-gardner: I second what Mark said.  In addition, I think you need to return a mask here, especially seeing how you use this in the model.  After the first run of the loop, you'll no longer have a mask.

nelson-liu:
thanks for the reviews, @DeNeutoy and @matt-gardner . I've addressed the comments, and added tests for all three masked cases. For the `compute_mask` method, I think the right thing to do is just re-return the mask of the document; thoughts?

matt-gardner:
Yes, returning the document mask seems like the right thing to do here.

PR #185, opened 2017-02-07
[MRG] add biGRU index selector and overlap layers, as well as tests and docs

nelson-liu:
Add layers to do the BiGRU index selecting for the question-evidence common word feature in the Gated Attention Reader, and also an `Overlap` layer.

Overlap has some hacky casting in the mask to get around the bug in the tensorflow backend of keras that casts unmasked input to `uint8` before concatenation.

built docs:
biGRU index selector in api summary: https://137-47793051-gh.circle-artifacts.com/0/home/ubuntu/deep_qa/doc/_build/html/layers/core_layers.html#bigruindexselector
biGRU index selector in full api reference: https://137-47793051-gh.circle-artifacts.com/0/home/ubuntu/deep_qa/doc/_build/html/api_doc/deep_qa.layers.html#module-deep_qa.layers.bigru_index_selector

overlap in api summary: https://137-47793051-gh.circle-artifacts.com/0/home/ubuntu/deep_qa/doc/_build/html/layers/core_layers.html#overlap
overlap in full api reference: https://137-47793051-gh.circle-artifacts.com/0/home/ubuntu/deep_qa/doc/_build/html/api_doc/deep_qa.layers.html#module-deep_qa.layers.overlap


src/main/python/deep_qa/layers/overlap.py
```diff
@@ -0,0 +1,58 @@
...
+    This Layer takes 2 inputs: a tensor_a (e.g. a document) and a tensor_b
+    (e.g. a question). It returns a one-hot vector with the same shape as tensor_a,
+    indicating at each index whether the element in tensor_a appears in tensor_b.
+
+    Inputs:
+        - tensor_a: shape ``(batch_size, length_a)``
+        - tensor_b shape ``(batch_size, length_b)``
+
+    Output:
+        - One-hot indicating overlap: shape ``(batch_size, length_a, 2)``
```
DeNeutoy: Why is the 2 there? Can't it just be `(batch_size, length_a)`? If it's going to be this way, can you just change the docs above to reflect that it is not the same shape as `tensor_a`.

src/main/python/deep_qa/layers/overlap.py
```diff
@@ -0,0 +1,58 @@
...
+        if K.backend() == "tensorflow":
+            tensor_a = inputs[0]
+            length_a = K.int_shape(tensor_a)[1]
+            return K.cast(K.ones_like(K.repeat_elements(K.expand_dims(tensor_a, 2),
+                                                        length_a,
+                                                        axis=2)), "bool")
+        else:
+            return None
+
+    def call(self, inputs, mask=None):
```
DeNeutoy: This needs to take into account masking - currently, elements that are masked at the same index in both tensors will be captured as overlapping, which we presumably don't want.

src/test/python/deep_qa_test/layers/overlap_test.py
```diff
@@ -0,0 +1,33 @@
+import numpy
```
DeNeutoy: Add overlap test for masked case, which I think will fail at the moment.

src/main/python/deep_qa/layers/overlap.py
nelson-liu: well it's a one hot vector, so we need `[0, 1]` or `[1, 0]` depending if there's overlap or not.

src/main/python/deep_qa/layers/overlap.py
nelson-liu: mm yeah good point.

src/main/python/deep_qa/layers/overlap.py
DeNeutoy: But can't it just be [1] if there is overlap and [0] if not?

PR #187, opened 2017-02-07
Add Gated Attention Reader (pt. II)

nelson-liu:
This PR won't work until the others are all merged, but this PR adds the gated attention reader model.

todo:
- [x] docs


nelson-liu:
> Also, I'm pretty concerned about how masking is treated in lines 120 through 126 - I think it's wrong. Can you just use MatrixAttention instead of the MaskedBatchDot and then MaskedSoftmax?

I tried using MatrixAttention before, but ran into memory issues from tiling the matrix (hence why i swapped to maskedbatchdot and maskedsoftmax)

matt-gardner:
The trouble with how you did it, though, is that `MaskedBatchDot` returns `None` for a mask, opting instead for zeros in the similarity matrix.  Then you pass those zeros through `MaskedSoftmax` with no mask.  It's not doing what you think it's doing.  Instead, I'd recommend a `BatchDot` layer that handles the mask properly, and then you can pass the result into `MaskedSoftmax` to get the right behavior.

nelson-liu:
i see, yeah that makes sense. instead of creating a `batchdot` layer, could i just add a mask to the `maskedbatchdot` layer?

matt-gardner:
If you're returning a mask, there's no reason to use `masked_batch_dot`; you can just use `K.batch_dot`.

nelson-liu:
> Just one high-level comment: you can just specify the default key in the seq2seq_encoder params, and give the right fallback behavior when you get the seq2seq encoder on lines 109 and 115. This would make it so you don't have to repeat yourself so much in the config file.

I was trying to favor being explicit over implicit, but yeah i agree. 

> If you're returning a mask, there's no reason to use masked_batch_dot; you can just use K.batch_dot.

that makes sense, I'll make a `BatchDot` layer that properly passes on a mask.

nelson-liu:
though im not sure how i feel about the naming --- `MaskedBatchDot` doesn't pass forward a mask, while `BatchDot` does.

matt-gardner:
Honestly, we probably don't need the MaskedBatchDot layer or function, because we should just be propagating the mask correctly.  It's really unlikely that you'd use the output of a batch dot without some other mask-using operating first (like a softmax).  We can leave them there, but it's probably a good idea to leave notes that they probably aren't necessary.

src/main/python/deep_qa/models/reading_comprehension/gated_attention_reader.py
```diff
@@ -0,0 +1,271 @@
...
+            # is not necessarily the embedding size in the second gated
+            # attention layer and beyond.
+
+            # We encode the question embeddings with a seq2seq encoder.
+            question_encoder = self._get_seq2seq_encoder(name="question_{}".format(i))
+            # shape: (batch size, question_length, 2*seq2seq hidden size)
+            encoded_question = question_encoder(question_embedding)
+
+            # We encode the document embeddings with a seq2seq encoder.
+            # Note that this is not the same encoder as used for the question
```
DeNeutoy: question -> question.

src/main/python/deep_qa/models/reading_comprehension/gated_attention_reader.py
```diff
@@ -0,0 +1,271 @@
...
+            # shape: (batch size, document_length, 2*seq2seq hidden size)
+            document_embedding = gated_attention_dropout([document_embedding])
+
+        # Last Layer
+        if self.use_qd_common_feature:
+            # get the one-hot features for common occurence
+            # shape (batch size, document_indices, 2)
+            qd_common_feature = Overlap()([document_indices,
+                                           question_indices])
+            # concatenate qd_common_feature with document embeddings
```
DeNeutoy: Capitals and punctuation.

src/main/python/deep_qa/models/reading_comprehension/gated_attention_reader.py
```diff
@@ -0,0 +1,271 @@
...
+        final_encoded_document = document_encoder(document_embedding)
+
+        if self.cloze_token is None:
+            # Get a final encoding of the question from a biGRU that does not return
+            # the sequence, and use it to calculate attention over the document.
+            final_question_encoder = self._get_encoder(name="question_final")
+            # shape: (batch size, 2*seq2seq hidden size)
+            final_encoded_question = final_question_encoder(question_embedding)
+        else:
+            # get a final encoding of the question by concatenating the forward
```
DeNeutoy: Capitals and punctuation for these comments.

src/main/python/deep_qa/models/reading_comprehension/gated_attention_reader.py
```diff
@@ -0,0 +1,271 @@
...
+        # the total probability of the option.
+        options_sum_layer = OptionAttentionSum(self.multiword_option_mode,
+                                               name="options_probability_sum")
+        # shape: (batch size, num_options)
+        options_probabilities = options_sum_layer([document_indices,
+                                                   document_probabilities,
+                                                   options_indices])
+
+        # We normalize the option_probabilities by dividing each
+        # element by L1 norm (sum) of the whole tensor.
```
DeNeutoy: This confused me a bit - it's not the L1 norm of the whole tensor, it's per-batch. Given this is a common operation (batch normalisation) perhaps just change the wording here slightly?

PR #191, opened 2017-02-10
add BatchDot layer, tests, and docs

nelson-liu:
this PR adds a `BatchDot` layer that simply takes the output of `K.batch_dot` and propagates a correct mask with it. I didn't bother testing the call function because I'm presuming `K.batch_dot` is well tested :)

### DIFF
--- a/src/main/python/deep_qa/layers/backend/batch_dot.py
+++ b/src/main/python/deep_qa/layers/backend/batch_dot.py
@@ -1,0 +1,86 @@
+import keras.backend as K
+from keras.layers import Layer
+from overrides import overrides
+
+
+class BatchDot(Layer):
+    """
+    This ``Layer`` calls ``K.batch_dot()`` on two inputs ``tensor_a`` and ``tensor_b``.
+    This function will work for tensors of arbitrary size as long as
+    ``abs(K.ndim(tensor_a) - K.ndim(tensor_b)) < 1``, due to limitations in ``K.batch_dot()``.
+
+    We always assume the dimension to perform the dot is the last one, and that
+    the masks have one fewer dimension that the tensors. Note that this layer
+    does not return zeroes in places that are masked, but does pass a correct
+    mask forward. If this then gets fed into ``masked_softmax``, for instance,
+    your tensor will be correctly normalized. We always assume the dimension to
+    perform the dot is the last one, and that the masks have one fewer
+    dimension than the tensors.
+
+    Inputs:
+        - tensor_a: tensor with ``ndim >= 2``.
+        - tensor_b: tensor with ``ndim >= 2``.
+
+    Output:
+        - a_dot_b
+
+    """
+    def __init__(self, **kwargs):
+        self.supports_masking = True
+        super(BatchDot, self).__init__(**kwargs)
+
+    @overrides
+    def compute_mask(self, inputs, mask=None):
+        # pylint: disable=unused-argument
+        tensor_a, tensor_b = inputs
+        mask_a, mask_b = mask
+        a_dot_axis = K.ndim(tensor_a) - 1
+        b_dot_axis = K.ndim(tensor_b) - 1
+
+        if K.ndim(tensor_a) < K.ndim(tensor_b):
+            # To simplify the logic below, we'll make sure that tensor_a is always the bigger one.
+            tensor_a, tensor_b = tensor_b, tensor_a
+            mask_a, mask_b = mask_b, mask_a
+
+        if mask_a is None and mask_b is None:
+            return None
+        elif mask_a is None:
+            # (batch_size, a_length)
+            mask_a = K.sum(K.ones_like(tensor_a), axis=-1)
+        elif mask_b is None:
+            # (batch_size, b_length)
+            sum_axis = -1
+            if b_dot_axis < a_dot_axis:
+                sum_axis -= 1
+            mask_b = K.sum(K.ones_like(tensor_b), axis=sum_axis)
+        # Casting masks to float since we TF would complain if we multiplied bools.
+        float_mask_a = K.cast(mask_a, 'float32')
+        float_mask_b = K.cast(mask_b, 'float32')
+
+        if b_dot_axis < a_dot_axis:
+            float_mask_b = K.expand_dims(float_mask_b, dim=-1)
+        else:
+            float_mask_a = K.expand_dims(float_mask_a, dim=-1)
+            float_mask_b = K.expand_dims(float_mask_b, dim=-2)
+        # (batch_size, a_length, b_length)
+        a2b_mask = float_mask_a * float_mask_b
+        return a2b_mask
+
+
+    @overrides
+    def get_output_shape_for(self, input_shape):
+        # This assumes that we do the dot product over the last dimension, so this
+        # will have to change if that assumption changes in masked_batch_dot.
+        a_out_shape = tuple([input_shape[0][i] for i in range(0, len(input_shape[0]) - 1)])
+        b_out_shape = tuple([input_shape[1][i] for i in range(1, len(input_shape[1]) - 1)])
+        final_out_shape = a_out_shape + b_out_shape
+        if len(final_out_shape) == 1:
+            final_out_shape += (1,)
+        return final_out_shape
+
+    @overrides
+    def call(self, x, mask=None):
+        tensor_a, tensor_b = x
+        a_dot_axis = K.ndim(tensor_a) - 1
+        b_dot_axis = K.ndim(tensor_b) - 1
+        return K.batch_dot(tensor_a, tensor_b, axes=(a_dot_axis, b_dot_axis))
--- a/src/main/python/deep_qa/layers/backend/batch_dot.py
+++ b/src/main/python/deep_qa/layers/backend/batch_dot.py
@@ -58,20 +58,20 @@ class BatchDot(Layer):
             # tensor_a has less dimensions than tensor_b.
             # We tile tensor_a to have the same shape as tensor_b
             float_mask_a = K.expand_dims(float_mask_a, dim=-1)
-            # We take the shape of mask_b as a hacky way of getting
+            # We take the shape of tensor_b as a hacky way of getting
             # around K.int_shape() and the Theano backend.
             tiled_float_mask_a = K.repeat_elements(float_mask_a,
-                                                   K.int_shape(mask_b)[-1],
+                                                   K.int_shape(tensor_b)[-2],
                                                    mask_b_dot_axis)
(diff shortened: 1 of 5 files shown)


src/main/python/deep_qa/layers/backend/batch_dot.py
```diff
@@ -0,0 +1,86 @@
...
+            float_mask_b = K.expand_dims(float_mask_b, dim=-2)
+        # (batch_size, a_length, b_length)
+        a2b_mask = float_mask_a * float_mask_b
+        return a2b_mask
+
+
+    @overrides
+    def get_output_shape_for(self, input_shape):
+        # This assumes that we do the dot product over the last dimension, so this
+        # will have to change if that assumption changes in masked_batch_dot.
```
matt-gardner: Update comment.

src/test/python/deep_qa_test/layers/backend/batch_dot_test.py
```diff
@@ -0,0 +1,32 @@
...
+import numpy
+from numpy.testing import assert_almost_equal
+import keras.backend as K
+
+from deep_qa.layers.backend.batch_dot import BatchDot
+
+
+class TestMaskedBatchDotLayer:
+    def test_compute_mask(self):
+        tensor_a = K.variable(numpy.array([[[0.3, 0.1], [0.4, 0.2], [0.8, 0.1]],
```
matt-gardner: You can just do `numpy.random.rand(shape)`; it would make this a little easier to read, and easier to write more tests.

src/test/python/deep_qa_test/layers/backend/batch_dot_test.py
```diff
@@ -0,0 +1,32 @@
...
+                                           [[0.3, 0.1], [0.4, 0.2], [0.8, 0.1]]]))
+        mask_a = K.variable(numpy.array([[1, 0, 1], [1, 1, 0]]))
+        tensor_b = K.variable(numpy.array([[[0.2, 0.6], [0.4, 0.3], [0.5, 0.7], [0.1, .6]],
+                                           [[0.2, 0.6], [0.4, 0.3], [0.5, 0.7], [0.1, .6]]]))
+        mask_b = K.variable(numpy.array([[0, 1, 1, 1], [1, 0, 1, 1]]))
+        calculated_mask = K.eval(BatchDot().compute_mask([tensor_a, tensor_b],
+                                                         [mask_a, mask_b]))
+        assert_almost_equal(calculated_mask, numpy.array([[[0.0, 1.0, 1.0, 1.0],
```
matt-gardner: I feel like this would be easier to read and write if you did something like

```python
mask_a = numpy.ones(shape)
mask_a[0][1] = 0
mask_a[1][2] = 0
...
assert_almost_equal(calculated_mask[0][1], numpy.zeros(shape))
assert_almost_equal(calculated_mask[1][2], numpy.zeros(shape))
...
```

The reason I suggest this is because it'd be nice to have some more tests.  Switching the order of the masks above makes me nervous, because I'm not sure the mask shape is going to match the output shape anymore.  Can you add a test where you switch the order of the inputs, and make sure you get the right output?  Also, it'd be nice to have some tests with higher-order tensors, to be sure this still works.  And even though `K.batch_dot` is well tested, it'd still be nice to have one simple test with order 4 or so tensors, so that we can be sure that our layer does what we think it does, and that it breaks if keras changes underneath us.

Specifically, one test that does something like the following would probably be enough:

```python
batch_size = 3
tensor_a = numpy.random.randInt(7, shape=(batch_size, 4, 5))
tensor_b = numpy.random.randInt(7, shape=(batch_size, 6, 2, 5))
# Manually set some values to 1 here, which will be masked later (1 and not 0 so that masked values are still non-zero in the output)
input_tensor_a = Input(shape=(4, 5))
masked_tensor_a = Masking(input_tensor_a, mask_value=1)
input_tensor_b = Input(shape=(6, 2, 5))
masked_tensor_b = Masking(input_tensor_b, mask_value=1)
a_dot_b = BatchDot()([masked_tensor_a, masked_tensor_b])
a_dot_b_mask = OutputMask()(a_dot_b)
model = Model(inputs=[input_tensor_a, input_tensor_b], output=[a_dot_b, a_dot_b_mask])
a_dot_b_tensor, mask_tensor = model.predict([tensor_a, tensor_b])
# Test that the dot happened like you expected, something like this:
for i in range(batch_size):
    assert_almost_equal(a_dot_b_tensor[i], numpy.tensordot(tensor_a[i], tensor_b[i], 1))
    # Check that the values in the output mask are 0 where you set values to 1 above.
a_dot_b_tensor, mask_tensor = model.predict([tensor_b, tensor_a])
# Repeat the asserts above, but flipped.
```

src/test/python/deep_qa_test/layers/backend/batch_dot_test.py
```diff
@@ -0,0 +1,242 @@
...
+                                                         [mask_a, mask_b]))
+        assert calculated_mask.shape == (batch_size, 3, 4)
+        assert_almost_equal(calculated_mask, numpy.array([[[1.0, 1.0, 1.0, 1.0],
+                                                           [0.0, 0.0, 0.0, 0.0],
+                                                           [1.0, 1.0, 0.0, 1.0]],
+                                                          [[1.0, 1.0, 1.0, 1.0],
+                                                           [1.0, 1.0, 0.0, 1.0],
+                                                           [0.0, 0.0, 0.0, 0.0]]]))
+
+    def test_compute_mask_a_smaller_than_b(self):
```
matt-gardner: This isn't testing `compute_mask` anymore, right?

src/test/python/deep_qa_test/layers/backend/batch_dot_test.py
```diff
@@ -0,0 +1,242 @@
...
+            # Check that the values in the output mask are 0 where the
+            # values were set to 1 above.
+        assert mask_tensor[0][1][0] == 0
+        assert mask_tensor[0][1][1] == 0
+        assert mask_tensor[0][2][1] == 0
+        assert mask_tensor[1][3][0] == 0
+        assert mask_tensor[1][3][1] == 0
+
+    def test_output_shapes(self):
+        mbd = BatchDot()
```
matt-gardner: `mbd` doesn't make sense as a name for `BatchDot`.

src/main/python/deep_qa/layers/backend/batch_dot.py
```diff
@@ -0,0 +1,120 @@
...
+    ``abs(K.ndim(tensor_a) - K.ndim(tensor_b)) < 1``, due to limitations in ``K.batch_dot()``.
+
+    We always assume the dimension to perform the dot is the last one, and that
+    the masks have one fewer dimension that the tensors. Note that this layer
+    does not return zeroes in places that are masked, but does pass a correct
+    mask forward. If this then gets fed into ``masked_softmax``, for instance,
+    your tensor will be correctly normalized. We always assume the dimension to
+    perform the dot is the last one, and that the masks have one fewer
+    dimension than the tensors.
+
```
matt-gardner: Some more documentation here about what happens with various example shapes would be good.

nelson-liu: done

src/main/python/deep_qa/layers/backend/batch_dot.py
```diff
@@ -8,6 +8,8 @@ class BatchDot(Layer):
     This ``Layer`` calls ``K.batch_dot()`` on two inputs ``tensor_a`` and ``tensor_b``.
     This function will work for tensors of arbitrary size as long as
     ``abs(K.ndim(tensor_a) - K.ndim(tensor_b)) < 1``, due to limitations in ``K.batch_dot()``.
+    The when the input tensors have more than three dimensions, they must have the same shape, except
```
matt-gardner: `s/The when/When/`

src/main/python/deep_qa/layers/backend/batch_dot.py
```diff
@@ -24,6 +26,69 @@ class BatchDot(Layer):
...
+
+    Next, let's look at an example where ``tensor_a`` and ``tensor_b`` are "uneven" (different number
+    of dimensions). Let the shape of ``tensor_a`` be ``(2, 4, 2)``, and let the shape of ``tensor_b``
+    be ``(2, 4, 3, 2)``. The mask accompanying these inputs always has one less dimension, so the
+    ``tensor_a_mask`` has shape ``(2, 4)`` and ``tensor_b_mask`` has shape ``(2, 4, 3)``. The shape
+    of the ``batch_dot`` output would thus be ``(2, 4, 3)``. In the case of uneven tensors, we always
+    expand the last dimension of the smaller tensor to make them even. Thus in this case, we expand
+    ``tensor_a`` to get a new shape of ``(2, 4, 2, 1)``. Now we are taking the ``batch_dot`` of a
+    tensor with shape ``(2, 4, 2, 1)`` and ``(2, 4, 3, 2)``. Note that the first two dimensions of
+    this tensor are the same ``(2, 4)`` -- this is an requirement imposed by ``K.batch_dot``.
```
matt-gardner: `s/an requirement/a requirement/`

matt-gardner: Also, add a note that this is _different_ from the documented behavior of `K.batch_dot` - the documentation is wrong.

src/main/python/deep_qa/layers/backend/batch_dot.py
```diff
@@ -24,6 +26,69 @@ class BatchDot(Layer):
...
+    that the mask has the same shape.
+
+    >>> import keras.backend as K
+    >>> tensor_a = K.ones(shape=(2, 4, 2))
+    >>> tensor_b = K.ones(shape=(2, 4, 3, 2))
+    >>> tensor_a_expanded = K.expand_dims(tensor_a, dim=-1)
+    >>> unsqueezed_bd = K.batch_dot(tensor_a_expanded, tensor_b, axes=(2,3))
+    >>> final_bd = K.squeeze(unsqueezed_bd, axis=K.ndim(tensor_a)-1)
+    >>> K.eval(final_bd).shape
+    (2, 4, 3)
```
matt-gardner: Do these actually get run as tests by anything?

nelson-liu: it can if you want it to --- `py.test -v --doctest-modules src/main/python/deep_qa/layers/`
I tried doing `py.test -v --doctest-modules src/main/python/deep_qa/`, but it complains about some code in `contrib/` with nltk data or something.

nelson-liu:
merging this

PR #193, opened 2017-02-12
add files for transfer experiments / enable training on multiple datasets sequentially

nelson-liu:
This is a bunch of python json configs for the transfer experiments I've been running, and it also has some scala files to create the combined SciQ+WDW dataset.


example_experiments/transfer/asreader_joint_wdw_sciq_train_sciq_test.json
```diff
@@ -0,0 +1,30 @@
+// train a model on a combined dataset with SciQ train and a subset of WhoDidWhat train and see how well it does on SciQ.
```
matt-gardner: I don't know that there's a whole lot of value in having json files in `example_experiments` that are just one-line changes from other examples.  The scala experiment code is the right place to put all of the experiments that you actually run; this is just for examples of how to run various models.

Also, just a note for something I've been thinking about: I think we should split this repo into two: one for the python code, which we can then set up as a typical python library, and put it on pypi, and so on, and one for the scala experiment code that uses the python library.  I think that'd make everything cleaner and easier to document and understand.  Not sure when we'll actually get around to doing this, but it's on my mind.

src/main/python/deep_qa/training/trainer.py
```diff
@@ -53,6 +53,8 @@ def __init__(self, params: Dict[str, Any]):
         # Upper limit on the number of training instances.  If this is set, and we get more than
         # this, we will truncate the data.
         self.max_training_instances = params.pop('max_training_instances', None)
+        self.max_tuning_instances = params.pop('max_tuning_instances', None)
```
matt-gardner: I think it makes more sense to just change the API so that `train_files` is optionally a list of lists, so you can specify some number of datasets to train the model with, sequentially. Then you don't have to add another parameter if you want to train on three datasets in turn, and everything is just cleaner.

src/main/scala/org/allenai/deep_qa/experiments/TransferExperiments.scala
```diff
@@ -0,0 +1,28 @@
...
+        ("instances to keep" -> linesToKeep) ~
+        ("output directory" -> s"/efs/data/dlfa/who_did_what/processed/truncated/train/")
+
+    val combinedDataset: JValue =
+      ("dataset type" -> "combined") ~
+        ("datasets" -> Seq(
+          truncatedQuestions,
+          SciQDatasets.sciQTrainDataset)) ~
+        ("output directory" -> s"/efs/data/dlfa/sciq/sciq_and_who_did_what_combined/train/")
+        combinedDataset
```
matt-gardner: Indentation on this line is wrong.  It should be indented the same as `val combinedDataset`.

src/main/scala/org/allenai/deep_qa/pipeline/QaBackgroundToMcStep.scala
```diff
@@ -19,7 +19,7 @@ import scala.sys.process.ProcessLogger
   * background sentences has the format
   * "[index][tab][background1][tab][background2][tab]...". The output file
   * is in the format of a MCReadingComprehensionInstance with
-  * "[index][tab][passage][tab][question][tab][options][tab][label]".
+  * "[passage][tab][question][tab][options][tab][label]".
   * The background sentences compose the passage in this case.
  */
 class QaBackgroundToMcStep(
```
matt-gardner: This should be `QaBackgroundToRcStep`.

example_experiments/transfer/asreader_joint_wdw_sciq_train_sciq_test.json
nelson-liu: > I don't know that there's a whole lot of value in having json files in example_experiments that are just one-line changes from other examples. The scala experiment code is the right place to put all of the experiments that you actually run; this is just for examples of how to run various models.

I agree, I sorta just dumped them there for now. haven't gotten around to writing the proper hooks for making the scala code work with the new python capabilities.


>Also, just a note for something I've been thinking about: I think we should split this repo into two: one for the python code, which we can then set up as a typical python library, and put it on pypi, and so on, and one for the scala experiment code that uses the python library. I think that'd make everything cleaner and easier to document and understand. Not sure when we'll actually get around to doing this, but it's on my mind.

+1

src/main/python/deep_qa/training/trainer.py
nelson-liu: good idea!

example_experiments/asreader_wdw_sciq_train_sciq_test.json
```diff
@@ -0,0 +1,30 @@
...
+    },
+    "optimizer": {"type": "adam",
+                  "clipnorm": 10.0
+                 },
+    "embedding_dropout": 0.0,
+    "patience": 0,
+    "embedding_size": 256,
+    "num_epochs": 5,
+    "train_files": [["/efs/data/dlfa/who_did_what/processed/strict/train.tsv"], ["/efs/data/dlfa/sciq/processed/train_no_passage_with_background_as_passages.tsv"]],
+    "validation_files": ["/efs/data/dlfa/sciq/processed/test_no_passage_with_background_as_passages.tsv"],
```
matt-gardner: This should crash.  If `train_files` is a list, `validation_files` also needs to be a list.

src/main/python/deep_qa/training/trainer.py
```diff
@@ -266,8 +285,13 @@ def train(self):
             debug_masks = self.debug_params.get('masks', [])
             debug_data = self.debug_params['data']
             if debug_data == "training":
-                self.debug_dataset = self.training_dataset
```
matt-gardner: `self._debug` looks at `self.debug_input`.  I think it'd be better here to just allow debugging on the _current_ dataset.  So, I would move this whole block to line 319, in the training dataset loop.  Then you don't need to use `training_[0-9]`.

src/main/python/deep_qa/training/trainer.py
```diff
@@ -236,17 +237,35 @@ def train(self):
 
         # First we need to prepare the data that we'll use for training.
         logger.info("Getting training data")
-        self.training_dataset = self._load_dataset_from_files(self.train_files)
+        if isinstance(self.train_files[0], list):
+            # We want to tune on other datasets.
+            self.training_datasets = []
```
matt-gardner: In order to avoid keeping lots of datasets around in memory, I would probably recommend loading the data in the training loop.  This interacts poorly with pre-training, though...

We need to think about how to fit the data indexer in cases like this.  Maybe leave a TODO about that.

example_experiments/asreader_wdw_sciq_train_sciq_test.json
nelson-liu: why is that?

matt-gardner: To avoid subtle bugs.  Did you really want to do early stopping on SciQ when you were pre-training on WDW?  To avoid even the possibility of those kinds of errors (which would not be obvious at all in the output), it's best to make sure the user specifies lists in both places if they are doing this.

src/test/python/deep_qa_test/training/text_trainer_test.py
```diff
@@ -54,21 +54,16 @@ def new_debug(output_dict, epoch):  # pylint: disable=unused-argument
...
-            assert word_masks[1][0] == 1
-            assert word_masks[1][1] == 1
-            assert word_masks[1][2] == 1
-            assert word_masks[2][0] == 0
-            assert word_masks[2][1] == 1
-            assert word_masks[2][2] == 1
-            assert word_masks[3][0] == 0
-            assert word_masks[3][1] == 0
-            assert word_masks[3][2] == 1
+            print(word_masks)
```
matt-gardner: Remove print statement.

example_experiments/transfer/asreader_wdw_sciq_mix_train_sciq_val.json
```diff
@@ -0,0 +1,30 @@
+// train a model on a combined dataset with SciQ train and a subset of WhoDidWhat train and see how well it does on SciQ.
```
matt-gardner: Your commit says you're "temporarily" adding these files back - did you mean to keep them in the PR?

nelson-liu:
im going to close this PR and abandon the associated changes in the python code -- a much more elegant solution to this problem is to load a model and fit it some more. there's some scala code in here (from when the repos were joined) that is important, so i'll make a pr to deep_qa_experiments with those

nelson-liu:
nevermind, @matt-gardner already moved the scala code to the experiments repo so no action needs to be taken here.

PR #194, opened 2017-02-13
added noisy_or layer and tests

bsharpataz:
PR 1 of many... noisy_or layer and test
the constraint is something I played with and my use, but may just delete....? I don't know how that works in the PR world, if I should remove it for the PR or leave it in, or put it in its own file somewhere...


bsharpataz:
These tests pass fine on my local... ?

src/test/python/deep_qa_test/layers/noisy_or_test.py
```diff
@@ -0,0 +1,45 @@
...
+        axis = 2
+        noisy_or_layer = NoisyOr(axis=axis)
+        output = noisy_or_layer(input_layer)
+        model = Model([input_layer], output)
+
+        # Testing general unmasked case.
+        original_data = np.array([[[0.2, 0.1],
+                                  [0.5, 0.3],
+                                  [0.3, 0.7]]])
+        q = K.get_value(noisy_or_layer.q)
```
nelson-liu: hmm we've always used `eval`

bsharpataz: so K.eval(noisy_or_layer.q) instead?

src/main/python/deep_qa/layers/noisy_or.py
```diff
@@ -0,0 +1,88 @@
...
+        self.prefix = prefix
+        super(NoisyOr, self).__init__(**kwargs)
+
+    def build(self, input_shape):
+        # Create a trainable weight variable for the noise parameter
+        self.noise_parameter = self.add_weight(shape=(1,),
+                                               name=self.prefix + '_q',
+                                               initializer=self.param_init,
+                                               regularizer=L1L2Regularizer(l2=0.001),
+                                               # constraint=BetweenZeroAndOne(),
```
matt-gardner: Why is this commented out?  Did it not work?

src/main/python/deep_qa/layers/noisy_or.py
```diff
@@ -0,0 +1,88 @@
...
+from keras.layers import Layer
+from keras.initializations import identity
+from keras.constraints import Constraint, NonNeg
+from keras.regularizers import L1L2Regularizer
+from overrides import overrides
+
+class BetweenZeroAndOne(Constraint):
+    """Constrains the weights to be between zero and one
+    """
+    def __call__(self, p):
```
matt-gardner: Avoid names like `p` wherever possible; descriptive names make the code _way_ more readable.

src/main/python/deep_qa/layers/noisy_or.py
```diff
@@ -0,0 +1,88 @@
...
+        # handle negatives
```
matt-gardner: By "handle negatives" here, you mean that you're clipping anything below zero to be epsilon?  I'm pretty sure that's what this logic is doing.  Complete sentences explaining what the code is doing when it's not clear are better than sentence fragments: https://google.github.io/styleguide/cppguide.html#Punctuation,_Spelling_and_Grammar.

src/main/python/deep_qa/layers/noisy_or.py
```diff
@@ -0,0 +1,88 @@
...
+        p *= K.cast(p >= 0., K.floatx())
+        leaky_zeros_mask = K.epsilon() * K.cast(K.equal(p, K.zeros_like(p)), 'float32')
```
matt-gardner: Can't you just do `K.cast(p == 0., 'float32')`?  Using `K.zeros_like(p)` allocates a new tensor, which shouldn't be necessary here.

src/main/python/deep_qa/layers/noisy_or.py
```diff
@@ -0,0 +1,88 @@
...
+    def compute_mask(self, inputs, mask=None):
+        # pylint: disable=unused-argument
+        # We do not need a mask beyond this layer.
+        # (probably right)
+        if mask is not None:
+            return K.any(mask, axis=-1)
+        return None
+
+    def call(self, x, mask=None):
+        """
```
matt-gardner: I think this docstring is a lot more useful on the class, instead of the `call()` method.  You could duplicate it here, if you want, but this information needs to be in a class docstring.  For a good example, see here: https://github.com/allenai/deep_qa/blob/master/src/main/python/deep_qa/layers/attention/gated_attention.py.

src/main/python/deep_qa/layers/noisy_or.py
```diff
@@ -0,0 +1,88 @@
...
+    def __init__(self, axis=-1, param_init='uniform', prefix="noisy_or", **kwargs):
+        self.axis = axis
+        self.param_init = param_init
+        self.supports_masking = True
+        self.prefix = prefix
+        super(NoisyOr, self).__init__(**kwargs)
+
+    def build(self, input_shape):
+        # Create a trainable weight variable for the noise parameter
+        self.noise_parameter = self.add_weight(shape=(1,),
```
matt-gardner: pylint is complaining about this being defined outside of `__init__`.  I agree with pylint here - any class variable should show up in `__init__` - it helps with code readability, and IDE tooling.  So just add `self.noise_parameter = None` to the `__init__` method, ideally with a comment about what it's used for (even if it seems obvious).

src/main/python/deep_qa/layers/noisy_or.py
```diff
@@ -0,0 +1,88 @@
...
+        return p
+
+
+class NoisyOr(Layer):
+
+    def __init__(self, axis=-1, param_init='uniform', prefix="noisy_or", **kwargs):
+        self.axis = axis
+        self.param_init = param_init
+        self.supports_masking = True
+        self.prefix = prefix
```
matt-gardner: What's `self.prefix`?  Can you just use `self.name`?

src/main/python/deep_qa/layers/noisy_or.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def call(self, x, mask=None):
+        """
+        Calculate the noisy-or probability across a given tensor axis based on
+        noisy-or equation:
+            p(x) = 1 - Prod_(i=1:N)[1 - q * p(x|y_n)]
+
+        Parameters
+        ----------
+        inputs: Tensor
```
matt-gardner: Can you give the expected shape here?

src/main/python/deep_qa/layers/noisy_or.py
```diff
@@ -0,0 +1,88 @@
...
+        # (probably right)
+        if mask is not None:
+            return K.any(mask, axis=-1)
+        return None
+
+    def call(self, x, mask=None):
+        """
+        Calculate the noisy-or probability across a given tensor axis based on
+        noisy-or equation:
+            p(x) = 1 - Prod_(i=1:N)[1 - q * p(x|y_n)]
```
matt-gardner: A comment saying that `q` is `self.noise_parameter` would be nice.

src/main/python/deep_qa/layers/noisy_or.py
```diff
@@ -0,0 +1,88 @@
...
+        -------
+        output: Tensor
+            The combined probability of the even happening, given the parent events
+            Output dims are one less than input dims
+        """
+        # dim: (b, ..., num_probs)
+        noisy_probs = self.noise_parameter * x
+
+        # dim: (b, ..., num_probs)
+        noisy_probs = K.ones_like(x) - noisy_probs
```
matt-gardner: Does doing `1.0 - noisy_probs` or `noisy_probs * -1 + 1` work?  If it did, you could avoid a tensor allocation.  Not sure if it does or not (and I added the second one because of how operator overloading works; it might work if the first doesn't).

src/test/python/deep_qa_test/layers/noisy_or_test.py
```diff
@@ -0,0 +1,45 @@
...
+import numpy as np
+from numpy.testing import assert_array_almost_equal
+
+from keras import backend as K
+from keras.layers import Input, Masking
+from keras.models import Model
+from deep_qa.layers.noisy_or import NoisyOr, BetweenZeroAndOne
+
+class TestNoisyOr(TestCase):
+    # pylint: disable=no-self-use
```
matt-gardner: You can put this at the top of the file, as done in other tests.

src/test/python/deep_qa_test/layers/noisy_or_test.py
```diff
@@ -0,0 +1,45 @@
...
+                                   [0.65, 0.8],
+                                   [0.9, 0.15]]])
+        batch_result = model.predict([batch_original_data])
+        batch_desired_result = 1.0 - np.prod(1.0 - (q * batch_original_data), axis=axis)
+        assert_array_almost_equal(batch_result, batch_desired_result)
+
+
+    def test_between_zero_and_one_constraint(self):
+        p = K.variable(np.asarray([0.35, -0.4, 1.0, 1.2]), dtype='float32')
+        desired_result = np.asarray([0.35, 0.0, 1.0, 1.0])
```
matt-gardner: Should the second value be epsilon, not 0?

src/test/python/deep_qa_test/layers/noisy_or_test.py
```diff
@@ -0,0 +1,45 @@
...
+        axis = 2
+        noisy_or_layer = NoisyOr(axis=axis)
+        output = noisy_or_layer(input_layer)
+        model = Model([input_layer], output)
+
+        # Testing general unmasked case.
+        original_data = np.array([[[0.2, 0.1],
+                                  [0.5, 0.3],
+                                  [0.3, 0.7]]])
+        q = K.get_value(noisy_or_layer.noise_parameter)
```
matt-gardner: `K.eval()`

src/test/python/deep_qa_test/layers/noisy_or_test.py
```diff
@@ -0,0 +1,45 @@
...
+    # pylint: disable=no-self-use
+    def test_general_case(self):
+
+        input_layer = Input(shape=(3,2,), dtype='float32', name="input")
+        axis = 2
+        noisy_or_layer = NoisyOr(axis=axis)
+        output = noisy_or_layer(input_layer)
+        model = Model([input_layer], output)
+
+        # Testing general unmasked case.
```
matt-gardner: I think this is a subset of the batched case, and you don't really need to have this test.  The batched version is good enough.

src/main/python/deep_qa/layers/noisy_or.py
bsharpataz: I think I did this to be consistent with the way Keras does Constraints.  I can certainly change it if you want, or we can leave it for consistency?

src/main/python/deep_qa/layers/noisy_or.py
bsharpataz: it works, but performance currently goes down.  I'd like to keep it around maybe as the model changes, since in theory the noise parameter really should be between 0 and 1, but if it doesn't help empirically, we shouldn't use it

src/main/python/deep_qa/layers/noisy_or.py
matt-gardner: Oh, yes, definitely.  I forgot that you were overriding a method from keras.  If you're overriding a method, you need to match its signature exactly, so you did the right thing.  It's keras' fault that the code is less readable than it should be; there's not much we can do about that.

src/main/python/deep_qa/layers/noisy_or.py
matt-gardner: Then it's best to put an option in the code, to either use the constraint or not.  Commented-out code should almost never be checked in.

src/main/python/deep_qa/layers/noisy_or.py
bsharpataz: just tried it and seems to compare the tensor to the single value of 0 and says it's false... so i think I need the zero_like (and also, down later, the ones_like)

src/main/python/deep_qa/layers/noisy_or.py
bsharpataz: it does locally, I think Nelson was saying it would break one of the backends, but I can see if it works in semaphore which checks both, right?

bsharpataz: When it was in, the theano check failed (tf passed) so I changed it back, if theano now passes we'll know that we need the K.ones_like

src/main/python/deep_qa/layers/noisy_or.py
bsharpataz: i may have lied... I think theano disliked something else, but nelson showed me how to better run tests locally (including pylint) so hopefully I'll be less annoying to you all

src/main/python/deep_qa/layers/noisy_or.py
```diff
@@ -0,0 +1,89 @@
...
+        # pylint: disable=unused-argument
+        # We do not need a mask beyond this layer.
+        # (probably right)
+        if mask is not None:
+            return K.any(mask, axis=-1)
+        return None
+
+    def call(self, x, mask=None):
+        # dim: (b, ..., num_probs, ...)
+        noisy_probs = K.eval(self.noise_parameter) * x
```
matt-gardner: What's the error that theano is giving you?  I'm pretty sure this is not the right solution to whatever the error is.

matt-gardner: You probably need a `K.expand_dims`, instead of this.

bsharpataz: it's a dimension mismatch, tf was perfectly happy with the self.noise_parameter * x, but theano doesn't like it.  I was concerned about using this, but for what it's worth, I tried it and got the same performance on the dataset with it as I did without it, and the tests pass

matt-gardner: Actually, I ran into this same problem when I was trying to implement similarity functions.  You can find my solution here: https://github.com/allenai/deep_qa/blob/ba567499184b27e6f7c1747360ab86b54c5e8277/src/main/python/deep_qa/tensors/similarity_functions/linear.py#L42-L48.

bsharpataz: I guess because it's a variable, not just a float?

bsharpataz: I can try to post the error stack/output if you want it

bsharpataz: ah, ok, let me add that in, thanks!

matt-gardner: Same performance meaning it trains just fine?  (What performance are you getting, by the way?)

If that's the case, then it's fine to leave it, but I would still hide the `K.eval()` behind a check for `K.backend() == "theano"`, and leave a note about how theano has a hard time broadcasting.

matt-gardner: I was just worried about `K.eval()` breaking the computation graph, so the weight doesn't get updated correctly.  Can you verify that the weight is the same after training with and without this in tensorflow?

src/main/python/deep_qa/layers/noisy_or.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def get_output_shape_for(self, input_shape):
+        if self.axis == -1:
+            return input_shape[0:-1]
+        return input_shape[:self.axis - 1] + input_shape[self.axis:]
+
+    @overrides
+    def compute_mask(self, inputs, mask=None):
+        # pylint: disable=unused-argument
+        # We do not need a mask beyond this layer.
```
matt-gardner: This comment is wrong, because you're returning a mask.  Just remove the comment.

src/main/python/deep_qa/layers/noisy_or.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class NoisyOr(Layer):
+    """
+    Calculate the noisy-or probability across a given tensor axis based on
+    noisy-or equation:
+        p(x) = 1 - Prod_(i=1:N)[1 - q * p(x|y_n)]
+    where q is the noise parameter
+
+    Parameters
```
matt-gardner: This is nice.  Thanks for the good documentation.  One quibble, though - can you make it match the formatting used [here](https://github.com/allenai/deep_qa/blob/ba567499184b27e6f7c1747360ab86b54c5e8277/src/main/python/deep_qa/layers/attention/gated_attention.py#L33-L52)?  That will make it play nicely with our auto-generated docs.  Also, can you ask Nelson how to add this layer to the docs?  (Or follow what you see [here](https://github.com/allenai/deep_qa/pull/191/files), and ask Nelson if you need help.  He can point you to how to see the docs generated from your PR, to make sure they're what you expect.)

src/main/python/deep_qa/layers/noisy_or.py
```diff
@@ -0,0 +1,95 @@
...
+            noise_param = self.noise_parameter
+        noisy_probs = noise_param * x
+
+        # dim: (b, ..., num_probs, ...)
+        noisy_probs = 1.0 - noisy_probs
+
+        # dim: (b, ..., ...)
+        probability_product = K.prod(noisy_probs, axis=self.axis)
+
+        return K.ones_like(probability_product) - probability_product
```
matt-gardner: Switch the `K.ones_like()` to `1.0` here, too.

src/main/python/deep_qa/layers/noisy_or.py
DeNeutoy: In this case I think it might work if you just change the shape of the weight to be `()`, rather than `(1,)` as well, might be easier.

src/main/python/deep_qa/layers/noisy_or.py
bsharpataz: I worked on this -- an am not sure it's exactly what you wanted, but the docstring looks pretty consistent with some other classes (they're not all consistent with each other).  I changed the variable inputs to the call function, essentially naming the input to try to better match the documentation as well as some of the other classes.  previously I just had 'x' and can go back to that if you want -- just let me know

src/main/python/deep_qa/layers/noisy_or.py
```diff
@@ -0,0 +1,93 @@
...
+        p += leaky_ones_mask
+        return p
+
+
+class NoisyOr(Layer):
+    r"""
+    This layer takes as input a tensor of probabilities and calculates the noisy-or probability
+    across a given axis based on the noisy-or equation:
+        :math:`p(x) = 1 - \prod_{i=1:N}(1 - q * p(x|y_n))`
+    where q is the noise parameter
```
matt-gardner: Period at the end of this.

src/main/python/deep_qa/layers/noisy_or.py
```diff
@@ -0,0 +1,93 @@
...
+
+    Inputs:
+        - probabilities, shape ``(batch, ..., N, ...)``,
+          where N is the number of y's in the above equation
+          (i.e. the number of probabilities being combined in the product),
+          in the dimension corresponding to the specified axis
```
matt-gardner: Period at the end here.  It might look ok to read now, but if someone adds another sentence in a line below this, it'll get really hard to read.  Also, could you add a sentence about masking?  I would make it something like:

```
- probabilities, shape ``(batch, ..., N, ...)``, with optional mask of the same shape, where N...
```

And similarly in the outputs:

```
- X, shape ``(batch, ..., ...)``, with an optional mask of the same shape, one less ... [then after the text you have:] The output mask is the result of `K.any()` on the input mask, along the specified axis.
```

The wording of that last one got a little awkward with all of the clauses in that sentence, but I think you get the idea.

src/main/python/deep_qa/layers/noisy_or.py
```diff
@@ -0,0 +1,93 @@
...
+        The initialization of the noise parameter.
+    noise_param_constraint : Keras Constraint, default=None
+        Optional, a constraint which would be applied to the noise parameter.
+    """
+    def __init__(self, axis=-1, name="noisy_or", param_init='uniform', noise_param_constraint=None, **kwargs):
+        self.axis = axis
+        self.param_init = param_init
+        self.supports_masking = True
+        self.name = name
+        # The noisy-or equation includes a noise parameter (q) which is learned during training
```
matt-gardner: Period.

src/main/python/deep_qa/layers/noisy_or.py
```diff
@@ -0,0 +1,93 @@
...
+        self.noise_parameter = None
+        self.noise_param_constraint = noise_param_constraint
+        super(NoisyOr, self).__init__(**kwargs)
+
+    def build(self, input_shape):
+        # Add the trainable weight variable for the noise parameter
```
matt-gardner: Period.

src/main/python/deep_qa/layers/noisy_or.py
```diff
@@ -0,0 +1,93 @@
...
+                                               name=self.name + '_noise_param',
+                                               initializer=self.param_init,
+                                               regularizer=L1L2Regularizer(l2=0.001),
+                                               constraint=self.noise_param_constraint,
+                                               trainable=True)
+        super(NoisyOr, self).build(input_shape)
+
+    def get_output_shape_for(self, input_shape):
+        if self.axis == -1:
+            return input_shape[0:-1]
```
matt-gardner: You don't need the `0` here - `return input_shape[:-1]` is fine.

src/main/python/deep_qa/layers/noisy_or.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @overrides
+    def compute_mask(self, inputs, mask=None):
+        # pylint: disable=unused-argument
+        if mask is not None:
+            return K.any(mask, axis=-1)
+        return None
+
+    def call(self, inputs, mask=None):
+        # dim: (b, ..., num_probs, ...)
```
matt-gardner: We've typically formatted shape comments either like [this](https://github.com/allenai/deep_qa/blob/ba567499184b27e6f7c1747360ab86b54c5e8277/src/main/python/deep_qa/models/multiple_choice_qa/decomposable_attention.py#L49), or like [this](https://github.com/allenai/deep_qa/blob/ba567499184b27e6f7c1747360ab86b54c5e8277/src/main/python/deep_qa/models/reading_comprehension/attention_sum_reader.py#L71).  Because the comment is just the shape, no period necessary here, but I would try to follow one of those two styles.  Either use `shape` or nothing (instead of `dim`), and expand `b` to `batch_size` or `batch size`.

src/main/python/deep_qa/layers/noisy_or.py
bsharpataz: So, thanks for this, as I was making the change I noticed that in my compute_mask the axis for K.any() was actually hard-coded to -1, so I changed that.  Please verify that what I did was right (I think it is?) and then I'll merge (assuming things pass and all).

matt-gardner: Looks good!

PR #195, opened 2017-02-13
Fix model serialization

nelson-liu:
This PR fixes (or at least tries to fix) model serialization and adds a test to ensure it works with a basic `TextTrainer` and the `AttentionSumReader`

### DIFF
--- a/src/main/python/deep_qa/models/reading_comprehension/attention_sum_reader.py
+++ b/src/main/python/deep_qa/models/reading_comprehension/attention_sum_reader.py
@@ -143,3 +143,13 @@ class AttentionSumReader(TextTrainer):
     def _set_max_lengths_from_model(self):
         self.max_sentence_length = self.model.get_input_shape_at(0)[1]
         # TODO(matt): implement this correctly
+
+    @classmethod
+    def _get_custom_objects(cls):
+        custom_objects = super(AttentionSumReader, cls)._get_custom_objects()
+        custom_objects["Attention"] = Attention
+        custom_objects["L1Normalize"] = L1Normalize
+        custom_objects["OptionAttentionSum"] = OptionAttentionSum
+        from ...layers.time_distributed_embedding import TimeDistributedEmbedding
+        custom_objects["TimeDistributedEmbedding"] = TimeDistributedEmbedding
+        return custom_objects
--- a/src/main/python/deep_qa/models/reading_comprehension/attention_sum_reader.py
+++ b/src/main/python/deep_qa/models/reading_comprehension/attention_sum_reader.py
@@ -150,6 +150,4 @@ class AttentionSumReader(TextTrainer):
         custom_objects["Attention"] = Attention
         custom_objects["L1Normalize"] = L1Normalize
         custom_objects["OptionAttentionSum"] = OptionAttentionSum
-        from ...layers.time_distributed_embedding import TimeDistributedEmbedding
-        custom_objects["TimeDistributedEmbedding"] = TimeDistributedEmbedding
         return custom_objects
--- a/src/main/python/deep_qa/models/text_classification/true_false_model.py
+++ b/src/main/python/deep_qa/models/text_classification/true_false_model.py
@@ -63,10 +63,4 @@ class TrueFalseModel(TextTrainer):
     @classmethod
     def _get_custom_objects(cls):
         custom_objects = super(TrueFalseModel, cls)._get_custom_objects()
-        from ...layers.vector_matrix_split import VectorMatrixSplit
-        custom_objects["VectorMatrixSplit"] = VectorMatrixSplit
-        from ...layers.time_distributed_embedding import TimeDistributedEmbedding
-        custom_objects["TimeDistributedEmbedding"] = TimeDistributedEmbedding
-        from ...layers.wrappers import FixedTimeDistributed
-        custom_objects["FixedTimeDistributed"] = FixedTimeDistributed
         return custom_objects
--- a/src/main/python/deep_qa/models/text_classification/true_false_model.py
+++ b/src/main/python/deep_qa/models/text_classification/true_false_model.py
@@ -59,3 +59,14 @@ class TrueFalseModel(TextTrainer):
     @overrides
     def _set_max_lengths_from_model(self):
         self.max_sentence_length = self.model.get_input_shape_at(0)[1]
+
+    @classmethod
+    def _get_custom_objects(cls):
+        custom_objects = super(TrueFalseModel, cls)._get_custom_objects()
+        from ...layers.vector_matrix_split import VectorMatrixSplit
+        custom_objects["VectorMatrixSplit"] = VectorMatrixSplit
+        from ...layers.time_distributed_embedding import TimeDistributedEmbedding
+        custom_objects["TimeDistributedEmbedding"] = TimeDistributedEmbedding
+        from ...layers.wrappers import FixedTimeDistributed
+        custom_objects["FixedTimeDistributed"] = FixedTimeDistributed
+        return custom_objects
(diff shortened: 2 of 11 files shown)


src/main/python/deep_qa/training/trainer.py
```diff
@@ -12,10 +12,65 @@
 from ..common.params import get_choice
 from ..data.dataset import Dataset
 from ..data.instances.instance import Instance
-from ..layers.wrappers import OutputMask
 from .models import DeepQaModel
 from .optimizers import optimizer_from_params
 
+# custom layers for use in _get_custom_objects
+from ..layers.adaptive_recurrence import AdaptiveRecurrence
```
matt-gardner: I would say that it's each class's responsibility to ensure that the layers it uses get into `get_custom_objects`.  It's not reasonable to require that `Trainer` knows about all implemented layers - that breaks encapsulation, and probably leads to circular imports.

src/main/python/deep_qa/training/trainer.py
```diff
@@ -518,6 +573,7 @@ def _save_best_model(self):
         epoch_weight_file = "%s_weights_epoch=%d.h5" % (self.model_prefix, self.best_epoch)
         final_weight_file = "%s_weights.h5" % self.model_prefix
         copyfile(epoch_weight_file, final_weight_file)
+        print("Saved the best model to {}".format(final_weight_file))
```
matt-gardner: Use `logger` instead of `print`.

src/main/python/deep_qa/training/trainer.py
```diff
@@ -554,4 +610,55 @@ def _get_custom_objects(cls):
         If you've used any Layers that Keras doesn't know about, you need to specify them in this
         dictionary, so we can load them correctly.
         """
-        return {}
+        return {
+                "DeepQaModel": DeepQaModel,
```
matt-gardner: This is the only line you should keep; the rest should go in whatever class actually uses these layers.

src/test/python/deep_qa_test/models/reading_comprehension/attention_sum_reader_test.py
```diff
@@ -35,5 +37,13 @@ def test_train_does_not_crash(self):
...
                 "embedding_size": 5,
         }
-        solver = get_model(AttentionSumReader, args)
-        solver.train()
+        model = get_model(AttentionSumReader, args)
+        model.train()
+
+        # load the model that we serialized
+        loaded_model = get_model(AttentionSumReader, args)
+        loaded_model.load_model()
```
matt-gardner: Great, thanks for writing this test, and I'm glad it actually works.  I'm surprised it wasn't more work to make this work right.  Will the model files actually get deleted after this test?

src/main/python/deep_qa/training/trainer.py
nelson-liu: does logger get suppressed? it didn't come up when i was trying to run the test.

src/test/python/deep_qa_test/models/reading_comprehension/attention_sum_reader_test.py
nelson-liu: yeah, i was pretty happy that it wasn't as arduous a task as I thought it would be. the model files do seem to get deleted after the test.

src/main/python/deep_qa/training/trainer.py
matt-gardner: You need [these lines](https://github.com/allenai/deep_qa/blob/d4157d35d69dd7df5beece83bec58f64aca5a0f3/src/test/python/deep_qa_test/models/reading_comprehension/bidirectional_attention_test.py#L16-L18), which should go into a `DeepQaTestCase` super class at some point.

nelson-liu:
There's also a bug still on the loose somewhere with regards to actually saving the best epoch --- sometimes when i run experiments it fails to save the model (complains that it cant find the file), but I haven't tried to reproduce it yet, since it goes away after I run the model again.

src/test/python/deep_qa_test/models/reading_comprehension/attention_sum_reader_test.py
matt-gardner: I think there are some things that definitely won't work right now, but at least if you just want to load and predict, things should work.  That's good to know.

nelson-liu:
i only fix model serialization for the TrueFalseModel / the ASReader. It's a bit too much for me at the moment to fix all the models, but I don't think it would be hard to do on a case-by-case basis as we need them. Depending on whether this or #187 gets merged first, i'll also fix serialization for the GAReader

src/main/python/deep_qa/training/trainer.py
```diff
@@ -12,65 +12,10 @@
 from ..common.params import get_choice
 from ..data.dataset import Dataset
 from ..data.instances.instance import Instance
+from ..layers.wrappers import OutputMask
```
matt-gardner: Do you need this import?

src/main/python/deep_qa/models/reading_comprehension/attention_sum_reader.py
```diff
@@ -143,3 +143,13 @@ def _set_max_lengths(self, max_lengths: Dict[str, int]):
...
         self.max_sentence_length = self.model.get_input_shape_at(0)[1]
         # TODO(matt): implement this correctly
+
+    @classmethod
+    def _get_custom_objects(cls):
+        custom_objects = super(AttentionSumReader, cls)._get_custom_objects()
+        custom_objects["Attention"] = Attention
+        custom_objects["L1Normalize"] = L1Normalize
+        custom_objects["OptionAttentionSum"] = OptionAttentionSum
+        from ...layers.time_distributed_embedding import TimeDistributedEmbedding
```
matt-gardner: Why is this imported here?  If you don't need it above, why do you need it in `custom_objects`?

src/main/python/deep_qa/training/trainer.py
nelson-liu: yup, https://github.com/allenai/deep_qa/pull/195/files#diff-d000cf08b260c20e46aa5a98fa72557fR418

src/main/python/deep_qa/models/reading_comprehension/attention_sum_reader.py
nelson-liu: good question, i'm not too sure either --- the model import complains that it needs it, though.

src/main/python/deep_qa/training/trainer.py
matt-gardner: Oh, yeah, for debug models.

src/main/python/deep_qa/models/text_classification/true_false_model.py
```diff
@@ -59,3 +59,14 @@ def _instance_type(self):
     @overrides
     def _set_max_lengths_from_model(self):
         self.max_sentence_length = self.model.get_input_shape_at(0)[1]
+
+    @classmethod
+    def _get_custom_objects(cls):
+        custom_objects = super(TrueFalseModel, cls)._get_custom_objects()
+        from ...layers.vector_matrix_split import VectorMatrixSplit
```
matt-gardner: These should also go in `TextTrainer`.

nelson-liu:
@matt-gardner want to take one final look before merging?

src/main/python/deep_qa/training/text_trainer.py
```diff
@@ -579,4 +579,11 @@ def _get_custom_objects(cls):
         for value in encoders.values():
             if value.__name__ not in ['LSTM']:
                 custom_objects[value.__name__] = value
+        custom_objects["TimeDistributedEmbedding"] = TimeDistributedEmbedding
+
+        # These are used in the words_and_characters tokenizer.
```
matt-gardner: You could leave a note here that we might consider making the Tokenizer API return custom objects.

nelson-liu:
done, will merge when CI is green

PR #196, opened 2017-02-14
More minor fixes to make running BiDAF work

matt-gardner:
The main fix here is in the direction that character span instances are padded from.  That fix made the model actually train correctly, and now it looks like performance is within a few percent of the original BiDAF implementation.  There are also some other more minor things.


src/main/python/deep_qa/data/instances/question_passage_instance.py
```diff
@@ -99,7 +99,8 @@ def pad(self, max_lengths: Dict[str, int]):
         max_lengths_tmp['word_sequence_length'] = max_lengths_tmp['num_question_words']
         self.question_indices = self.pad_word_sequence(self.question_indices, max_lengths_tmp)
         max_lengths_tmp['word_sequence_length'] = max_lengths_tmp['num_passage_words']
-        self.passage_indices = self.pad_word_sequence(self.passage_indices, max_lengths_tmp)
+        self.passage_indices = self.pad_word_sequence(self.passage_indices, max_lengths_tmp,
```
nelson-liu: im not sure how i feel about this, since it affects models like the GAReader / ASReader. is there anyway you can change this just for BiDAF?

matt-gardner: Yeah, I was going to fix those tests today, but had too many meetings.  I'll do this soon.  And I think this is the right decision for all of those models - if you're truncating the passage, you should keep the beginning of the passage (for WDW, e.g., the correct entity is biased towards appearing early in the document).

PR #197, opened 2017-02-14
Optimize batch_dot layer

nelson-liu:
Sorry, in my excitement to merge the batch_dot layer I forgot to check if we could save an allocation via TF/Theano broadcasting. Turns out we can, thanks for the suggestion @matt-gardner

### DIFF
--- a/src/main/python/deep_qa/layers/backend/batch_dot.py
+++ b/src/main/python/deep_qa/layers/backend/batch_dot.py
@@ -113,8 +113,6 @@ class BatchDot(Layer):
             mask_b = K.sum(K.ones_like(tensor_b), axis=-1)
         float_mask_a = K.cast(mask_a, "float32")
         float_mask_b = K.cast(mask_b, "float32")
-        mask_a_dot_axis = K.ndim(mask_a) - 1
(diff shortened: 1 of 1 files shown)


nelson-liu:
looks semaphore got stuck --- @matt-gardner could you restart the build when you have a chance?

src/main/python/deep_qa/layers/backend/batch_dot.py
```diff
@@ -123,24 +121,18 @@ def compute_mask(self, inputs, mask=None):
                                      axes=(a_dot_axis, b_dot_axis))
         elif a_dot_axis < b_dot_axis:
             # tensor_a has less dimensions than tensor_b.
-            # We tile tensor_a to have the same shape as tensor_b
+            # We would tile tensor_a to have the same shape as tensor_b,
```
matt-gardner: I don't think you need to reference using tile in the comment.  You could just give a shape comment here, and note that theano/tensorflow will handle broadcasting the last dimension in the multiplication below.

nelson-liu: hmm i prefer to leave it here, just in case the behavior changes under our feet in the future (and it's thus easier to figure out what went wrong)

PR #198, opened 2017-02-16
tuple_match layer

bsharpataz:
Today I added mask handling, and wrote a test for it.  I am not 110% sure it's what is needed, so please review.


src/main/python/deep_qa/layers/entailment_models/tuple_match.py
```diff
@@ -0,0 +1,180 @@
...
+
+    @overrides
+    def compute_mask(self, inputs, input_mask=None):  # pylint: disable=unused-argument
+        # Here, input_mask is optionally a list of two masks (corresponding to the list of inputs) or None.
+        # Each of the masks in input_mask has shape: (batch size, num_slots)
+        # We want to see if either of the inputs is all padding (i.e. the mask would be all 0s), if so, then the
+        # whole tuple_match should be masked, so we would return a 0, otherwise we return a 1.  As such, the
+        # shape of the returned mask is (batch size, 1)
+        if any(input_mask):
+            mask1 = input_mask[0]
```
matt-gardner: You can just do `mask1, mask2 = input_mask`.

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
```diff
@@ -0,0 +1,180 @@
...
+            mask2 = input_mask[1]
+            mask1 = K.cast(K.any(mask1), 'int32')
+            mask2 = K.cast(K.any(mask2), 'int32')
+            return K.cast((mask1 + mask2) >= 2, 'int32')
```
matt-gardner: I don't think you need the cast on this return value.

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
```diff
@@ -0,0 +1,180 @@
...
+
+        return None
+
+    def call(self, inputs, mask=None):
+        tuple1_input = inputs[0]    # shape: (batch size, num_slots, num_slot_words_t1)
```
matt-gardner: Here also, you can just do `tuple1_input, tuple2_input = inputs`.  Except you should probably use `x` instead of `inputs`, even though x isn't a great name, to match the keras API.  There are some situations in which it actually matters that the names of the arguments match the super class method.

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
```diff
@@ -0,0 +1,180 @@
...
+        tuple2_input = inputs[1]    # shape: (batch size, num_slots, num_slot_words_t2)
+        # Check that the tuples have the same number of slots.
+        assert K.int_shape(tuple1_input)[1] == K.int_shape(tuple2_input)[1]
+
+        # Expand tuple1 to shape: (batch size, num_slots, num_slot_words_t1, num_slot_words_t2)
+        expanded_tuple1 = K.expand_dims(tuple1_input, 3)    # now (b, num_slots, num_slot_words_tuple1, __ )
```
matt-gardner: You can use `1` instead of `__` here, to match what you would see if you actually printed out `K.int_shape(tensor)`.

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
```diff
@@ -0,0 +1,180 @@
...
+
+        # This generates a binary tensor of the same shape as tiled_tuple1 /
+        # tiled_tuple2 that indicates if given word matches between tuple1 and tuple2 in a particular slot.
+        # Currently, we only consider S_t1 <--> S_t2 etc overlap, not across slot types.
+        # shape: (batch size, num_slots, numslot_words_tuple1, num_slot_words_tuple2)
+        tuple_words_overlap = K.cast(K.equal(tiled_tuple1, tiled_tuple2), "float32")
+
+        # Exclude zeros (i.e. padded elements) from matching each other.
+        # tuple1_mask is 1 if tuple1 has a real element, 0 if it's a padding element.
+        tuple1_mask = K.cast(K.not_equal(tiled_tuple1, K.zeros_like(tiled_tuple1, dtype='float32')),
```
matt-gardner: Can you just use `tiled_tuple1 != 0`?  If it works, it'd be faster and use less memory.

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
```diff
@@ -0,0 +1,180 @@
...
+                             dtype='float32')
+        zeros_excluded_overlap = tuple_words_overlap * tuple1_mask
+
+        # Find non-padding elememts in tuple1.
+        # shape: (batch size, num_slots, num_slot_words_tuple1)
+        non_padded_tuple1 = K.cast(K.not_equal(tuple1_input, K.zeros_like(tuple1_input)), 'float32')
```
matt-gardner: Isn't this the same as `tuple1_mask`?

matt-gardner: Oh, this is the non-tiled version - you should probably put `tiled` in the name of `tuple1_mask`, so that's more clear.

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
```diff
@@ -0,0 +1,180 @@
...
+        # Count these non-padded elements to know how many words were in each slot of tuple1.
+        # shape: (batch size, num_slots)
+        num_tuple1_words_in_each_slot = K.sum(non_padded_tuple1, axis=2)
+
+        # Find the number of words that overlap in each of the slots.
+        # shape: (batch size, num_slots)
+        slot_overlap_sums = K.sum(K.sum(zeros_excluded_overlap, axis=3), axis=2)
+
+        # # Normalize by the number of words in tuple1.
+        # todo: (becky) should this be fixed to tuple1 or allowed to vary? Does switching input order work
```
matt-gardner: The right format for a todo is `TODO(becky):`

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
```diff
@@ -0,0 +1,180 @@
...
+        # for varying?
+        # This block of code prevents dividing by zero during normalization:
+        divisor = num_tuple1_words_in_each_slot
+        # If the divisor is zero at a position, we add epsilon to it.
+        is_zero_divisor = K.equal(divisor, K.zeros_like(divisor))
```
matt-gardner: Again here, can you just do `divisor == 0`?

src/test/python/deep_qa_test/layers/tuple_match_test.py
```diff
@@ -0,0 +1,76 @@
...
+        model = Model([self.tuple1_input, self.tuple2_input], output)
+
+        # Get the initial weights for use in testing
+        dense1_hidden_weights = K.eval(model.trainable_weights[0])
+        dense1_biases = K.eval(model.trainable_weights[1])
+        score_hidden_weights = K.eval(model.trainable_weights[2])
+        score_bias = K.eval(model.trainable_weights[3])
+
+        # Testing general unmasked case.
+        desired_overlap = np.array([[3/2,
```
matt-gardner: Overlap greater than 1?  Seems a little odd.  But it's probably ok.  (Well, I guess, empirically, it's fine, so don't worry about it.)

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
```diff
@@ -0,0 +1,180 @@
...
+        entailment score.
+        Each entailment feature is currently made by comparing the corresponding slots in the two tuples and
+        determining the degree of lexical overlap using the formula:
+            :math:`normalized overlap_s = \dfrac{|A_s \cap B_s|}{|A_s|}`
+        where s is the index of the slot, :math:`A_s` is answer tuple slot s and :math:`B_s` is background
+        tuple slot s.
+
+        Inputs:
+            - tuple_1_input (the answer tuple), shape ``(batch size, num_slots, num_slot_words_t1)``,
+              with an optional mask of shape (batch size, num_slots).  Here num_slot_words_t1 is the maximum
```
matt-gardner: You actually won't get a mask here, and if for some crazy reason you're passed a mask, you actually ignore it.  You calculate your masking based on the input being 0, because you're working directly with word indices.  So, I would change this to say that any input mask is ignored.

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
```diff
@@ -0,0 +1,180 @@
...
+        self.score_layer = score_layer_dense
+
+        # Add the weights from these layers to TupleMatch's trainable weights
+        for hidden in self.hidden_layers + [self.score_layer]:
+            self.trainable_weights.extend(hidden.trainable_weights)
+
+        super(TupleMatch, self).build(input_shape)
+
+    @overrides
+    def compute_mask(self, inputs, input_mask=None):  # pylint: disable=unused-argument
```
matt-gardner: The second argument needs to be `input`, not `inputs`.

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
matt-gardner: But, actually, this computation isn't correct.  You want to do the `K.anys` that you have, but on the input, not on the mask.  That is, if any _words_ are non-padding, you return a 1, otherwise, you return a 0.

src/test/python/deep_qa_test/layers/tuple_match_test.py
```diff
@@ -0,0 +1,76 @@
...
+        # Desired_overlap gets fed into the inner NN.
+        dense1_activation = np.dot(desired_overlap, dense1_hidden_weights) + dense1_biases
+        final_score = np.dot(dense1_activation, score_hidden_weights) + score_bias
+        desired_result = logistic.cdf(final_score)
+        result = model.predict([self.tuple1, self.tuple2])
+        assert_array_almost_equal(result, desired_result)
+
+    def test_masks_handled_correctly(self):
+        # tuple1 has shape (b, num_slots, t1_words_per_slot), so mask1 has shape (b, num_slots)
+        mask1 = K.variable(np.array([1, 1, 0]))
```
matt-gardner: You want this to take tuple inputs, with the mask being `None`, instead of specifying a mask.

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
```diff
@@ -0,0 +1,180 @@
...
+        entailment_features = number of tuple slots), and then passing these features into a shallow NN to get an
+        entailment score.
+        Each entailment feature is currently made by comparing the corresponding slots in the two tuples and
+        determining the degree of lexical overlap using the formula:
+            :math:`normalized overlap_s = \dfrac{|A_s \cap B_s|}{|A_s|}`
+        where s is the index of the slot, :math:`A_s` is answer tuple slot s and :math:`B_s` is background
+        tuple slot s.
+
+        Inputs:
+            - tuple_1_input (the answer tuple), shape ``(batch size, num_slots, num_slot_words_t1)``,
```
bsharpataz: in current experiments, I vary it, but the idea is that since they are formed using different sources they could be diff amts of noisy and so have diff allowed lengths... I could drop this if you want

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
bsharpataz: ok, agreed for now -- but at some point very soon I will be adding embeddings (which are masked), so should I have the current PR say ignored and deal with it later when I add embeddings?

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
```diff
@@ -0,0 +1,180 @@
...
+              maximum number of words in each of the slots in tuple_2. This need not match tuple 1.
+
+        Output:
+            - entailment score, shape ``(batch, 1)``
+
+        Parameters
+        ----------
+        params: dict
+            Specifies each of the following (with defaults if unspecified):
+            - num_entailment_features : int, default=4
```
bsharpataz: yes...... to build things doesn't it need to know it ahead of time? or no?

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
matt-gardner: Do you have an `Instance` that has different padding here?  It's fine if you do; feel free to represent the data in the way you think is best.  I wouldn't have thought to make these different, but I don't have any complaints about it, other than a slight increase in complexity.

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
matt-gardner: We need to keep this layer around as is, so that we can compare against it once we have a version with embeddings.  So, yes, this class should say that the mask is ignored, and you should make a _new_ class that adds embeddings.  I can help set that up when it's time to do it, if you want.

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
matt-gardner: I think this is `min(num_slot_words1, num_slot_words2)`.  Both of those are available to you from `input_shape` in the `build()` method, where your current `num_entailment_features` is used, so you can probably just remove this parameter and compute it in the `build` method.

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
```diff
@@ -0,0 +1,180 @@
...
+    def build(self, input_shape):
+        # Build the dense layer(s)
+        self.hidden_layers = []
+        dense_input_shape = (None, self.num_entailment_features)
+        for i in range(self.num_hidden_layers):
+            curr_dense = Dense(output_dim=self.hidden_layer_width,
+                               activation=self.hidden_layer_activation,
+                               bias=self.use_biases,
+                               name='tuple_match_hidden_layer_%d' % i)
+            curr_dense.build(input_shape=dense_input_shape)
```
bsharpataz: re: first comment, that's what I want, the inputs shape to the second should be the output of the first
re: second comment -- seems like the layers you're adding are all of the type of the main class? (the self.init and all) whereas here I am adding in Dense layers.  That said, if I don't have to  build the layers, it gets cleaner

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
bsharpataz: nope, I just tried it and it doesn't work (which is probably why option sum does it this way too)

src/test/python/deep_qa_test/layers/tuple_match_test.py
bsharpataz: yeah -- can seem weird, but if a question word is repeated several times...

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
bsharpataz: I changed it

bsharpataz: oh, i mean, I changed it from the first comment -- it isn't based on the num of words in slots, but rather the number of slots, so input_shape[0][1]

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
bsharpataz: pylint doesn't like this, says I am "Redefining built-in" -- want me to locally override pylint to be consistent with Keras?

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
bsharpataz: re: second comment again -- sorry, I really don't understand what's happening in the code you pointed me to... if you want me to use it, can you direct me a little? (sorry!)

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
matt-gardner: Weird.  It seemed like it worked on your constraint code.  Ok.

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
matt-gardner: Yeah, it's an awful decision that keras made.  But we have to follow that decision, or our code will break.  I ran into a crash because of this a while ago.  It's really unfortunate.

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
matt-gardner: Sorry, I should have been more clear.  Instead of using an internal `Dense` layer, you could just create the weights yourself, as done with the calls to `self.init()` in the code I linked (or you could just use `self.add_weight`, as you did in your last PR, which is probably cleaner; I didn't know about that earlier, or I would have recommended it to Pradeep instead of his `self.init()` code).  If you just call `add_weight` instead of creating a `Dense` layer, you can then use our `apply_feed_forward` method to run your tensor through all of the hidden layers.  I can show you how to do this, if you want.

If you do it that way, you don't have to worry about building an internal layer, which can cause subtle bugs.  It probably wouldn't be a problem with `Dense`, but doing it the way you did doesn't handle masking, so if you get into the habit of this, you could expect a behavior that you're not getting with a more complicated layer.  An `Embedding` layer, for instance, probably would do weird things.

matt-gardner: And, for my first comment, I was wrong.  I missed line 99, where you reset `dense_input_shape`.  Sorry.

matt-gardner: And I now see another confusing thing about the initialization code I pointed you to - he's calling [`self.init`](https://github.com/allenai/deep_qa/blob/cdd989046619cc563f0676b6f1c9df8201e6506d/src/main/python/deep_qa/layers/entailment_models/decomposable_attention.py#L39), not `self.__init__` - he defined a method to initialize weights, he's not constructing recursive objects.  Confusing name, looks like.

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
matt-gardner: Oh, right, but don't you still need a min in there somewhere?  Oh, nevermind, the number of slots is the same in both sets of tuples.  Right.  Sorry... =)

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
matt-gardner: So, yes, locally override pylint.

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
bsharpataz: where in this code do the weights get an activation?  or do I wrap an activation around the initialization?

matt-gardner: [`apply_feed_forward`](https://github.com/allenai/deep_qa/blob/cdd989046619cc563f0676b6f1c9df8201e6506d/src/main/python/deep_qa/tensors/backend.py#L121) takes an activation.  You'd have to create it with Keras' [`activations.get`](https://github.com/allenai/deep_qa/blob/cdd989046619cc563f0676b6f1c9df8201e6506d/src/main/python/deep_qa/layers/entailment_models/decomposable_attention.py#L119).

bsharpataz: ok, at the moment there are two different activations, per Tushar's code, a tanh on the middle layers and a sigmoid on the final layer.  Want me to keep things flexible as is, or switch to just sigmoids all throughout?

bsharpataz: I hear you about subtle bugs with using the Dense layers, though, I just don't know how you want me to proceed...

matt-gardner: Sorry if it feels like I'm telling you what to do here.  You should take my comments as suggestions that would help the code be more readable and maintainable, and an attempt to find things that might be errors.  If you want to do things differently than I suggest, that's fine; it's your code.

If you want to follow this suggestion, what I would probably do is have a bunch of hidden layers that do whatever activation you want (tanh in this case, sounds like), then have a final layer that has a different activation.  Pradeep's code does this, too: https://github.com/allenai/deep_qa/blob/cdd989046619cc563f0676b6f1c9df8201e6506d/src/main/python/deep_qa/layers/entailment_models/decomposable_attention.py#L145-L148.

bsharpataz: yeah, shoot  - you're right. sorry I didn't see that in the code.  Ok, fixing now, then it should be all done

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
```diff
@@ -0,0 +1,156 @@
...
+    - hidden_layer_activation : string, default='relu'
+        The activation of the NN hidden layers
+    - final_activation : string, default='sigmoid'
+        The activation of the NN output layer
+    """
+
+    def __init__(self, num_hidden_layers: int=1, hidden_layer_width: int=4,
+                 initialization: str='glorot_uniform', hidden_layer_activation: str='relu',
+                 final_activation: str='sigmoid'):
+        """
```
matt-gardner: This docstring isn't really adding anything, and the parameters are covered in the class docstring anyway.  I'd probably remove this.

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
```diff
@@ -0,0 +1,156 @@
...
+        return (input_shapes[0], 1)
+
+
+    def build(self, input_shape):
+        super(TupleMatch, self).build(input_shape)
+
+        # Add the weights for the hidden layers
+        hidden_layer_input_dim = input_shape[0][1]
+        for i in range(self.num_hidden_layers):
+            self.hidden_layer_weights.append(self.hidden_layer_init((hidden_layer_input_dim,
```
matt-gardner: I kind of like using `self.add_weight`, like you did in the other PR.  Feel free to leave it like this, though.  It's up to you.

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
```diff
@@ -0,0 +1,156 @@
...
+                                                                     self.hidden_layer_width),
+                                                                    name='%s_hiddenlayer_%d' % (self.name, i)))
+            hidden_layer_input_dim = self.hidden_layer_width
+        self.trainable_weights = self.hidden_layer_weights[:]
```
matt-gardner: You don't need the `[:]`.  Oh, yes you do, otherwise you'd be appending `self.score_layer` to it...  This is a pretty strong argument in favor of using `self.add_weight`, I think.  Less prone to errors like that (not that your code is wrong, just that I would have written it wrong, and I almost suggested to change it to something that's wrong).

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
```diff
@@ -0,0 +1,156 @@
...
+        # Add the weights for the final layer.
+        self.score_layer = self.hidden_layer_init((self.hidden_layer_width, 1), name='%s_score' % self.name)
+        self.trainable_weights.append(self.score_layer)
+
+
+
+    @overrides
+    def compute_mask(self, input, input_mask=None):  # pylint: disable=unused-argument,redefined-builtin
+        # Here, input_mask is ignored. To determine the returned mask, we want to see if either of the inputs
```
matt-gardner: I'd add `, because the input is plain word tokens` to the first sentence here, just so it's clear _why_ the mask is ignored.

nelson-liu:
(it's failing tests right now because i have yet to merge #167 , will rerun when i merge that in a bit)

matt-gardner:
Ok, I'll just ignore travis for this PR, and use the semaphore result.  I'll remove semaphore after my PR removing the scala code goes through.

src/main/python/deep_qa/layers/entailment_models/tuple_match.py
bsharpataz: agreed, and the first time I wrote it it was wrong :)

nelson-liu:
i merged master into this to get the latest config changes from travis

matt-gardner:
The semaphore failure seems pretty flaky to me; not sure what's going on there.  But travis passed, so I'm going to merge this now.

PR #201, opened 2017-02-16
Made self._get_encoder() work with no arguments

matt-gardner:
If you just want to call `self._get_encoder()`, and you don't specify a `default` key in the `encoder` params, the code would crash before this PR, even though we know how to handle a default encoder without any parameters.


nelson-liu:
Maybe add a simple test? Sorry for the hassle

matt-gardner:
The test fails on master, passes on this branch.

PR #203, opened 2017-02-17
Removed scala code, moved python code to standard places

matt-gardner:
This is huge.  I would recommend just browsing the code corresponding to this commit, instead of looking through the "files changed".


matt-gardner:
The place to browse the code is here: https://github.com/matt-gardner/deep_qa/tree/remove_scala_code.  I'll worry about fixing the travis config soon.

nelson-liu:
looks like you need to update `pytest.ini` though

matt-gardner:
I don't see `deep_qa/project` or `deep_qa/target` - what do you mean?

nelson-liu:
maybe its just me but when i pull down the PR locally i see those two folders --- it doesnt show up on your branch so i wouldn't worry

nelson-liu:
oh duh those are gitignored files used by sbt

PR #205, opened 2017-02-17
Made docs build correctly again, updated main doc page

matt-gardner:


matt-gardner:
@nelson-liu, merging this now, so that our docs work again.  The one thing I thought might actually need a review is removing the `make.bat` file.  I'm pretty sure we don't need it, but if there's an issue, say so, and we can revisit this PR (or submit a new one to undo what I'm merging now).

nelson-liu:
Hm interesting, surprised that the builds didn't fail. Thanks!

PR #207, opened 2017-02-17
Multi input time distributed

bsharpataz:
I'd ___love___ you to:

- check that the tests for the call method are sufficient to really make sure things are ok

- check the compute_mask method -- I tried to borrow from the TD with Mask class and extend it parallel to how Mark helped me extend the call method of the multi Input one

- help me write a test for the compute_mask (right now, it's a bunch of copied code from other tests that I half modified before quitting)

- make sure I didn't ruin things with the move to wrappers/ with separate files...


deep_qa/layers/wrappers/debug_output_mask.py
```diff
@@ -0,0 +1,16 @@
+from keras.layers import Layer
```
matt-gardner: No need to put `debug` in the name.  Just call this `output_mask.py`.

deep_qa/layers/wrappers/encoder_wrapper.py
```diff
@@ -0,0 +1,26 @@
+from keras import backend as K
+from deep_qa.layers.wrappers.fixed_time_distributed import FixedTimeDistributed
+
+
```
matt-gardner: Only two blank lines here.

deep_qa/layers/wrappers/fixed_time_distributed.py
```diff
@@ -0,0 +1,37 @@
+from keras import backend as K
+from keras.layers import TimeDistributed
+
```
matt-gardner: Two blank lines here.  (https://www.python.org/dev/peps/pep-0008/#blank-lines)

deep_qa/layers/wrappers/fixed_time_distributed.py
```diff
@@ -0,0 +1,37 @@
...
+class FixedTimeDistributed(TimeDistributed):
```
matt-gardner: I think we should call this `TimeDistributed`, so that if/when Keras fixes its class, we just have to change an import in the model code (does that make sense?).  I've been meaning to do this myself, but haven't yet.  So can you do the import above as `from keras.layers import TimeDistributed as KerasTimeDistributed`, and call this class `TimeDistributed`?  This is similar to how we did the [`Highway`](https://github.com/allenai/deep_qa/blob/47f38139f31b216c00762d28f1786d65dd68a9ee/deep_qa/layers/highway.py) layer.  And then you'd want to change the name of the file, too.

deep_qa/layers/wrappers/multi_input_time_distributed.py
```diff
@@ -0,0 +1,161 @@
+from keras import backend as K
+from keras.layers import InputSpec, TimeDistributed
+from deep_qa.layers.wrappers.fixed_time_distributed import FixedTimeDistributed
+from overrides import overrides
+
+
+class MultiInputTimeDistributed(FixedTimeDistributed):
+    r"""
```
matt-gardner: You don't need this `r` here if you're not doing math.

deep_qa/layers/wrappers/multi_input_time_distributed.py
```diff
@@ -0,0 +1,161 @@
...
+    This class extends the previous TimeDistributed fix (i.e., ``FixedTimeDistributed``)
```
matt-gardner: Just say `This class extends ``TimeDistributed`` to allow...`

deep_qa/layers/wrappers/multi_input_time_distributed.py
```diff
@@ -0,0 +1,161 @@
...
+    to allow for inputs that are lists of tensors.  Currently, while masking is supported masking (i.e., if any of the inputs
```
matt-gardner: I don't understand what this second sentence is saying.

deep_qa/layers/wrappers/multi_input_time_distributed.py
```diff
@@ -0,0 +1,161 @@
...
+    has a mask) is not supported.
+
+    Inputs:
+        - probabilities, shape ``(batch, ..., N, ...)``,
```
matt-gardner: Looks like the rest of this docstring needs updating.

deep_qa/layers/wrappers/multi_input_time_distributed.py
```diff
@@ -0,0 +1,161 @@
...
+    param_init : string, default='uniform'
+        The initialization of the noise parameter.
+    noise_param_constraint : Keras Constraint, default=None
+        Optional, a constraint which would be applied to the noise parameter.
+    """
+
+    @overrides
+    def build(self, input_shape):
+        num_inputs = len(input_shape)
+        input_specs = []
```
matt-gardner: You can just do `self.input_spec = []` here, and avoid the assignment on line 42.

deep_qa/layers/wrappers/multi_input_time_distributed.py
```diff
@@ -0,0 +1,161 @@
...
+        Name of the layer, ued to debug both the layer and its parameter.
+    param_init : string, default='uniform'
+        The initialization of the noise parameter.
+    noise_param_constraint : Keras Constraint, default=None
+        Optional, a constraint which would be applied to the noise parameter.
+    """
+
+    @overrides
+    def build(self, input_shape):
+        num_inputs = len(input_shape)
```
matt-gardner: You should check for tuple vs. list here, like you do in `get_output_shape_for`.

deep_qa/layers/wrappers/multi_input_time_distributed.py
```diff
@@ -0,0 +1,161 @@
...
+            for i in range(num_inputs):
+                child_input_shape = (input_shape[i][0],) + input_shape[i][2:]
+                child_output_shape = self.layer.get_output_shape_for(child_input_shape)
+                output_shapes.append((child_input_shape[0], timesteps) + child_output_shape[1:])
+            return output_shapes[0]
+        else:
+            raise Exception("Invalid input_shape type.")
+
+    @overrides
+    def compute_mask(self, x, input_mask=None):
```
matt-gardner: Actually, I think I would just recommend punting on this for now, because we should put the `call` and other code directly in our `TimeDistributed` class.  Then each of our subclasses that handle masks differently can incorporate the multi-input into their `compute_mask` methods.  The trouble with having a single implementation here is that different subclasses do different mask computations, so you can't really have a single general method.

If you want to include this in this PR, instead of punting on it, you can probably move what you have here to the `TimeDistributedWithMask` class.

matt-gardner:
Thanks for working on this, it's a really helpful change.

matt-gardner:
I'm going to try to finish this now, because it'd be easier for me than for you, and so you can get back to doing model design sooner.  If I finish this fast enough, I'll work on the suggestions I made to your other PR, too.  You've gotten them most of the way there, so it shouldn't take me too long to finish it.

bsharpataz:
sorry I didn't do this yet -- just wasn't working on ai2 stuff today

matt-gardner:
No need to apologize for not working on a Saturday.  It's not expected at all.  I'm just trying to help you get back to stuff that's more interesting for you.  Improving the library probably isn't the most exciting thing for you to be working on.

matt-gardner:
Ok, I think this is done.  It turned out it was more complicated than I originally thought it was (which makes me happy that I could do it for you, because it would have taken quite a bit of back-and-forth to figure out the issues and the solutions to them, and that would have just been a distraction for your internship project).  But it should work now; you should be able to just use our `TimeDistributed` class on multiple inputs, with masks, and it should just work.

matt-gardner:
And I'm going to add you as a reviewer to the PR - can you look over the changes I made?

bsharpataz:
thanks again for all your help on this!  looks great to me, and seems like you cleaned it all up by combining use cases, but then again, take my input with the appropriately sized grain of salt!

PR #208, opened 2017-02-17
turn off the codecov notifications

nelson-liu:
hopefully this works, might have to tweak it a bit


nelson-liu:
looks like this turned off the commenting (which is what we intended), but the github status checks still aren't showing up. merging for now.

PR #212, opened 2017-02-18
WordProcessor (terrible name)

bsharpataz:
added an intermediate step in the splitting-tokenization pipeline (WordProcessor).  This required adding a tiny bit of extra code to word_tokenizer and word_character_tokenizer, but should be backwards compatible.  Please check though.  When this PR is ready, I can update the docs (didn't want to do that until the structure was finalized).


deep_qa/data/tokenizers/word_splitter.py
```diff
@@ -1,17 +1,19 @@
...
 from overrides import overrides
 
-class WordSplitter:
-    """
-    A WordSplitter splits strings into words.  This is typically called a "tokenizer" in NLP, but
-    we need "Tokenizer" to refer to something else, so we're using WordSplitter here instead.
+
+class WordSplitter():
     """
+        A WordSplitter splits strings into words.  This is typically called a "tokenizer" in NLP, but
```
matt-gardner: Spacing is off here.

deep_qa/data/tokenizers/word_splitter.py
```diff
@@ -1,17 +1,19 @@
...
 
+
 from overrides import overrides
 
-class WordSplitter:
-    """
-    A WordSplitter splits strings into words.  This is typically called a "tokenizer" in NLP, but
-    we need "Tokenizer" to refer to something else, so we're using WordSplitter here instead.
+
+class WordSplitter():
```
matt-gardner: No need for the parentheses here.

deep_qa/data/tokenizers/word_splitter.py
matt-gardner: Actually, I don't think you made any material changes to this file - you can just revert the whole file.

deep_qa/data/tokenizers/word_processor.py
```diff
@@ -0,0 +1,72 @@
...
+
+class StemmedFilteredWordProcessor(WordProcessor):
+    """
+    The StemmedFilteredWordProcessor splits the string into words using the specified WordSplitter and
+    then filters out stop words and stems the remaining words.
+    """
+    def __init__(self, word_splitter_choice: str):
+        self.stemmer = PorterStemmer()
+        # TODO (becky): load in the proper way at some point.
+        self.stopwords = ['I', 'a', 'aboard', 'about', 'above', 'accordance', 'according', 'across', 'after',
```
matt-gardner: Where did this list come from?  From Tushar?  There are some odd decisions (e.g., "true", "prepatory", "irrespective").

deep_qa/data/tokenizers/word_tokenizer.py
```diff
@@ -19,7 +20,8 @@ class WordTokenizer(Tokenizer):
     """
     def __init__(self, params: Dict[str, Any]):
         word_splitter_choice = get_choice_with_default(params, 'word_splitter', list(word_splitters.keys()))
-        self.word_splitter = word_splitters[word_splitter_choice]()
+        word_processor_choice = get_choice_with_default(params, 'word_processor', list(word_processors.keys()))
```
matt-gardner: I think the API would be cleaner and easier to extend if the parameters we passed looked something like this:

```json
{
  "processor": {
    "splitter": {
    ...
    },
    "stemmer": {
    ...
    },
    "stopwords": {
    ...
    }
  }
}
```

The tokenizer then just has one object that it constructs (the `WordProcessor`), and it only interacts directly with that object, with the single function call `get_tokens(text: str) -> List[str]`.  The `WordProcessor` then handles splitting words and whatever stemming or stop word removal you've specified in your parameters.  Does that seem reasonable?

matt-gardner:
I'm going to make the changes I recommended here, so that you can work on your instance type PR, then the actual model.

matt-gardner:
Ok, this is done.  @bsharpataz, can you look over the changes that I made?

bsharpataz:
I like the changes, and I agree with you about splitting apart the stemming from the filtering (btw, stop word list was shown to me by tushar -- it exists in a code base somewhere I just don't know how to load it -- it's a Peter Clark list, apparently).
The thing that worries me is that if an old experiment doesn't specify "processor" in their experiment file, but they did specify perhaps the NLTK WordSplitter, wouldn't this change break their expected result? since the default would be {} for processor and then "simple" for the splitter?

matt-gardner:
You're right that this is a non-backwards-compatible API change, but I think that's ok.  There isn't anything that uses the NLTK word splitter.  Going forward, we'll probably need to think about API changes and backwards compatibility, but I don't think we're to that point yet.

bsharpataz:
ok, great, thanks

PR #213, opened 2017-02-20
MCTupleWithBackgroundInstance

bsharpataz:


deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+                source = element.split("source:")[1]
+            else:
+                objects.append(element)
+        return TextTuple(subject, verb, objects, context, time, location, source)
+
+
+class MCTuplesWithBackgroundInstance(TextInstance):
+    """
+    A MCTuplesWithBackgroundInstance is a kind of TextInstance that has text in multiple slots (stored
+    as a TextTuple), for both the answer choices and background information.
```
nelson-liu: double backticks around `TextTuple`

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+    This can be used to store SVO triples.
+
+    Parameters
+    ----------
+    - answer_tuples: List[List[TextTuple]]
```
nelson-liu: These docstrings unfortunately do not conform to the numpydoc standard, see [gatedattentionreader](https://github.com/allenai/deep_qa/blob/bbeb31a3749885824f0a0d9efbb8a7f6a2758c06/deep_qa/models/reading_comprehension/gated_attention_reader.py#L30) for an example of how it should be formatted or see the [official guide, section 4 parameters](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#sections)

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+        This is a list of length = num answer options.  Each answer has a list of TextTuples (made
```
nelson-liu: no need for the `=`. I feel like you're treating `num answer options` as a variable here --- if so, add double backticks and replace spaces with underscores. Else this might be better worded as `"The length of `answer_tuples` must be the same as the number of answer options"`.

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+        from the question+answer_candidate tuples).
```
nelson-liu: `"from the question and candidate answer tuples"`?

deep_qa/data/instances/mc_tuples_with_background_instance.py
nelson-liu: I generally think it's more legible to have types written out, so in this case `List of List of TextTuple`.

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+    """
+    def __init__(self, answer_tuples: List[List[TextTuple]], background_tuples: List[TextTuple],
+                 object_handling: str="collapse", use_context: bool=True, label: int=None, index: int=None):
+        super(MCTuplesWithBackgroundInstance, self).__init__(label, index)
+        self.answer_tuples = answer_tuples
+        self.background_tuples = background_tuples
+        self.object_handling = object_handling
+        self.use_context = use_context
+
+    def __str__(self):
```
nelson-liu: same comments on this `__str__` method as above.

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+    def to_indexed_instance(self, data_indexer: DataIndexer):
+        answers_indexed = [self.index_tuples(curr_answer_tuples, data_indexer) for
+                           curr_answer_tuples in self.answer_tuples]
+        background_indexed = self.index_tuples(self.background_tuples, data_indexer)
+        return IndexedMCTuplesWithBackgroundInstance(answers_indexed,
+                                                     background_indexed,
+                                                     self.label,
+                                                     self.index)
+
+    # This is a helped method for to_indexed_instance which indexes a list of TextTuples
```
nelson-liu: this would be better as a method docstring

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+        currently handle this, but plan to eventually replace the referring expression with its resolved
+        value.
+            Example: it[sky]|is|blue --> would become --> sky|is|blue
+
+        Some tuples additionally contain prefixed information about some of the "objects" such as "L:" for
+        locations and T: for time.  At present this information is simply filtered out, but can be used
+        later.
+        """
+        fields = line.split("\t")
+        assert len(fields) == 5, \
```
nelson-liu: i think this is better written as https://github.com/allenai/deep_qa/blob/bbeb31a3749885824f0a0d9efbb8a7f6a2758c06/deep_qa/data/instances/character_span_instance.py#L59. Does the index _have_ to be there?

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+        num_answers = int(fields[1])
+        # Answer Candidate tuples
+        answers = fields[2].split("###")
+        answer_tuples = []
+        for curr_answer in answers:
+            curr_answer_tuple_strings = curr_answer.split("$$$")
+            curr_answer_tuples = [TextTuple.from_string(tuple_string)
+                                  for tuple_string in curr_answer_tuple_strings if len(tuple_string) > 0]
+            answer_tuples.append(curr_answer_tuples)
+        assert len(answer_tuples) == num_answers, \
```
nelson-liu: prefer conditional + `ConfigurationError` over asserts

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+            "ERROR: {0} answers found, {1} specified.".format(len(answer_tuples), num_answers)
```
nelson-liu: it's unnecessary to index the replacement targets in format, but I don't mind either way.

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+        Lets the instance know if you want to retain a slot for information marked as context.  If False,
+        the context information is ignored.
+
+    - label: int, default=None
+        The class label (i.e. the index of the correct multiple choice answer) -- corresponds to the
+        indices in self.answer_tuples.
+
+    - index: int, default=None
+        The index of the question.
+
```
nelson-liu: extra whitespace

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+                             for background_string in background_tuple_strings if len(background_string) > 0]
+        # Label
+        label = int(fields[4])
+        assert label < num_answers
+
+        return cls(answer_tuples, background_tuples, label=label, index=index)
+
+
+class IndexedMCTuplesWithBackgroundInstance(IndexedInstance):
+
```
nelson-liu: extra whitespace

deep_qa/data/instances/mc_tuples_with_background_instance.py
nelson-liu: this happens a few more times but i'll stop commenting on it for the sake of brevity

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+    def __init__(self,
+                 answers_indexed: List[List[List[List[int]]]],
```
nelson-liu: yikes this seems hard to keep track of

doc/api_doc/deep_qa.data.instances.rst
```diff
@@ -57,7 +57,8 @@ deep_qa.data.instances.multiple_true_false_instance
     :undoc-members:
     :show-inheritance:
 
- deep_qa.data.instances.mc_question_answer_instance
+
```
nelson-liu: remove extra whitespace

doc/api_doc/deep_qa.data.instances.rst
```diff
@@ -66,6 +67,16 @@ deep_qa.data.instances.multiple_true_false_instance
     :undoc-members:
     :show-inheritance:
 
+
```
nelson-liu: this is extra whitespace

tests/data/instances/mc_tuples_with_background_instance_test.py
```diff
@@ -0,0 +1,284 @@
+# pylint: disable=no-self-use,invalid-name
+
+from unittest import TestCase
+import numpy as np
+
+from deep_qa.data.data_indexer import DataIndexer
+from deep_qa.data.dataset import TextDataset
+from deep_qa.data.instances.mc_tuples_with_background_instance import MCTuplesWithBackgroundInstance, \
+    IndexedMCTuplesWithBackgroundInstance
```
nelson-liu: you can just repeat the import from the line above, like `from deep_qa.data.instances.mc_tuples_with_background_instance import IndexedMCTuplesWithBackgroundInstance`. In general i feel like the `\` is inelegant, but that might just be me.

tests/data/instances/mc_tuples_with_background_instance_test.py
```diff
@@ -0,0 +1,284 @@
...
+
+
+class TestMcTuplesWithBackgroundInstance(TestCase):
+
```
nelson-liu: extra space

tests/data/instances/mc_tuples_with_background_instance_test.py
```diff
@@ -0,0 +1,284 @@
...
+    def setUp(self):
+        qIdx = 0
+        num_ans = 2
+        answerOptions = "cat|is|a mammal|context: which is a nocturnal mammal$$$cat|is|nocturnal|which " \
+                        "is a nocturnal mammal###snake|is|a mammal|which is a nocturnal " \
+                        "mammal$$$snake|is|nocturnal|which is a nocturnal mammal"
+        background = "cat|eat|food$$$cat|is|a mammal|with fur|which is a mammal$$$snake|is example " \
```
nelson-liu: minor thing, but `+` vs `\`?

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+    def empty_instance(cls):
+        return IndexedMCTuplesWithBackgroundInstance([], [], label=None, index=None)
+
+    @overrides
+    def get_lengths(self) -> Dict[str, int]:
+        # We care only about the longest slot here because all slots will be padded to the same length.
+        max_answer_length = 0
+        max_num_slots = 0
+        max_question_tuples = 0
+        for curr_answer in self.answers_indexed:
```
matt-gardner: I'm not a huge fan of prepending `curr_` to variables like this - you can just use pluralization most of the time to get a different (and easy to understand) variable name, and `curr` is a shortening of a word, which you should almost always avoid in variable names.  I would use `answer` here, and `answer_slots` and `answer_length` for the variables below.

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+        max_num_slots = max(max_num_slots, background_max_num_slots)
+
+        return {'num_options': len(self.answers_indexed),
+                'num_question_tuples': max_question_tuples,
+                'num_background_tuples': len(self.background_indexed),
+                'question_slot_length': max_answer_length,
+                'background_slot_length': background_max_slot_length,
+                'num_slots': max_num_slots}
+
+    # Helper method for self.get_lengths() -- gets the max lengths for a list of indexed tuples
```
matt-gardner: Docstring instead of comment.

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+    @staticmethod
+    def get_lengths_from_indexed_tuples(indexed_tuples: List[List[List[int]]]):
+        max_num_slots = 0
+        max_slot_length = 0
+        for indexed_tuple in indexed_tuples:
+            num_slots = len(indexed_tuple)
+            max_num_slots = max(max_num_slots, num_slots)
+            for slot in indexed_tuple:
+                max_slot_length = max(max_slot_length, len(slot))
```
matt-gardner: You need to call `self._get_word_sequence_lengths` here, to handle the case where the model using this `Instance` is using the `WordAndCharacter` tokenizer.  It'll make the return value and the logic in the method that calls this a little more complicated.

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+        num_options = len(self.answers_indexed)
+        if num_options > desired_num_options:
+            raise Exception("Case of too many answer options ({0} provided) isn't currently "
+                            "handled.".format(num_options))
+        elif num_options < desired_num_options:
+            # Add empty answer option
+            additional_options = [[] for _ in range(desired_num_options - num_options)]
+            self.answers_indexed += additional_options
+        # Pad the number of tuples per option
+        for curr_answer in range(len(self.answers_indexed)):
```
matt-gardner: Here you're using `curr_answer` to refer to an index, which is a confusing name to me.  I would call this `answer_index`, like you do below, or just do `for answer in self.answers_indexed:`, and append to or truncate `answer` directly.

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+                                                                 desired_question_slot_length)
+                                                  for answer_tuple in self.answers_indexed[answer_index]]
+        # Pad the number of background tuples
+        self.background_indexed = self.pad_num_tuples(self.background_indexed, desired_num_background_tuples)
+        # Pad each of the background tuples for number of slots and slot length
+        self.background_indexed = [self.pad_tuple(background_tuple, desired_num_slots,
+                                                  desired_background_slot_length)
+                                   for background_tuple in self.background_indexed]
+
+    # Helper method to pad a list of tuples to the desired number of tuples
```
matt-gardner: To docstring.

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+    @staticmethod
+    def pad_num_tuples(tuples_in, desired_length):
```
matt-gardner: Prefer using `self.pad_sequence_to_length` - it'll make this code shorter, re-use an already well-test piece of code, and be more flexible, because that method has more options than you have here.

matt-gardner: As in, you can literally replace all calls to `pad_num_tuples` with calls to `self.pad_sequence_to_length`, giving that method the appropriate parameters.

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+        num_tuples = len(tuples_in)
+        # If there are more than desired, we remove them
+        if num_tuples > desired_length:
+            tuples_in = tuples_in[:desired_length]
+        elif num_tuples < desired_length:
+            additional_tuples = [[] for _ in range(desired_length - num_tuples)]
+            tuples_in += additional_tuples
+        return tuples_in
+
+    # Helper method to pad an individual tuple both to the desired number of slots as well as the slot length
```
matt-gardner: To docstring.

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+    @overrides
+    def pad(self, max_lengths: Dict[str, int]):
+        """
+        Pads (or truncates) all slots in the answers_indexed and background_indexed to specified lengths.
+        Additionally, adjusts the number of slots in all tuples to the desired number: if there are fewer
+        slots than desired, will add empty lists at the end, and if there are more will
+        concatenate object slots.
+        Also, pads the number of answer options, the number of tuples per answer option, and the number
+        of background tuples to the specified lengths, adding empty ones as needed.
+        Note: Adjusting the number of slots happens before adjusting the lengths of all the slots.
```
matt-gardner: Can you explain this note a little more?  It's not clear from the comment that you're merging slots in some cases.

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+            all_slots = tuple_in
+            tuple_in = all_slots[:desired_num_slots]
+            for i in range(num_extra_slots):
+                tuple_in[desired_num_slots - 1] += all_slots[desired_num_slots + i]
+        elif num_slots < desired_num_slots:
+            additional_slots = [[] for _ in range(desired_num_slots - num_slots)]
+            tuple_in += additional_slots
+        # Pad the slots to the desired length
+        max_lengths_to_use = {'word_sequence_length': desired_slot_length}
+        tuple_in = [self.pad_word_sequence(indices, max_lengths_to_use) for indices in tuple_in]
```
matt-gardner: This doesn't treat character lengths correctly.  You might have a `word_character_length` key in the lengths dictionary that gets passed in to `pad`.  You need to pass that key through to this method, if it's present.

tests/data/instances/mc_tuples_with_background_instance_test.py
matt-gardner: I agree.  Though, if you rename this to `TupleInferenceInstance` (including moving the file), the line might be short enough to put these on the same line.

tests/data/instances/mc_tuples_with_background_instance_test.py
```diff
@@ -0,0 +1,284 @@
...
+        assert len(self.instance_2.answer_tuples) == 3
+        assert len(self.instance_2.answer_tuples[1]) == 0
+        # test no background tuples
+        assert len(self.instance_3.background_tuples) == 0
+
+
+    def test_words_method(self):
+        words = self.instance.words()['words']
+        word_set = set(words)
+        original_words = set("cat is a mammal which is a nocturnal mammal cat is nocturnal which is a nocturnal " \
```
matt-gardner: Use `expected_words` instead of `original_words`.

tests/data/instances/mc_tuples_with_background_instance_test.py
```diff
@@ -0,0 +1,284 @@
...
+                             "mammal snake is a mammal which is a nocturnal mammal snake is nocturnal which is "
+                             "a nocturnal mammal cat eat food cat is a mammal with fur which is a mammal snake "
+                             "is example of reptile some snakes are highly venomous".split(" "))
+        assert word_set == set(original_words)
```
matt-gardner: This is already a set.

tests/data/instances/mc_tuples_with_background_instance_test.py
```diff
@@ -0,0 +1,284 @@
...
+        for answer_option_tuples in self.indexed_instance.answers_indexed:
+            for ans_tuple in answer_option_tuples:
+                for slot in ans_tuple:
+                    assert len(slot) == question_slot_length
+
+        for background_tuple in self.indexed_instance.background_indexed:
+            for slot in background_tuple:
+                assert len(slot) == background_slot_length
+
+    def test_as_training_data_produces_correct_numpy_arrays(self):
```
matt-gardner: I think this test covers almost all of your individual tests above, and makes them unnecessary.  You can check this by looking at coverage statistics with and without the above tests.  (Maybe @nelson-liu can tell us the easiest way to check coverage without pushing a commit?)

tests/data/instances/mc_tuples_with_background_instance_test.py
```diff
@@ -0,0 +1,284 @@
...
+                       'num_slots': 2,
+                       'question_slot_length': 2,
+                       'background_slot_length': 3,
+                       'num_options': 3}
+        self.indexed_instance.pad(max_lengths)
+
+        inputs, label = self.indexed_instance.as_training_data()
+        assert np.all(label == np.asarray([0, 0, 1]))
+
+        desired_options = np.asarray([[[[1, 2], [0, 3]],
```
matt-gardner: It'd be nice if there were a test similar to this one that made sure that using the `WordAndCharactersTokenizer` worked with this instance.

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+        self.verb = verb
+        self.objects = objects
+        self.context = context
+        self.time = time
+        self.location = loc
+        self.source = source
+
+    def __str__(self):
+        # NOTE: not currently displaying locaton, time, and source info
+        return "TextTuple( [\n\tsubj: {0} \n\tverb: {1} \n\tobjects: {2} \n\tcontext: {3}\n] )"
```
bsharpataz: haha -- nope, didn't actually pass them :)  yay me!
if it's ok to leave it like this, I personally find this much more readable than the other way....
(but I'll, you know, pass in stuff)

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+        # Currently, source, time and location aren't being used, but this can be changed in the future.
+
+        return text_list
+
+
+    @classmethod
+    def from_string(cls, input_string: str):
+        tuple_fields = input_string.lower().split("|")
+        num_fields = len(tuple_fields)
+        assert num_fields >= 2, "num_fields not >= 2" + input_string
```
bsharpataz: when I go to make a doc string on a method, some auto text pops up -- can you tell me if I am populating this correctly?
``` """
        Converts a string representation of the tuple into a TextTuple.
        :param input_string: str
        :return TextTuple 
        """ ```

Comment (bsharpataz) [inline deep_qa/data/instances/mc_tuples_with_background_instance.py]: well, the ``` interacted with the """ but I think you get the picture

File: deep_qa/data/instances/mc_tuples_with_background_instance.py [hunk 33]
```diff
@@ -0,0 +1,358 @@
...
+        subject = tuple_fields[0]
+        verb = tuple_fields[1]
+        objects = []
+        context, time, location, source = None, None, None, None
```
bsharpataz: changed, and I didn't know you could do this, but for the record this bothers me that python works like this, I'd think that that would make them all pointers to the same thing, s.t. if you change one you change them all.  Guess I still don't get pass-by-reference!

deep_qa/data/instances/mc_tuples_with_background_instance.py
bsharpataz: I just feel like that format would be a big long mess here, but I guess this is a mess too.  *strong* feelings from either of you?

deep_qa/data/instances/mc_tuples_with_background_instance.py
nelson-liu: i think that's the python docstring convention (pep8 i think), but we're using the numpy conventions here so it doesn't quite match.

deep_qa/data/instances/mc_tuples_with_background_instance.py
nelson-liu: i think it's a good idea to have them all have the same format. in addition, the `__str__` method of the instance is actually called on debug, I believe, and having a large instance with too many newlines could be quite cumbersome unto itself.

deep_qa/data/instances/mc_tuples_with_background_instance.py
bsharpataz: I just wanted it as a check, since the file is essentially unreadable to a human, in case somewhere between tushar passing me data, me splicing in the answer label from another source, or me loading in the data it got screwed up.

tests/data/instances/mc_tuples_with_background_instance_test.py
bsharpataz: oooh, done.

tests/data/instances/mc_tuples_with_background_instance_test.py
bsharpataz: it doesn't seem to like the + -- gives me a red underline.  I'm not sure when I can use one and when I can't.  Strong feelings?

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+        if include_context and self.context:
+            text_list.append(self.context)
+        # Currently, source, time and location aren't being used, but this can be changed in the future.
+
+        return text_list
+
+
+    @classmethod
+    def from_string(cls, input_string: str):
+        tuple_fields = input_string.lower().split("|")
```
bsharpataz: I will msg tushar about changing it, but until he changes his end, I am hesitant to change mine, as I need to maintain compatibility.

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+            elif element.startswith("T:"):
+                time = element.split("T:")[1]
+            elif element.startswith("source:"):
+                source = element.split("source:")[1]
+            else:
+                objects.append(element)
+        return TextTuple(subject, verb, objects, context, time, location, source)
+
+
+class MCTuplesWithBackgroundInstance(TextInstance):
```
bsharpataz: agreed, adding.

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+    - label: int, default=None
+        The class label (i.e. the index of the correct multiple choice answer) -- corresponds to the
+        indices in self.answer_tuples.
+
+    - index: int, default=None
+        The index of the question.
+
+    """
+    def __init__(self, answer_tuples: List[List[TextTuple]], background_tuples: List[TextTuple],
+                 object_handling: str="collapse", use_context: bool=True, label: int=None, index: int=None):
```
bsharpataz: I hear what you're saying, but how would the Instance receive the model variables, though? or rather, how would the model set the instance variables?  seems like tokenizer is actually instantiated in TextTrainer and passed parameters, but the instance comes into being in TextDataset, and no parameters are passed, and though the instance_class comes from the model, it's not an "object" (prob using word incorrectly)
again -- not disagreeing, this is a problem, I just don't immediately see how to fix it here.

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+            # indices: List[List[int]]
+            indices = [self._index_text(phrase, data_indexer) for phrase in text_list]
+            indexed.append(indices)
+        return indexed
+
+    @classmethod
+    def read_from_line(cls, line: str, default_label: int=None):
+        """
+        Reads a TupleInstances from a line.  The format has one option (currently):
+            [question index][tab][num answer coptions][tab][all question+answer tuples][tab]
```
bsharpataz: redundant with... ?

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+            [background tuples]
+        The question+answer tuples are formatted as:
+            [a_1-tuple_1]$$$...$$$[a_1-tuple_n]###...###[a+_m-tuple_1]$$$...$$$||[a_m-tuple_p]
+        such that ``$$$`` serves as the tuple delimiter and ``###`` serves as the answer candidate
+        delimiter.
+        Each of the tuples is formatted as:
+            [s words]|[v words]|[o_1 words]|...|[o_j words]|[context: context words]
+        such that ``|`` serves as the tuple-slot delimiter.  Note that some slots may contain prefixes
```
bsharpataz: ok, again, I'll msg tushar

deep_qa/data/instances/mc_tuples_with_background_instance.py
```diff
@@ -0,0 +1,358 @@
...
+        index = int(fields[0])
+        # Number of answer candidates
+        num_answers = int(fields[1])
+        # Answer Candidate tuples
+        answers = fields[2].split("###")
+        answer_tuples = []
+        for curr_answer in answers:
+            curr_answer_tuple_strings = curr_answer.split("$$$")
+            curr_answer_tuples = [TextTuple.from_string(tuple_string)
+                                  for tuple_string in curr_answer_tuple_strings if len(tuple_string) > 0]
```
bsharpataz: cool, thanks

deep_qa/data/instances/mc_tuples_with_background_instance.py
bsharpataz: would that be to pad the words to a set character length? if so, is there a super class method for that padding?

tests/data/instances/mc_tuples_with_background_instance_test.py
bsharpataz: I was thinking that with the above I was testing truncating and expanding (padding), but i can certainly remove tests if you want!

tests/data/instances/mc_tuples_with_background_instance_test.py
bsharpataz: um... yes... so, with that tokenizer, the innermost list would not be a List[Int] (i.e. list of word indices, it would be a List[List[Int]], where each elem of the outer list is a list of indices represetning [word_index, char0_index, ..., charN_index] right?

bsharpataz: if that's the case, at the very least any typing I did in code would need to be removed and typing in comments would need to be altered, but that's easy... I need to check for other changes, and then I can try to write this test.

deep_qa/data/instances/mc_tuples_with_background_instance.py
bsharpataz: if I removed newlines, it would be hard to read, which was my intent here.  what do you think of me changing this to be a diff name (like... display()) and leave the __str__ method to print class-type info?

bsharpataz:
ok all -- millions of comments back and forth.  I have changed most of these -- but left a few unaddressed for now until I hear back.  I'll go ahead and re-push to make viewing these comments more manageable (at least for me)

deep_qa/data/instances/mc_tuples_with_background_instance.py
matt-gardner: I think the better way to check for those things is to write tests for the code that does them, instead of having a data representation where the same information is encoded twice in different places.

tests/data/instances/mc_tuples_with_background_instance_test.py
matt-gardner: How about formatting it like:

```python
background = ("cat|eat|..."
              "next line..."
              "last line...")

File: deep_qa/data/instances/mc_tuples_with_background_instance.py (hunk 34 above)
Comment (matt-gardner) [inline deep_qa/data/instances/mc_tuples_with_background_instance.py]: I see, yeah, you need to match what Tushar does.

File: deep_qa/data/instances/mc_tuples_with_background_instance.py (hunk 36 above)
Comment (matt-gardner) [inline deep_qa/data/instances/mc_tuples_with_background_instance.py]: Notice [this line](https://github.com/allenai/deep_qa/blob/aa1c81296866001d465287d98ff675c4ef9b3ea9/deep_qa/data/instances/instance.py#L77), where the tokenizer is used in the instance object.  There is no definition of `self.tokenizer` in the `self.__init__` method - `tokenizer` is a _class_ variable, not an _instance_ variable.  It's basically the same thing as `static` in java.  So, you can make these parameters _class_ variables, with reasonable defaults, and change them in the model class if you need to, by setting `TupleInferenceInstance.object_handling`, etc.  This sets it for every instance of the class, which is what you want (assuming you only ever want one behavior for a particular run of your program, which I think is a reasonable assumption).

File: deep_qa/data/instances/mc_tuples_with_background_instance.py (hunk 37 above)
Comment (matt-gardner) [inline deep_qa/data/instances/mc_tuples_with_background_instance.py]: With the number of answer options in the third column.  This is related to my earlier comment about encoding the same information twice.

File: deep_qa/data/instances/mc_tuples_with_background_instance.py (hunk 26 above)
Comment (matt-gardner) [inline deep_qa/data/instances/mc_tuples_with_background_instance.py]: `self.pad_word_sequence` does the right thing with that, if you've passed in the `lengths` dictionary correctly.

File: tests/data/instances/mc_tuples_with_background_instance_test.py (hunk 29 above)
Comment (matt-gardner) [inline tests/data/instances/mc_tuples_with_background_instance_test.py]: This test also covers truncating and expanding, because you call `indexed_instance.pad()`, and in some cases that truncates the input and in other cases it pads.  You can check using the coverage stats whether this is actually covering the same code, as I think it is.  Especially if you use the super class methods (`pad_sequence_to_length`), which are already tested, you'll have fewer branches to cover and can get to 100% coverage with many fewer tests.

Not that having tests is bad, it's just more work to write and maintain, and it's not worth the extra work if it's testing exactly the same thing.

File: tests/data/instances/mc_tuples_with_background_instance_test.py (hunk 30 above)
Comment (matt-gardner) [inline tests/data/instances/mc_tuples_with_background_instance_test.py]: Yes, you're right about that.  The type annotations are not actually enforced by anything, they are there to improve readability.  So I think it's fine to leave the annotations as they are, as they document the typical case, even though the type might actually change if you use a different tokenizer.

And I think the only things you should need to change are things I've already suggested - if you make sure you're calling super class methods to actually do the indexing of words (you are) and to handle padding word sequences (you're not yet, but I gave comments on that), it should just work.

**Comment by bsharpataz** [comment]
ok -- I have addressed many of these, and just repushed, so hopefully github will collapse the old/unaddressed ones.

I know I still need to write the test that Matt wants, working on that now...

**Comment by bsharpataz** [comment]
oh, and re: the num_answers column in the input -- I'll add that to the
tushar request and when it's changed on his end I'll change it here

File: deep_qa/data/instances/tuple_inference_instance.py [hunk 40]
```diff
@@ -0,0 +1,443 @@
...
+    index: int, default=None
+        The index of the question.
+    """
+
+    def __init__(self, answer_tuples: List[List[TextTuple]], background_tuples: List[TextTuple],
+                 label: int=None, index: int=None):
+        super(TupleInferenceInstance, self).__init__(label, index)
+        self.answer_tuples = answer_tuples
+        self.background_tuples = background_tuples
+        self.object_handling = tuple_inference_parameters['object_handling']
```
matt-gardner: This should be outside of `__init__`, without the global object.  The class should look like this:

```python
class TupleInferenceInstance(TextInstance):
    """docstring here"""
    object_handling = 'collapse'
    use_context = False

    def __init__(self, answer_tuples: ...):
        ...
```

After that, you can call `self.object_handling` just fine in other methods in this class, and you can change it from a model class by calling `TupleInferenceInstance.object_handling = "something else"`.

deep_qa/data/instances/tuple_inference_instance.py
```diff
@@ -0,0 +1,443 @@
...
+        """
+        desired_num_options = max_lengths['num_options']
+        desired_num_question_tuples = max_lengths['num_question_tuples']
+        desired_num_background_tuples = max_lengths['num_background_tuples']
+        desired_num_slots = max_lengths['num_slots']
+        desired_question_slot_length = max_lengths['question_slot_length']
+        desired_background_slot_length = max_lengths['background_slot_length']
+        desired_word_character_length = None
+        if 'max_word_character_length' in max_lengths:
+            desired_word_character_length = max_lengths['max_word_character_length']
```
matt-gardner: There's no "max" in this name.

deep_qa/data/instances/tuple_inference_instance.py
```diff
@@ -0,0 +1,443 @@
...
+        # Pad the number of tuples per option.
+        for answer_index in range(len(self.answers_indexed)):
+            self.answers_indexed[answer_index] = self.pad_sequence_to_length(self.answers_indexed[answer_index],
+                                                                             desired_num_question_tuples,
+                                                                             lambda: [],
+                                                                             truncate_from_right=False)
+        # Pad each of the tuples for slot length and words per slot
+        for answer_index in range(len(self.answers_indexed)):
+            self.answers_indexed[answer_index] = [self.pad_tuple(answer_tuple,
+                                                                 desired_num_slots,
```
matt-gardner: You could make this a bit simpler if you just pass `lengths` along to this method, and read the values that you need there.

deep_qa/data/instances/tuple_inference_instance.py
```diff
@@ -0,0 +1,443 @@
...
+        ----------
+        tuple_in: List[List[int]]
+            (num_tuple_slots, num_slot_words)
+            The indexed tuple to be padded.
+
+        desired_num_slots: int
+
+        desired_slot_length: int
+
+        desired_word_character_length: Any
```
matt-gardner: If a parameter is optional like this, we still say the type is `int`.

deep_qa/data/instances/tuple_inference_instance.py
```diff
@@ -0,0 +1,443 @@
...
+            has a length equal to desired slot_length.
+        '''
+        tuple_in = self.pad_sequence_to_length(tuple_in, desired_num_slots,
+                                               default_value=lambda: [], truncate_from_right=False)
+        # Pad the slots to the desired length.
+        max_lengths_to_use = {'word_sequence_length': desired_slot_length}
+        if desired_word_character_length is not None:
+            max_lengths_to_use['word_character_length'] = desired_word_character_length
+        tuple_in = [self.pad_word_sequence(indices, max_lengths_to_use) for indices in tuple_in]
+        for slot in tuple_in:
```
matt-gardner: No need for this assert.  That's what tests are for.

deep_qa/data/instances/tuple_inference_instance.py
```diff
@@ -0,0 +1,443 @@
...
+        if label >= len(answer_tuples):
+            raise ConfigurationError("Invalid label, label is >= the number of answers.")
+
+        return cls(answer_tuples, background_tuples, label=label, index=index)
+
+
+class IndexedTupleInferenceInstance(IndexedInstance):
+    def __init__(self,
+                 # (num_answer_options, num_answer_tuples, num_tuple_slots, num_slot_words)
+                 answers_indexed: List,
```
matt-gardner: I liked having the type annotation that you had previously.  I just thought that you could add a comment _in addition_ to the type annotation.  I'd also probably put it in a docstring, or something, instead of an inline comment like this.

deep_qa/data/instances/tuple_inference_instance.py
```diff
@@ -0,0 +1,443 @@
...
+
+        max_slot_words: int
+            The max number of words found in any of the slots.
+
+        max_word_length: int
+            The length of the longest word, with a return value of -1 if the model isn't using character padding.
+        '''
+        max_num_slots = 0
+        max_slot_words = 0
+        max_word_length = 0
```
matt-gardner: This needs to be `None`, not 0.

matt-gardner: Ok, I see, you check for it above, so it's probably ok.  The default value doesn't match the docstring, though.

deep_qa/data/instances/tuple_inference_instance.py
```diff
@@ -0,0 +1,443 @@
...
+        max_word_length = max(max_word_length, background_word_length)
+
+        lengths = {'num_options': len(self.answers_indexed),
+                   'num_question_tuples': max_question_tuples,
+                   'num_background_tuples': len(self.background_indexed),
+                   'question_slot_length': max_answer_length,
+                   'background_slot_length': background_length,
+                   'num_slots': max_num_slots}
+        if max_word_length > 0:
+            lengths['word_character_length'] = max_word_length
```
matt-gardner: Looks like you need a `return lengths` here.

deep_qa/data/instances/tuple_inference_instance.py
bsharpataz: yes and no -- if I remove the ability for the question slot length to be diff than the background slot length, then yes.  Otherwise, pad_tuple doesn't know whether the tuple is a question or background, so I can simplify it by passing the max_lengths, but I have to add a flag to let it know if it's a question or background.  Preference?

deep_qa/data/instances/tuple_inference_instance.py
bsharpataz: I removed it -- but to make sure I understand things, this dict of params comes from my model, right, so in theory I could name it whatever I want, as long as I pass the right thing to the tokenizer? as in... since i have two separate lengths for sentences, in pad_tuple I construct this =dict to pass to the tokenizer:
```
max_lengths_to_use = {'word_sequence_length': desired_slot_length}
if desired_word_character_length is not None:
       max_lengths_to_use['word_character_length'] = desired_word_character_length
tuple_in = [self.pad_word_sequence(indices, max_lengths_to_use) for indices in tuple_in]
```

Basically -- I can simplify things by removing the extra dimension, but can I do that after the whole PR (like, instance, model, etc) is done bc the changes are in both... or do you want me to do it now so we don't do this all again. idk.

deep_qa/data/instances/tuple_inference_instance.py
bsharpataz: Sorry -- I guess I misunderstood the previous response.  So it is not the case that the typing is enforced in python?  I thought you were saying "yes, you have to remove it from the type declarations, but no you don't have to remove it from the comments bc thats the default behavior" -- so not enforced and back in?

deep_qa/data/instances/tuple_inference_instance.py
bsharpataz: and as long as it matches what's in get_lengths()?

deep_qa/data/instances/tuple_inference_instance.py
bsharpataz: thanks -- this was really helpful!

deep_qa/data/instances/tuple_inference_instance.py
matt-gardner: `word_character_length` and `word_sequence_length` are keys defined in the super classes, and so for the API to work correctly, you need to use those keys when you call `pad_word_sequence`.    Yes, you could feasibly just change the name from your internal key when you pass it in, but it'd be easier to understand what's going on if you just used the same name.

matt-gardner: `word_sequence_length` is an exception, if you have several different word sequence lengths - you _have_ to have different keys in that case.  But you almost certainly don't have multiple word character lengths.

deep_qa/data/instances/tuple_inference_instance.py
matt-gardner: Do you know what the average values are for the question slot length vs. the background slot length?  If they are not very different, I think it would make the code a lot simpler to just use the same length for both.

deep_qa/data/instances/tuple_inference_instance.py
matt-gardner: Yeah, type annotations aren't actually enforced, so I'd put it back in.

doc/data/instances.rst
```diff
@@ -152,6 +152,15 @@ true_false_instance
     :undoc-members:
     :show-inheritance:
 
+tuple_inference_instance
+---------------------------
+
+.. automodule:: deep_qa.data.instances.mc_tuples_with_background_instance
```
matt-gardner: This needs to be updated.

tests/data/instances/tuple_inference_instance_test.py
```diff
@@ -0,0 +1,176 @@
...
+        indexed = new_instance.to_indexed_instance(data_indexer)
+
+        max_lengths = {'num_question_tuples': 1,
+                       'num_background_tuples': 1,
+                       'num_slots': 2,
+                       'word_sequence_length': 2,
+                       'num_options': 1,
+                       'word_character_length': 3}
+        indexed.pad(max_lengths)
+        expected_indexed_tuple = [[0, 0, 0], [a_word_index, a_index, 0]],\
```
matt-gardner: I don't think you need the `\` here, because you're inside of a list.

PR #215, opened 2017-02-21
The MCTupleWithBackgroundEntailmentModel

bsharpataz:
OK -- so there are a few files here which are included with other PRs (so don't feel the need to do a style-based code review on ones you've already done that for.  When the other PRs pass, I'll merge upstream into this and resubmit).  They're included here because I've been trying to make sure things are actually doing what I want them to do -- I get that that's ugly and can redo this, but you wanted it PR'd tonight so you could look at it. So, sorry!

I think the issue is that the TimeDistributed isn't building the inner layer (or it could be the inner layer (TupleMatch) not playing well with the TimeDistributed building, but the model only has 2 trainable weights, whereas before the newer versions it had 27.  The tuple_match test is passing ok -- feel free to check that though to make sure that it's correct and sufficient.

thanks!


bsharpataz:
the model is also likely not in the docs, I can add that tomorrow -- but you wanted it submitted :(

bsharpataz:
Home now, but i was thinking that i'll remove the script from this PR tomorrow.  That said, you can review the rest when you get the chance. Thanks!

tests/models/multiple_choice_qa/multiple_choice_tuple_inference_model_test.py
```diff
@@ -0,0 +1,29 @@
...
+
+class TestMultipleChoiceTupleInferenceModel(TestCase):
+    def setUp(self):
+        os.makedirs(TEST_DIR, exist_ok=True)
+        write_multiple_choice_tuple_inference_files()
+
+    def tearDown(self):
+        shutil.rmtree(TEST_DIR)
+
+    def test_train_does_not_crash(self):
```
matt-gardner: Can you test loading, too?  The test ends up looking like this: https://github.com/allenai/deep_qa/blob/942d66980b589d924e07bc127f0c38e4a7c89352/tests/models/multiple_choice_qa/multiple_true_false_memory_network_test.py#L26-L35.

Making sure this works is really important for actually running this in the evaluation framework, which you should be able to do very soon.

deep_qa/models/multiple_choice_qa/multiple_choice_tuple_inference.py
```diff
@@ -0,0 +1,167 @@
...
+
+    num_tuple_slots: int, default=4
+        The number of slots in each tuple.
+
+    max_slot_length: int, default=5
+        The number of words in each slot of the tuples.
+
+    num_options: int, default=4
+        The number of answer options/candidates.
+
```
bsharpataz: added to the current tuple match layer

deep_qa/models/multiple_choice_qa/multiple_choice_tuple_inference.py
```diff
@@ -0,0 +1,167 @@
...
+        # Expand and tile the question input to be:
+        # shape: (batch size, num_options, num_question_tuples, num_background_tuples, num_tuple_slots,
+        #         num_slot_words)
+        tiled_question = Repeat(axis=3, repetitions=K.int_shape(background_input)[1])(question_input)
+
+        # Expand and tile the background input to match question input.
+        # shape: (batch size, num_options, num_question_tuples, num_background_tuples, num_tuple_slots,
+        #               num_slot_words)
+        # First, add num_options.
+        tiled_background = Repeat(axis=1, repetitions=K.int_shape(question_input)[1])(background_input)
```
bsharpataz: yes! very much agreed, thanks!

deep_qa/models/multiple_choice_qa/multiple_choice_tuple_inference.py
```diff
@@ -0,0 +1,167 @@
...
+        # Find the matches between the question and background tuples.
+        # shape: (batch size, num_options, num_question_tuples, num_background_tuples, 1)
+        matches = match_layer([tiled_question, tiled_background])
+        # Squeeze to get rid of the last dim of length 1.
+        # shape: (batch size, num_options, num_question_tuples, num_background_tuples)
+        matches = Squeeze(axis=-1)(matches)
+
+        # Find the probability that any given question tuple is entailed by the given background tuples.
+        # shape: (batch size, num_options, num_question_tuples)
+        combine_background_evidence = NoisyOr(axis=-1, param_init=self.noisy_or_param_init, name="noisyOr_1")
```
bsharpataz: but it's the Layer's name, aren't we doing the camel case for layer/model names? e.g. TextTrainer?
it could be confusing that I am naming the instance of the NoisyOr layer "combine_background_evidence" -- which I actually like because it describes what we're actually doing when we use the noisy-or here, but if you have really string feelings, I can change that to something like... noisy_or_1 or something?

matt-gardner: That name is what gets printed out in the Keras model summary.  If you notice, all names there have been converted from camel case (which is appropriate for class names in python) to underscores, or just all lowercase.  So you get this like "vectormatrixmerge", "timedistributed_4" and "word_question_embedding".  This should be either "noisy_or_1", "noisyor_1", or something like "combine_background_evidence".

bsharpataz: oh!  this is a case of whole line highlighting confusing me!  I totally see what you're talking about now, sorry -- though you meant the name of the NoisyOr class...! gotcha, will change

deep_qa/layers/entailment_models/tuple_match.py
```diff
@@ -34,31 +34,52 @@ class TupleMatch(Layer):
...
     - initialization : string, default='glorot_uniform'
         The initialization of the NN weights
+
     - hidden_layer_activation : string, default='relu'
         The activation of the NN hidden layers
+
     - final_activation : string, default='sigmoid'
         The activation of the NN output layer
+
+    - name : string, default='tuple_match'
```
matt-gardner: This one is already covered by the super class, so you don't need to have it here.

deep_qa/layers/entailment_models/tuple_match.py
```diff
@@ -34,31 +34,52 @@ class TupleMatch(Layer):
...
+
+    Notes
+    _____
+    This layer is incompatible with the WordsAndCharacters tokenizer.
     """
 
     def __init__(self, num_hidden_layers: int=1, hidden_layer_width: int=4,
                  initialization: str='glorot_uniform', hidden_layer_activation: str='tanh',
-                 final_activation: str='sigmoid'):
+                 final_activation: str='sigmoid', name: str='tuple_match', **kwargs):
```
matt-gardner: No need for the `name` param here.

deep_qa/layers/entailment_models/tuple_match.py
```diff
@@ -34,31 +34,52 @@ class TupleMatch(Layer):
...
         # Parameters for the shallow neural network
         self.num_hidden_layers = num_hidden_layers
         self.hidden_layer_width = hidden_layer_width
-        self.hidden_layer_init = initializations.get(initialization)
-        self.hidden_layer_activation = activations.get(hidden_layer_activation)
-        self.final_activation = activations.get(final_activation)
+        self.hidden_layer_init = initialization
+        self.hidden_layer_activation = hidden_layer_activation
+        self.final_activation = final_activation
+        self.name = name
```
matt-gardner: Or here.

deep_qa/models/multiple_choice_qa/tuple_inference.py
```diff
@@ -66,27 +69,36 @@ def _instance_type(self):
...
     @overrides
     def _set_max_lengths(self, max_lengths: Dict[str, int]):
+        super(TupleInferenceModel, self)._set_max_lengths(max_lengths)
         self.num_question_tuples = max_lengths['num_question_tuples']
         self.num_background_tuples = max_lengths['num_background_tuples']
         self.num_tuple_slots = max_lengths['num_slots']
-        self.max_slot_length = max_lengths['word_sequence_length']
+        self.num_slot_words = max_lengths['word_sequence_length']
         self.num_options = max_lengths['num_options']
+        if 'word_character_length' in max_lengths:
```
matt-gardner: This is set in the super class, so you don't need to have this here.

deep_qa/models/multiple_choice_qa/tuple_inference.py
```diff
@@ -66,27 +69,36 @@ def _instance_type(self):
...
-                'word_sequence_length': self.max_slot_length,
-                'num_options': self.num_options,
-                }
+        max_lengths = super(TupleInferenceModel, self)._get_max_lengths()
+        max_lengths['num_question_tuples'] = self.num_question_tuples
+        max_lengths['num_background_tuples'] = self.num_background_tuples
+        max_lengths['num_slots'] = self.num_tuple_slots
+        max_lengths['word_sequence_length'] = self.num_slot_words
+        max_lengths['num_options'] = self.num_options
+        if self.word_character_length is not None:
```
matt-gardner: This happens in the super class.  You only need to worry about the things you define in `self.__init__`.

tests/models/multiple_choice_qa/tuple_inference_model_test.py
```diff
@@ -0,0 +1,40 @@
...
+                "word_sequence_length": 10,
+                "num_answer_options": 4}
+        solver = get_model(TupleInferenceModel, args)
+        solver.train()
+
+    def test_model_trains_and_loads_correctly(self):
+        model = get_model(TupleInferenceModel, {'save_models': True})
+        model.train()
+
+        loaded_model = get_model(TupleInferenceModel)
```
matt-gardner: I would combine these two tests, so that we don't run the training twice in the tests (it takes at least several seconds, sometimes as much as 30-40).  Just add from here down into the other test, and use the name from this one.

PR #217, opened 2017-02-23
Fixed some loading issues with DenseConcatMemoryUpdater

matt-gardner:


deep_qa/layers/memory_updaters.py
```diff
@@ -53,41 +53,34 @@ class DenseConcatNoQuestionMemoryUpdater(Dense):
...
         # For this memory update, we don't need the question encoding. This is the first
         # concatentated vector, so we remove it.
-        x = x[:, self.encoding_dim:]
+        x = x[:, self.output_dim:]
         return super(DenseConcatNoQuestionMemoryUpdater, self).call(x)
 
     def get_output_shape_for(self, input_shape):
         return (input_shape[0], int(input_shape[1] / 3))
 
-    def get_config(self):
```
nelson-liu: can you explain why this doesn't need a `get_config`?

matt-gardner: It's inheriting from `Dense`, which has a `get_config` method, and doesn't add any new parameters.

nelson-liu:
looks like tests here are failing because of the same pylint issue in #219 --- i'll push that PR to master first and then you can update?

PR #219, opened 2017-02-23
add proper get_config to GatedAttention layer

### LINKED ISSUE #218: GatedAttention doesn't load correctly
The `GatedAttention` layer doesn't implement `get_config`, and so the gating function doesn't get loaded correctly.  You can check this in a test by specifying a non-default gating function when training the model; it will load with the wrong one.

nelson-liu:
Fixes #218 by adding a `get_config` function to the GatedAttention layer.

### DIFF
--- a/build_tools/travis/install.sh
+++ b/build_tools/travis/install.sh
@@ -28,7 +28,7 @@ if [[ ! -d /home/travis/miniconda3 ]]
     # If we are running pylint, use Python 3.5.2 due to
     # bug in pylint. https://github.com/PyCQA/pylint/issues/1295
     if [[ "$RUN_PYLINT" == "true" ]]; then
-        conda create -n testenv --yes python=3.5.2
+        conda create -n testenv352 --yes python=3.5.2
     else
         conda create -n testenv --yes python=3.5
     fi
@@ -37,7 +37,11 @@ cd ..
 popd
 
 # Activate the python environment we created.
-source activate testenv
+if [[ "$RUN_PYLINT" == "true" ]]; then
+    source activate testenv352
+else
+    source activate testenv
+fi
 
 # Install requirements via pip in our conda environment
 pip install -r requirements.txt
--- a/build_tools/travis/install.sh
+++ b/build_tools/travis/install.sh
@@ -25,12 +25,18 @@ if [[ ! -d /home/travis/miniconda3 ]]
     export PATH=/home/travis/miniconda3/bin:$PATH
     echo $PATH
     conda update --yes conda
-    conda create -n testenv --yes python=3.5
+    # If we are running pylint, use Python 3.5.2 due to
+    # bug in pylint. https://github.com/PyCQA/pylint/issues/1295
+    if [[ "$RUN_PYLINT" == "true" ]]; then
+        conda create -n testenv --yes python=3.5.2
+    else
+        conda create -n testenv --yes python=3.5
+    fi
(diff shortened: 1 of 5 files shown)


deep_qa/layers/attention/gated_attention.py
```diff
@@ -124,3 +123,8 @@ def call(self, inputs, mask=None):
             raise ConfigurationError("Invalid gating function "
                                      "{}, expected one of {}".format(self.gating_function,
                                                                      GATING_FUNCTIONS))
+
+    def get_config(self):
+        config = {'gating_function': self.gating_function}
+        base_config = super(GatedAttention, self).get_config()
+        return dict(list(base_config.items()) + list(config.items()))
```
matt-gardner: Can't you just do `config.update(base_config)`?  Seems more readable.

nelson-liu: i think so, but keras uses this pattern across their codebase. https://github.com/fchollet/keras/blob/3d400116b95074f454605605f5339b90e28b9247/keras/layers/advanced_activations.py#L37

nelson-liu: i'll change it and we can see if anything breaks

build_tools/travis/install.sh
```diff
@@ -25,12 +25,18 @@ if [[ ! -d /home/travis/miniconda3 ]]
     export PATH=/home/travis/miniconda3/bin:$PATH
     echo $PATH
     conda update --yes conda
-    conda create -n testenv --yes python=3.5
+    # If we are running pylint, use Python 3.5.2 due to
```
matt-gardner: Are you sure this will work?  Seems like you would need to create a `testenv352`, and activate one or the other based on `$RUN_PYLINT`.  The testenv is cached and only created once.

nelson-liu: yeah good point, that's more robust

matt-gardner: Because of the `if [[ ! -d /home/travis/miniconda3 ]]` check at the top, this will only get run once, and so only one of these environments will ever get created, and then the `activate` line will fail when the cache gets loaded.  Does that make sense?  So I think you just need to have two `conda create -n ...` commands, like you do above, without any `ifs`.

deep_qa/layers/attention/gated_attention.py
matt-gardner: I guess I should note that keras uses `nb_` instead of `num_` across their codebase.  That doesn't mean we should make the same poor choice...

build_tools/travis/install.sh
nelson-liu: ah i see yeah that makes sense

deep_qa/models/reading_comprehension/gated_attention_reader.py
```diff
@@ -152,7 +159,7 @@ def _build_model(self):
             # (batch size, document length, question length)
             normalized_qd_attention = MaskedSoftmax()([qd_attention])
 
-            gated_attention_layer = GatedAttention()
+            gated_attention_layer = GatedAttention(self.gating_function)
```
matt-gardner: Looks like you should give this a name.  This is why the theano test is failing.

nelson-liu: yup i think that did the trick, thanks

PR #222, opened 2017-02-24
Add minimum sphinx version

nelson-liu:
sphinx.ext.autosectionlabel, which we use, was added in 1.4. ReadTheDocs uses 1.3.5 by default. I think that if i just specify a higher version manually, then it should work out.


nelson-liu:
nevermind i think a better solution is to just not use `sphinx.ext.autosectionlabel`

PR #224, opened 2017-02-25
Rename TupleMatch and make module

bsharpataz:
OK -- this is just the rearranging/renaming part so I can start making a new version -- I think I need to do something diff in the custom_objects, but I don't know if I add all that it could ever be, or just add the base TupleMatch and the others fall in line... figured you'd have other changes anyway, so I can make the adjustments with the rest.

Thanks!


deep_qa/layers/tuple_matchers/tuple_match.py
```diff
@@ -0,0 +1,23 @@
+from keras.layers import Layer
+
+class TupleMatch(Layer):
+    '''
+    TupleMatch layers take two tuples (of either the same size or different sizes), which are passed in as a
+     a list of two tensor inputs, and compare them to determine how well they match.  For example, an answer
```
matt-gardner: Spacing here is a little off, and you have "passed in as a a list".  Can you give a comment about the expected input shapes here?  I'm not sure what situation you might have where the two tuples have different shapes.

bsharpataz: added examples to the docstring

matt-gardner:
If I've approved it with only minor changes like this, you can just merge the PR when you're satisfied.  No need to have me review it again.

bsharpataz:
ok :)

PR #230, opened 2017-03-01
Fixed WeightedSum, added option to skip mask computation

matt-gardner:


nelson-liu:
this LGTM, looks like circle status isn't updating on github, even though it has passed(AWS issues perhaps)? merging.

PR #233, opened 2017-03-02
Ensure additional fitting works after loading models

nelson-liu:
This PR adds some tests to verify that we're able to fit loaded, serialized models on additional data for fine-tuning.

- [X] Add a tests for fitting after loading for MultipleTrueFalseMemoryNetwork, which should already be implemented correctly
- [x] Fix `_set_max_lengths_from_model` for `TrueFalseModel`
- [x] Add and pass tests for fitting after loading for `TrueFalseModel`
- [x] Fix `_set_max_lengths_from_model` for `BiDAF`
- [x] Add and pass tests for fitting after loading for `BiDAF`

### DIFF
--- a/deep_qa/models/text_classification/true_false_model.py
+++ b/deep_qa/models/text_classification/true_false_model.py
@@ -59,6 +59,7 @@ class TrueFalseModel(TextTrainer):
     @overrides
     def _set_max_lengths_from_model(self):
         self.max_sentence_length = self.model.get_input_shape_at(0)[1]
+        self.max_word_length = self.model.get_input_shape_at(0)[2]
 
     @classmethod
     def _get_custom_objects(cls):
--- a/deep_qa/models/text_classification/true_false_model.py
+++ b/deep_qa/models/text_classification/true_false_model.py
@@ -58,7 +58,7 @@ class TrueFalseModel(TextTrainer):
 
     @overrides
     def _set_max_lengths_from_model(self):
-        self.set_max_lengths_from_model_input(self.model.get_input_shape_at(0)[1:])
+        self.set_text_lengths_from_model_input(self.model.get_input_shape_at(0)[1:])
 
     @classmethod
     def _get_custom_objects(cls):
--- a/deep_qa/models/text_classification/true_false_model.py
+++ b/deep_qa/models/text_classification/true_false_model.py
@@ -58,8 +58,8 @@ class TrueFalseModel(TextTrainer):
 
     @overrides
     def _set_max_lengths_from_model(self):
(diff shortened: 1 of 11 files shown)


nelson-liu:
@matt-gardner  --- theano tests seem to fail on my machine for even `MultipleTrueFalseMemoryNetwork`, which is interesting considering TF is fine / this model should already be correct. I'm looking into it, but wanted to let you know just in case you had any idea off the top of your head why this could be the case.

nelson-liu:
interesting, seems like TF is failing on the CPU as well when trying to fit `MultipleTrueFalseMemoryNetwork` after loading.

nelson-liu:
@matt-gardner this is (finally) ready to be looked at. a more real-world test would be to load an actual weights file from one of the training epochs of BiDAF on SQuAD (still running...) and see if I can get fitting after loading to work on that.

deep_qa/models/text_classification/true_false_model.py
```diff
@@ -59,6 +59,7 @@ def _instance_type(self):
     @overrides
     def _set_max_lengths_from_model(self):
         self.max_sentence_length = self.model.get_input_shape_at(0)[1]
```
matt-gardner: This should just be a single call to the super class: `super(TrueFalseModel, self)._set_max_lengths_from_model_input(self.model.get_input_shape_at(0)[1:]`

nelson-liu: good catch, forgot to fix. thanks

matt-gardner: Oh, wait.  Just `self._set_...`.  No need for the `super` there, or in other places, because you changed the method name.

matt-gardner: Which means you should probably name it something like `set_word_sequence_lengths_from_model_input`.

deep_qa/training/text_trainer.py
```diff
@@ -277,7 +277,7 @@ def _set_max_lengths_from_model(self):
         """
         raise NotImplementedError
 
-    def _set_max_lengths_from_model_input(self, input_slice):
+    def set_max_lengths_from_model_input(self, input_slice):
```
matt-gardner: I think this needs a more specific name, because it's only doing a particular kind of length.

matt-gardner: I'd be happy with `set_text_length_from_model_input`, or something similar.

nelson-liu: agreed, any suggestions?

nelson-liu:
merging this, thanks for looking @matt-gardner

PR #237, opened 2017-03-03
Adding sequence tagging models and instances

matt-gardner:
We still need to write code to get SQuAD into a sequence tagging format, but after that, this _should_ just work.


deep_qa/data/instances/sequence_tagging/pretokenized_tagging_instance.py
```diff
@@ -0,0 +1,61 @@
...
+        tags = []
+        for field in fields:
+            token, tag = field.split("###")
+            tokens.append(token)
+            tags.append(tag)
+        return cls(tokens, tags, index)
+
+    @overrides
+    def words_in_label(self):
+        return [tag for tag in self.label]
```
DeNeutoy: In this specific case, these are tokens, right? It confused me a bit that they were called tags rather than tokens. I would expect the tag to be in `{B, I, O}`.

matt-gardner: Yeah, these are tags.  I'll call the method `tags_in_label` to make this more clear.

PR #238, opened 2017-03-04
Rename max_length dictionary members

nelson-liu:
(Wanted to do this before it fell off my radar). This PR renames the canonical keys we use in the `max_length` dictionary (used to set the max lengths for padding) to match the class variables in `text_trainer`.

Fairly trivial PR, i just ran:
```
grep -rl "word_sequence_length" . | xargs sed -i 's/word_sequence_length/max_sentence_length/g'
grep -rl "word_character_length" . | xargs sed -i 's/word_character_length/max_word_length/g'
```
with a `git commit` after each.

renaming `_get_word_sequence_lengths` to `_get_max_sentence_lengths` was unintended, but I kind of like the new name --- thoughts? it should actually probably get a new name altogether since you can get both the `max_sentence_length` and the `max_word_length` from it...

### DIFF
--- a/deep_qa/contrib/span_prediction/span_model.py
+++ b/deep_qa/contrib/span_prediction/span_model.py
@@ -20,15 +20,15 @@ from loaders import  load_parses, read_barrons, get_science_terms # pylint: disa
 
 
 # constrained legal span lengths. Upper diagonal band in begin vs. end matrix.
-def legal_spans(max_sentence_length, max_span_length):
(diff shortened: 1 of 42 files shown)


matt-gardner:
I think I would actually prefer to standardize the other way, removing "max" entirely.  For example, the keys that you used in the AS Reader are formatted like `num_*`: https://github.com/allenai/deep_qa/blob/b000dc22924a15d0817062e6acc682851b718409/deep_qa/models/reading_comprehension/attention_sum_reader.py#L121-L124.

PR #241, opened 2017-03-08
Update requirements.txt for tensorflow-gpu

nelson-liu:
tensorflow-gpu seems to have the ability to run CPU code...so I think it is a reasonable default.

The main reason i made this change is because setting up the solver to run on docker installs from requirements.txt, and we probably want the GPU version on there.


nelson-liu:
looks like it's mostly OK...just complaining that there's no CUDA. i'll force CPU in the travis config?

nelson-liu:
nevermind, looks like it needs cuda even if it runs in CPU-only mode :/ yet another step to add to the already lengthy process of running the python server.

PR #242, opened 2017-03-08
Remove trailing whitespace from comment

nelson-liu:
I decided to make a PR instead, so we can make sure it's actually fixed.


matt-gardner:
Sorry...  Thanks for fixing this.

PR #243, opened 2017-03-09
fix BiDAF set_max_lengths_from_model

nelson-liu:
This fixes an error in the `_set_max_lengths_from_model` function. I think it clearly makes sense to pass `self.model.get_input_shape_at(0)[1][1:]`, considering the words are at `self.model.get_input_shape_at(0)[1][1]`.

I'd like to add a non-regression test to make sure that this issue doesn't come up again, but i'm not sure what sort of examples would be appropriate. thoughts?


matt-gardner:
I think the right way to test this is to re-index the data after you load a model in the loading tests.

nelson-liu:
@matt-gardner , think this is good to go?

PR #244, opened 2017-03-09
Hybrid model

bsharpataz:
for your consideration... 
also -- if you see a reason why I am using gobs of memory, please let me know! I can cut back on the num_background_tuples, etc but that's not a real solution, right?


tests/layers/tuple_matchers/threshold_tuple_matcher_test.py
```diff
@@ -0,0 +1,140 @@
...
+        # Apply the final activation function.
+        desired_result = activations.get(match_layer.final_activation)(K.dot(neural_network_feed_forward,
+                                                                             match_layer.score_layer))
+        result = model.predict([self.tuple1, self.tuple2])
+        assert_array_almost_equal(result, K.eval(desired_result))
+
+
+    def test_returns_masks_correctly(self):
+        # Test when one tuple is all padding.
+        # Here, since tuple3 is all padding, we want to return a mask value of 0 for this pair
```
matt-gardner: Comment here is off.

tests/layers/tuple_matchers/threshold_tuple_matcher_test.py
```diff
@@ -0,0 +1,140 @@
...
+        # Testing general unmasked case.
+        desired_overlap = K.variable(numpy.asarray([[0, 0, 0]]))
+        # Desired_overlap gets fed into the inner NN.
+        neural_network_feed_forward = apply_feed_forward(desired_overlap, layer_nn,
+                                                         activations.get(match_layer.hidden_layer_activation))
+        # Apply the final activation function.
+        desired_result = activations.get(match_layer.final_activation)(K.dot(neural_network_feed_forward,
+                                                                             match_layer.score_layer))
+        result = mask_model.predict([tuple1_words, tuple2_words])
+        assert_array_almost_equal(result, K.eval(desired_result))
```
matt-gardner: Don't you want the result to be masked here?  So it doesn't really matter what the result is, and you should just check that it's masked.  You can do that using the `OutputMask` layer (something like what's done here: https://github.com/allenai/deep_qa/blob/c9a013993c6f15f1a0d83ec68b726669f38cf543/tests/layers/vector_matrix_split_test.py#L53-L62).

bsharpataz: That's true as it's written, what I should really do though is have one slot not padding in tuple2, but also not overlapping tuple 1 (or overlapping, whichever) and make sure that the masked elements aren't matching/getting summed...  If that makes sense to you I'll make that change.

matt-gardner: Sure, that seems fine, because you have the test above making sure that the mask is computed correctly.

deep_qa/layers/tuple_matchers/threshold_tuple_matcher.py
```diff
@@ -0,0 +1,236 @@
...
+
+        # Expand tuple2 to shape: (batch size, num_slots, num_slot_words_t1, num_slot_words_t2, embedding_dim)
+        # First expand to shape: (batch size, num_slots, 1, num_slot_words_tuple2, embedding_dim)
+        expanded_tuple2 = K.expand_dims(tuple2_input, 2)
+        # Tile to desired dimensions.
+        tiled_tuple2 = K.repeat_elements(expanded_tuple2, K.int_shape(tuple1_input)[2], axis=2)
+
+        # Generate the similarity scores of each of the word pairs.
+        # shape: (batch size, num_slots, num_slot_words_tuple1, num_slot_words_tuple2)
+        tuple_word_similarities = self.similarity_function.compute_similarity(tiled_tuple1, tiled_tuple2)
```
DeNeutoy: One thing you can try here is using `K.tile` instead of `K.repeat_elements`. The implementation of `K.repeat_elements` in the tensorflow backend of keras is very naive (to the point of being wrong): https://github.com/fchollet/keras/blob/master/keras/backend/tensorflow_backend.py#L1541
It effectively makes K copies of the thing you want to copy in memory and then concatenates them together, because it doesn't use `tf.tile`. If you use `K.tile`, I think tensorflow will just keep K pointers to a single tensor, so this might help.

matt-gardner: That's a really good point, thanks for pointing that out @DeNeutoy.  I'm going to see about updating this all over the place.

deep_qa/layers/tuple_matchers/threshold_tuple_matcher.py
```diff
@@ -0,0 +1,236 @@
...
+
+        # Find the number of words that overlap in each of the slots.
+        # shape: (batch size, num_slots)
+        slot_overlap_sums = K.sum(K.sum(zeros_excluded_overlap, axis=3), axis=2)
+
+        # # Normalize by the number of words in tuple1.
+        # TODO(becky): should this be fixed to tuple1 or allowed to vary? Does switching input order work
+        # for varying?
+        # This block of code prevents dividing by zero during normalization:
+        divisor = num_tuple1_words_in_each_slot
```
bsharpataz: it is on line 212 I think...?

matt-gardner: Oh, right.  Looks like I can't read =).

tests/layers/tuple_matchers/threshold_tuple_matcher_test.py
```diff
@@ -0,0 +1,140 @@
...
+                                                                             match_layer.score_layer))
+        result = model.predict([self.tuple1, self.tuple2])
+        assert_array_almost_equal(result, K.eval(desired_result))
+
+
+    def test_returns_masks_correctly(self):
+        # Test when one tuple is all padding.
+        # Here, since tuple3 is all padding, we want to return a mask value of 0 for this pair
+        tuple1 = K.variable(self.tuple1)
+        mask1 = K.variable(numpy.asarray([[1, 1, 1]]))
```
bsharpataz: right you are.  thanks for the in-depth code review, matt! fixed and as you predicted, test failed, fixed the compute_mask as you did in word_overlap tuple matcher (NOW i better understand *why* you did that fix too), and we're now passing, but please do check again when I re-submit this sucker!

matt-gardner:
Hmm... I'm going to recommend closing this PR and opening a new one from the same branch.  No need to review anything; I've already looked at it, and it's good to merge.  We just need to convince Travis to actually run.

PR #245, opened 2017-03-09
Added a cumulative_sum function and an envelope layer

matt-gardner:


deep_qa/layers/backend/envelope.py
```diff
@@ -0,0 +1,56 @@
...
+
+    Specifically, the computation done here is the following::
+
+        after_span_begin = cumulative_sum(span_begin)
+        after_span_end = cumulative_sum(span_end)
+        before_span_end = 1 - after_span_end
+        envelope = after_span_begin * before_span_end
+
+    Inputs:
+        - span_begin: tensor with shape (batch_size, sequence_length), representing a probability
```
nelson-liu: do we want double backticks around the shapes?

tests/tensors/backend_test.py
```diff
@@ -0,0 +1,22 @@
...
+import numpy
+from numpy.testing import assert_allclose
+from keras import backend as K
+
+from deep_qa.tensors.backend import cumulative_sum
+from ..common.test_case import DeepQaTestCase
+
+
+class TestBackendTensorFunctions(DeepQaTestCase):
+    def test_cumulative_sum(self):
```
nelson-liu: sorry, i know it's annoying but can you add a quick test for the masked case as well?

matt-gardner: There isn't a masked case with this - it's a backend function that just takes a tensor.  You would need another function that's `masked_cumulative_sum` in order to have a masked case.

nelson-liu: sorry, i meant for `Envelope`.

matt-gardner: I'm not so sure it's necessary there, as the mask is not used in the computation, and we just return one of the input masks.

nelson-liu: fair enough, i was meaning for it more as a sanity check but that's ok

PR #246, opened 2017-03-10
add custom score_instance function for BiDAF to return start/end span indices

nelson-liu:
This PR enables us to directly get the most likely span begin/end instance from BiDAF for use in our solver server.

### DIFF
--- a/deep_qa/models/reading_comprehension/bidirectional_attention.py
+++ b/deep_qa/models/reading_comprehension/bidirectional_attention.py
@@ -236,6 +236,7 @@ class BidirectionalAttentionFlow(TextTrainer):
         custom_objects["WeightedSum"] = WeightedSum
         return custom_objects
 
+    @overrides
     def score_instance(self, instance: CharacterSpanInstance):
         inputs, _ = self._prepare_instance(instance)
         try:
@@ -246,7 +247,8 @@ class BidirectionalAttentionFlow(TextTrainer):
             print('Inputs were: ' + str(inputs))
             raise
 
-    def get_best_span(self, span_begin_probs, span_end_probs):
+    @staticmethod
+    def get_best_span(span_begin_probs, span_end_probs):
         max_span_probability = 0
         best_word_span = (0, 1)
         begin_span_argmax = 0
--- a/deep_qa/models/reading_comprehension/bidirectional_attention.py
+++ b/deep_qa/models/reading_comprehension/bidirectional_attention.py
@@ -235,3 +235,29 @@ class BidirectionalAttentionFlow(TextTrainer):
         custom_objects["Repeat"] = Repeat
         custom_objects["WeightedSum"] = WeightedSum
         return custom_objects
+
+    def score_instance(self, instance: CharacterSpanInstance):
+        inputs, _ = self._prepare_instance(instance)
+        try:
+            span_begin_probs, span_end_probs = self.model.predict(inputs)
+            best_span, best_span_prob = self.get_best_span(span_begin_probs,
+                                                           span_end_probs)
+        except:
+            print('Inputs were: ' + str(inputs))
+            raise
+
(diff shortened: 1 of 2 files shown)


deep_qa/models/reading_comprehension/bidirectional_attention.py
```diff
@@ -235,3 +235,29 @@ def _get_custom_objects(cls):
...
+        inputs, _ = self._prepare_instance(instance)
+        try:
+            span_begin_probs, span_end_probs = self.model.predict(inputs)
+            best_span, best_span_prob = self.get_best_span(span_begin_probs,
+                                                           span_end_probs)
+        except:
+            print('Inputs were: ' + str(inputs))
+            raise
+
+    def get_best_span(self, span_begin_probs, span_end_probs):
```
matt-gardner: Can you add a test for this method?

deep_qa/models/reading_comprehension/bidirectional_attention.py
```diff
@@ -235,3 +235,29 @@ def _get_custom_objects(cls):
         custom_objects["Repeat"] = Repeat
         custom_objects["WeightedSum"] = WeightedSum
         return custom_objects
+
+    def score_instance(self, instance: CharacterSpanInstance):
```
matt-gardner: `@overrides` here

tests/models/reading_comprehension/bidirectional_attention_test.py
```diff
@@ -49,3 +50,13 @@ def test_trains_and_loads_correctly(self):
         # TODO(matt): fix the randomness that occurs here.
         # assert_allclose(model.model.predict(validation_input),
         #                 loaded_model.model.predict(validation_input))
+
+    def test_get_best_span(self):
+        # Note that the best span cannot be (1, 1) since even though
```
matt-gardner: Note that with our current training regime, this isn't actually correct.  See here: https://github.com/allenai/deep_qa/blob/master/tests/data/instances/character_span_instance_test.py#L62.  We need to change the end index to be exclusive, not inclusive.  But, so you can use the model that you've been training, you should probably change this code to assume inclusive span_end, and we can make a separate PR later to change it, and retrain the model.

tests/models/reading_comprehension/bidirectional_attention_test.py
```diff
@@ -49,3 +50,13 @@ def test_trains_and_loads_correctly(self):
...
+        # 0.3 * 0.5 is the greatest value, the end span index is constrained
+        # to occur after the begin span index.
+        span_begin_probs = numpy.array([0.1, 0.3, 0.05, 0.3, 0.25])
+        span_end_probs = numpy.array([0.5, 0.1, 0.2, 0.05, 0.15])
+        begin_end_idxs = BidirectionalAttentionFlow.get_best_span(span_begin_probs,
+                                                                  span_end_probs)
+        assert begin_end_idxs == (1, 3)
```
matt-gardner: Shouldn't this be `(1, 2)`, not `(1, 3)`?

nelson-liu: nope. The max is `0.3 (index 1 of start)` * `0.2 (index 2 of start)`. while this seems like it would be `(1, 2)`, the end index is purposefully incremented by one.

oops i just read your other comment, yes it would be `(1,2)` if we were doing inclusive span end. Does that answer your question?

matt-gardner: Ok, I'd make a big note somewhere that you're switching the model's inclusive `span_end` to be an exclusive `span_end` (and note that in the method itself _and_ in the test, please).  I'll have to think about how exactly to make this work with the envelope stuff I'm working on.  I think it's likely I'll change the training data to be exclusive, and so I'll need to find places where it breaks the code when I make the PR, so the notes will be helpful.

PR #247, opened 2017-03-10
Added Multiply and AddMask layers

matt-gardner:


deep_qa/layers/backend/add_mask.py
```diff
@@ -0,0 +1,50 @@
...
+        This is the value that we will compare to in ``compute_mask``.
+    """
+    def __init__(self, mask_value: float=0.0, **kwargs):
+        self.supports_masking = True
+        self.mask_value = mask_value
+        super(AddMask, self).__init__(**kwargs)
+
+    @overrides
+    def compute_mask(self, inputs, mask=None):  # pylint: disable=unused-argument
+        return K.cast(K.not_equal(inputs, self.mask_value), 'uint8')
```
nelson-liu: so is the final verdict that masks are `uint8`?

matt-gardner: I don't know...  But this worked, and not having a cast (using `bool`) did not.

nelson-liu: ok...I guess the mystery continues. thanks.

PR #248, opened 2017-03-10
Added the flaky pylint plugin, so that we can mark a couple of tests …

matt-gardner:
…to be re-run if they fail


nelson-liu:
this is neat. do you know how many times it'll retry, or something to that effect?

matt-gardner:
That's configurable.  By default it'll retry once, which is enough for the few flaky tests that we have that sometimes fail because of numerical instability.

PR #250, opened 2017-03-10
make BiDAF get_best_span support higher-order input

nelson-liu:
Previously, this wasn't working properly since `model.predict` outputs a np array of `shape (1,X)`, while I previously mistakenly assumed it was of `(X,)`.

So i fixed it and added tests.

### DIFF
--- a/deep_qa/models/reading_comprehension/bidirectional_attention.py
+++ b/deep_qa/models/reading_comprehension/bidirectional_attention.py
@@ -250,6 +250,14 @@ class BidirectionalAttentionFlow(TextTrainer):
 
     @staticmethod
     def get_best_span(span_begin_probs, span_end_probs):
+        if len(span_begin_probs.shape) > 2 or len(span_end_probs.shape) > 2:
+            raise ValueError("Input shapes must be (,X) or (1, X)")
+        if len(span_begin_probs.shape) == 2:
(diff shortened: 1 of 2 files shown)


deep_qa/models/reading_comprehension/bidirectional_attention.py
```diff
@@ -250,6 +250,14 @@ def score_instance(self, instance: CharacterSpanInstance):
 
     @staticmethod
     def get_best_span(span_begin_probs, span_end_probs):
+        if len(span_begin_probs.shape) > 2 or len(span_end_probs.shape) > 2:
```
DeNeutoy: Instead of doing all of this shape inference, why not do `span_begin_probs.squeeze()` and afterwards check it is the right shape. This would deal with `(1, x), (x, 1) and (x)` cases all at once.

nelson-liu: well there isn't really a `(x, 1)` case --- This check is here just to ensure that no higher-order input isn't passed in that we aren't expecting (since that would most likely be a bug somewhere else). My main rationale for separating them was to make it easier to see which things are failing from the traceback.

DeNeutoy: sure, ok.

nelson-liu:
thanks for looking at this @DeNeutoy

PR #251, opened 2017-03-11
modification to instance

bsharpataz:
small change bc I made an input file that has the actual question text in it


deep_qa/data/instances/tuple_inference_instance.py
```diff
@@ -219,7 +222,10 @@ def read_from_line(cls, line: str, default_label: int=None):
         later.
         """
         fields = line.split("\t")
-        if len(fields) == 4:
+        if len(fields) == 5:
+            question_text, index, answers_string, background_string, label = fields
```
matt-gardner: The convention is to always have the index be the first field, so if you're adding a new one, it should go after the index.

bsharpataz: ok, done

deep_qa/data/instances/tuple_inference_instance.py
```diff
@@ -219,7 +222,10 @@ def read_from_line(cls, line: str, default_label: int=None):
         later.
         """
         fields = line.split("\t")
-        if len(fields) == 4:
+        if len(fields) == 5:
+            index, question_text, answers_string, background_string, label = fields #pylint: disable=unused-variable
```
matt-gardner: If a variable is unused here, you should use an underscore like this: `index, _, ... = ...`, instead of disabling pylint.  But are you really not using part of this?  I see that you're not.  So you should probably note that in the docstring above.

deep_qa/data/instances/tuple_inference_instance.py
```diff
@@ -193,8 +193,11 @@ def index_tuples(self, tuples_in: List[TextTuple], data_indexer: DataIndexer):
     @classmethod
     def read_from_line(cls, line: str, default_label: int=None):
         """
-        Reads a TupleInstances from a line.  The format has one option (currently):
-            [question index][tab][all question+answer tuples][tab][background tuples][tab][label]
+        Reads a TupleInferenceInstance from a line.  The format has two options (currently):
+            Option 1:
+                [question index][tab][all question+answer tuples][tab][background tuples][tab][label]
+            Option 2:
+                same as option 1, but preceeded by [question text][tab]
```
matt-gardner: Comment needs to be updated.

tests/models/reading_comprehension/bidirectional_attention_test.py
```diff
@@ -1,12 +1,14 @@
 # pylint: disable=no-self-use,invalid-name
 import numpy
 from numpy.testing import assert_allclose
+import flaky
```
matt-gardner: `from flaky import flaky`

PR #253, opened 2017-03-11
Made TimeDistributed not squeeze the resulting tensor if you don't want it to

matt-gardner:


nelson-liu:
do you think a test is necessary for this?

matt-gardner:
I'll just promise that it'll get tested in my multiple choice BiDAF PR?  I use it there, and these methods will all get called.

nelson-liu:
sure sounds good

PR #254, opened 2017-03-13
quality of life fixes for data functions

nelson-liu:
Since i have nothing better to do on a sunday night, I figured i might as well look into why it takes so long to process data before we can train (inspired by a conversation with Mark on friday afternoon).

To start off, I refactored the `pad_sequence_to_length` function and made it faster by using `list.extend` versus repeated `list.append`. While the amortized runtime of `list.append` is `O(1)`, you lose a ton of efficiency by going back and forth from the python / C levels. On the other hand, `list.extend` does all of the iteration in C, which makes it a lot faster.

I'll also look into refactoring how we make word + character paddings, because that's also torturously slow right now.

Benchmarks will come in a separate comment.

### DIFF
--- a/deep_qa/data/instances/instance.py
+++ b/deep_qa/data/instances/instance.py
@@ -334,14 +334,12 @@ class IndexedInstance(Instance):
         much of the question set up. If you want to truncate from the other
         direction, you can.
         """
-
-        padded_sequence = []
-        for _ in range(desired_length):
-            padded_sequence.append(default_value())
-        sequence_length = min(len(sequence), desired_length)
-        if sequence_length != 0:
-            if truncate_from_right:
-                padded_sequence[-sequence_length:] = sequence[-sequence_length:]
-            else:
-                padded_sequence[:sequence_length] = sequence[:sequence_length]
-        return padded_sequence
+        if truncate_from_right:
+            truncated_or_extended = sequence[:desired_length]
+        else:
+            truncated_or_extended = sequence[-desired_length:]
+        if len(truncated_or_extended) < desired_length:
+            truncated_or_extended.extend([default_value()] *
+                                         (desired_length -
+                                          len(truncated_or_extended)))
+        return truncated_or_extended
--- a/deep_qa/data/instances/instance.py
+++ b/deep_qa/data/instances/instance.py
@@ -17,6 +17,7 @@ all of the concrete ``Instance`` types will have both a ``TextInstance`` and a
 corresponding ``IndexedInstance``, which you can see in the individual files
 for each ``Instance`` type.
 """
+import itertools
 from typing import Any, Callable, Dict, List
 
 from ..tokenizers import tokenizers
@@ -291,9 +292,24 @@ class IndexedInstance(Instance):
         padded_word_sequence = IndexedInstance.pad_sequence_to_length(
                 word_sequence, lengths['num_sentence_words'], default_value, truncate_from_right)
         if 'num_word_characters' in lengths:
-            padded_word_sequence = [IndexedInstance.pad_sequence_to_length(
-                    word, lengths['num_word_characters'], truncate_from_right=False)
-                                    for word in padded_word_sequence]
+            desired_length = lengths['num_word_characters']
+            longest_word = max(padded_word_sequence, key=len)
+            if desired_length > len(longest_word):
+                # since we want to pad to greater than the longest word, we add a
+                # "dummy word" to get the speed of itertools.zip_longest
+                padded_word_sequence.append([0]*desired_length)
+            # pad the list of lists to the longest sublist, appending 0's
+            words_padded_to_longest = list(zip(*itertools.zip_longest(*padded_word_sequence,
+                                                                      fillvalue=0)))
+            if desired_length > len(longest_word):
+                # now we remove the "dummy word" if we appended one.
+                words_padded_to_longest.pop()
+
+            # now we need to truncate all of them to our desired length.
+            # since truncate_from_right is always False, we chop off starting from
+            # the right.
+            padded_word_sequence = [list(word[:desired_length])
+                                    for word in words_padded_to_longest]
         return padded_word_sequence
 
     @staticmethod
--- a/deep_qa/data/instances/instance.py
+++ b/deep_qa/data/instances/instance.py
@@ -335,11 +335,19 @@ class IndexedInstance(Instance):
         direction, you can.
         """
         if truncate_from_right:
-            truncated_or_extended = sequence[:desired_length]
+            truncated = sequence[-desired_length:]
         else:
-            truncated_or_extended = sequence[-desired_length:]
-        if len(truncated_or_extended) < desired_length:
-            truncated_or_extended.extend([default_value()] *
-                                         (desired_length -
-                                          len(truncated_or_extended)))
-        return truncated_or_extended
+            truncated = sequence[:desired_length]
+        if len(truncated) < desired_length:
+            # If the length of the truncated sequence is less than the desired
+            # length, we need to pad.
+            padding_sequence = [default_value()] * (desired_length - len(truncated))
+            if truncate_from_right:
+                # When we truncate from the right, we add zeroes to the front.
+                padding_sequence.extend(truncated)
+                return padding_sequence
+            else:
+                # When we do not truncate from the right, we add zeroes to the end.
+                truncated.extend(padding_sequence)
+                return truncated
+        return truncated
--- a/deep_qa/data/tokenizers/word_splitter.py
+++ b/deep_qa/data/tokenizers/word_splitter.py
@@ -105,7 +105,7 @@ class SpacyWordSplitter(WordSplitter):
 
     @overrides
     def split_words(self, sentence: str) -> List[str]:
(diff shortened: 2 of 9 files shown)


nelson-liu:
```
In [1]: from typing import Any, Callable, Dict, List

In [2]:     def old_pad_sequence_to_length(sequence: List,
   ...:                                desired_length: int,
   ...:                                default_value: Callable[[], Any]=lambda: 0,
   ...:                                truncate_from_right: bool=True) -> List:
   ...:
   ...:         padded_sequence = []
   ...:         for _ in range(desired_length):
   ...:             padded_sequence.append(default_value())
   ...:         sequence_length = min(len(sequence), desired_length)
   ...:         if sequence_length != 0:
   ...:             if truncate_from_right:
   ...:                 padded_sequence[-sequence_length:] = sequence[-sequence_length:]
   ...:             else:
   ...:                 padded_sequence[:sequence_length] = sequence[:sequence_length]
   ...:         return padded_sequence
   ...:

In [3]:     def new_pad_sequence_to_length(sequence: List,
   ...:                                    desired_length: int,
   ...:                                    default_value: Callable[[], Any]=lambda: 0,
   ...:                                    truncate_from_right: bool=True) -> List:
   ...:         if truncate_from_right:
   ...:             truncated_or_extended = sequence[:desired_length]
   ...:         else:
   ...:             truncated_or_extended = sequence[-desired_length:]
   ...:         if len(truncated_or_extended) < desired_length:
   ...:             truncated_or_extended.extend([default_value()] *
   ...:                                          (desired_length -
   ...:                                           len(truncated_or_extended)))
   ...:         return truncated_or_extended
   ...:

In [4]: %timeit old_pad_sequence_to_length([0]*50, 700)
1000 loops, best of 3: 239 µs per loop

In [5]: %timeit new_pad_sequence_to_length([0]*50, 700)
100000 loops, best of 3: 7.66 µs per loop

In [6]: %timeit old_pad_sequence_to_length([0]*300, 700)
1000 loops, best of 3: 241 µs per loop

In [7]: %timeit new_pad_sequence_to_length([0]*300, 700)
100000 loops, best of 3: 8.88 µs per loop
```

so, looks like it's ~30 times faster to do it this way...ran this on a p2.xlarge instance, since that's what we're running the models on anyway.

nelson-liu:
hmm, perhaps the method doesn't work as I thought it did. it's really odd to me that setting the truncation mode would affect the end to which padding is added --- is this intended?
(this is the current behavior):
```
In [4]: pad_sequence_to_length([1,2,3,4,5], 2)
Out[4]: [4, 5]

In [5]: pad_sequence_to_length([1,2,3,4,5], 8)
Out[5]: [0, 0, 0, 1, 2, 3, 4, 5]

In [6]: pad_sequence_to_length([1,2,3,4,5], 8, truncate_from_right=False)
Out[6]: [1, 2, 3, 4, 5, 0, 0, 0]

In [7]: pad_sequence_to_length([1,2,3,4,5], 2, truncate_from_right=False)
Out[7]: [1, 2]
```

I think it would make a lot more sense to either pad on the left or on the right always, and you could probably make this a constructor param. thoughts? If there's a willful reason behind this, I'd be interested in hearing it.

In addition, i would have thought that `truncate_from_right=True` would mean that you would start chopping off the array starting from the right, when in fact the opposite is true...

Regardless, i've modified this PR to have the current behavior. Running the benchmarks again on the new code yields results that are almost the same as before.

nelson-liu:
also, is there a reason why we wrote our own tokenization / string processing functions? Those are also a _major_ bottleneck...why not use something like NLTK?

nelson-liu:
oh interesting, just saw this https://github.com/allenai/deep_qa/blob/master/deep_qa/data/tokenizers/word_splitter.py#L20

> Does really simple tokenization. NLTK was too slow, so we wrote our own simple tokenizer instead.

nelson-liu:
looks like spaCy blows them both out of the water (this is on a 569 token string). Perhaps it would be worth trying out a spaCytokenizer? Should be fairly trivial to implement.

```
In [15]: %timeit [token for token in en_nlp.tokenizer(lorem.lower())]
1000 loops, best of 3: 522 µs per loop

In [16]: %timeit SimpleWordSplitter().split_words(lorem.lower())
100 loops, best of 3: 5.63 ms per loop

In [17]: %timeit NltkWordSplitter().split_words(lorem.lower())
100 loops, best of 3: 12.7 ms per loop
```

matt-gardner:
It's been on my TODO for a long time to look into this.  Thanks for doing it.  I haven't actually looked at the code yet, because it doesn't look like you want a review yet; just responding to your PR comments.

It's actually pretty important to keep the option to pad from either side.  Sometimes you expect the most important content to come at the end of a sequence, so if you have to shorten it, you should throw away the beginning (this is true of question text).  Other times you want to throw out the end (this is true of news articles in WDW).

And, yeah, feel free to add the spacy tokenizer; looks like it's a faster, better option than what I wrote.  I also thought I did a _little_ better than just 2x NLTK, but I guess not... =)

matt-gardner:
I also think the location of the 0's might impact Keras' LSTM, though I'm not sure about that - Mark might know something more there.

nelson-liu:
This is ready for a review. SpaCy's tokenizer is a lot faster, but it also is very conservative. E.g. tokenizing `Mr. John Smith` will give `[Mr., John, Smith]` as expected, but tokenizing `mr. john smith` will give `[mr, ., john, smith]`. This means that the max lengths can be quite a bit larger than usual. I mentioned this on their repo, hoping to get a response. I think it's a worthy addition anyway, but I didn't set it as the default. if it had the same behavior as NLTK or something i would.

matt-gardner:
What if you tokenize, then lower?  Does it work alright?

nelson-liu:
yeah, that would work but i'm not sure if we'd lose the speed benefit (which is the reason we're using it in the first place) by going through and lowering everything again. i suppose ill run some benchmarks.

deep_qa/data/tokenizers/word_splitter.py
```diff
@@ -84,13 +84,30 @@ class NltkWordSplitter(WordSplitter):
...
         from nltk.tokenize import word_tokenize
         return word_tokenize(sentence.lower())
 
 
+class SpacyWordSplitter(WordSplitter):
+    """
+    A tokenizer that uses spaCy's Tokenizer, which is much faster than the others.
+    """
+    def __init__(self):
+        # Import is here it's slow, and can be unnecessary.
```
matt-gardner: You need a "because" or an "as" here.

nelson-liu:
surprisingly, turns out to be almost the same. i suppose i'll do that.

deep_qa/data/instances/instance.py
```diff
@@ -334,14 +350,20 @@ def pad_sequence_to_length(sequence: List,
...
-        sequence_length = min(len(sequence), desired_length)
-        if sequence_length != 0:
+        if truncate_from_right:
+            truncated = sequence[-desired_length:]
+        else:
+            truncated = sequence[:desired_length]
+        if len(truncated) < desired_length:
+            # If the length of the truncated sequence is less than the desired
+            # length, we need to pad.
+            padding_sequence = [default_value()] * (desired_length - len(truncated))
```
matt-gardner: This does not work for characters, right?

```python
>>> l = [[]]*5
>>> l[0].append(2)
>>> l
[[2], [2], [2], [2], [2], [2]]
```

If this version passes tests, we need to add another test.

Or does it pass because this will always be just padding, so it's ok?  But still, you'd end up with something way too long, right?  Hmm...

matt-gardner: I guess, now having looked at your character changes, but it should matter for background sentences, I think...

nelson-liu: the characters stuff doesn't even use this code anymore, though?

deep_qa/data/tokenizers/word_splitter.py
nelson-liu: indeed, thanks.

deep_qa/data/instances/instance.py
nelson-liu: ah ok. hmm, can you provide me with a test for background sentences?

matt-gardner: Oh, wait, no, it's here: https://github.com/allenai/deep_qa/blob/a71589144f11fbecd82040f29624082569e68f12/deep_qa/data/instances/instance.py#L289.  Background instances don't use it.  We certainly have tests that check for padding with multiple words (https://github.com/allenai/deep_qa/blob/a71589144f11fbecd82040f29624082569e68f12/tests/data/instances/text_instance_test.py#L60-L64).  I guess this is passing because it's just the padding instances that get duplicated...  Still, could you change that test to use 5 for `num_sentence_words`, instead of 3?  That will show whether this is actually an issue.

nelson-liu:
on WDW, the max passage length is 3090 with NLTK / SimpleWordTokenizer and 3539 with SpaCy. hmm...

matt-gardner:
You might try actually inspecting that passage - could be some rare case that we don't really care about anyway, and we should just be truncating.

PR #257, opened 2017-03-13
Optimize AttentionSumReader layers

nelson-liu:
While waiting for aristo-eng office hours, I decided to switch K.repeat_elements to K.tile in the OptionAttentionSum layer

### DIFF
--- a/deep_qa/layers/option_attention_sum.py
+++ b/deep_qa/layers/option_attention_sum.py
@@ -88,16 +88,16 @@ class OptionAttentionSum(Layer):
         # same thing for `document_probababilities` to select the probability
         # values corresponding to the words in the options.
         expanded_indices = K.expand_dims(K.expand_dims(document_indices, 1), 1)
-        tiled_indices = K.tile(expanded_indices, [1, K.int_shape(options)[1],
-                                                  K.int_shape(options)[2], 1])
+        tiled_indices = K.tile(expanded_indices,
+                               [1, K.int_shape(options)[1], K.int_shape(options)[2], 1])
 
         expanded_probabilities = K.expand_dims(K.expand_dims(document_probabilities, 1), 1)
         tiled_probabilities = K.tile(expanded_probabilities,
-                                     [1, K.int_shape(options)[1],
-                                      K.int_shape(options)[2], 1])
+                                     [1, K.int_shape(options)[1], K.int_shape(options)[2], 1])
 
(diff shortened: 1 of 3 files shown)


deep_qa/layers/option_attention_sum.py
```diff
@@ -88,18 +88,16 @@ def call(self, inputs, mask=None):
         # same thing for `document_probababilities` to select the probability
         # values corresponding to the words in the options.
         expanded_indices = K.expand_dims(K.expand_dims(document_indices, 1), 1)
-        tiled_indices = K.repeat_elements(K.repeat_elements(expanded_indices,
-                                                            K.int_shape(options)[1], axis=1),
-                                          K.int_shape(options)[2], axis=2)
+        tiled_indices = K.tile(expanded_indices, [1, K.int_shape(options)[1],
```
matt-gardner: I'd format this as:
```python
tiled_indices = K.tile(expanded_indices,
                       [1, K.int_shape(options)[1], K.int_shape(options)[2], 1])
```

deep_qa/layers/option_attention_sum.py
```diff
@@ -88,18 +88,16 @@ def call(self, inputs, mask=None):
...
+                                                  K.int_shape(options)[2], 1])
 
         expanded_probabilities = K.expand_dims(K.expand_dims(document_probabilities, 1), 1)
-        tiled_probabilities = K.repeat_elements(K.repeat_elements(expanded_probabilities,
-                                                                  K.int_shape(options)[1], axis=1),
-                                                K.int_shape(options)[2], axis=2)
+        tiled_probabilities = K.tile(expanded_probabilities,
```
matt-gardner: I'd do the same formatting here.

nelson-liu:
@matt-gardner, i turned this into a general PR for sprucing up the layers used in the attention sum reader. will do the same for the GAReader in a bit (maybe after doing some more search stuff)

matt-gardner:
Looks like theano failed on a flaky test - can you decorate that test?

matt-gardner:
And did we ever figure out if this is actually an optimization?  If this is actually slower, we should probably not make the change.

matt-gardner:
If you think it is indeed better, then feel free to merge.

nelson-liu:
eh, i'm pretty indifferent. i think i like the semantics of using `K.tile` more, but it's still quite confusing as to why it would be slower.

PR #258, opened 2017-03-14
Multiple choice BiDAF

matt-gardner:
Finally got this working.


matt-gardner:
Well, the multiple choice BiDAF test passes now (on tensorflow, at least), but the tensorboard logs test is crashing.  Any idea what's going on?

matt-gardner:
Finally passing CI...  =)

tests/models/multiple_choice_qa/multiple_choice_bidaf_test.py
```diff
@@ -0,0 +1,38 @@
...
+        args = {
+                'bidaf_params': bidaf_model_params,
+                'train_bidaf': False,
+                'similarity_function': {'type': 'linear', 'combination': 'x,y'},
+                }
+        self.write_who_did_what_files()
+        model, _ = self.ensure_model_trains_and_loads(MultipleChoiceBidaf, args)
+        # All of the params come from the linear similarity function in the attention layer,
+        # because we set `train_bidaf` to `False`.
+        assert sum([K.count_params(p) for p in model.model.trainable_weights]) == 41
```
nelson-liu: not sure if my comment posted, but how did you get 41 here?

matt-gardner: I looked at the output of the model summary...  =).  And then I checked that the number made sense.  It's a linear function, with a number of parameters equal to `|x| + |y| + 1` (`+ 1` because of the bias).  `|x|` here is `32`, because that's what you end up with in BiDAF, and `|y|` is `8`, because it's a bidirectional encoder with embedding size 4, I believe...

PR #260, opened 2017-03-15
Porting to Keras 2

nelson-liu:
I did some work wrt porting deep_qa to Keras 2... 

but there's still a lot of work to be done, off the top of my head:
- Model loading changed, so we need to fix that
- ShareableGRU was broken, so we need to fix that
- Make _sure_ the timedistributed stuff we're doing still works

This is a pretty massive PR, sorry in advance to whoever's reviewing. I tried to make my commit messages clear / have each commit message correspond to one "type" of change. Hopefully that makes it slightly easier to look at this.


deep_qa/data/tokenizers/word_and_character_tokenizer.py
```diff
@@ -104,22 +103,23 @@ def merge_shape(input_shapes):
...
         if ':' in input_name:
             input_name = input_name.split(':')[0]
         if input_name.split('_')[-1].isdigit():
             input_name = '_'.join(input_name.split('_')[:-1])
-        final_embedded_input = merge([word_embedding, word_encoding],
-                                     mode=merge_mode,
-                                     output_shape=merge_shape,
-                                     output_mask=merge_mask,
-                                     name='combined_word_embedding_for_' + input_name)
+        name = 'combined_word_embedding_for_' + input_name
```
nelson-liu: looks like this never gets used anywhere, did you mean to put it as the `name` argument of the `Lambda?`

nelson-liu:
great, theano tests officially passing

matt-gardner:
This is no longer WIP.  Nelson, if you want to take a look, now's the time.  Also, do you know how to exclude `contrib/` from the coverage results?

nelson-liu:
yeah, i don't think it's possible for me to do a thorough review here. I have faith in our unit tests, so let's merge. I'd be in favor of rerunning some of the models in `example_experiments` to ensure that we can still replicate the results of those papers.

nelson-liu:
(feel free to press the button when you're ready)

nelson-liu:
we should also probably edit the `example_experiments`, the biggest thing that I can think of is that `output_dim` is now `units`

PR #261, opened 2017-03-15
Made span end exclusive in CharacterSpanInstance

matt-gardner:


deep_qa/data/instances/character_span_instance.py
```diff
@@ -25,6 +25,8 @@ class CharacterSpanInstance(QuestionPassageInstance):
     This class should be used to represent training instances for the SQuAD (Stanford Question
     Answering) and NewsQA datasets, to name a few.
     """
+    stop_token = "@@STOP@@"
```
nelson-liu: can you add a comment explaining the purpose of the stop token?

PR #262, opened 2017-03-16
Made embedding_dim configurable based on vocab

matt-gardner:
I also changed `embedding_size` to `embedding_dim`, which is why this PR touched so many files.  The main difference is in `text_trainer.py`.  And while I was at it, I updated a couple of tests to add a new convenience method I added a few PRs ago, and that revealed some problems in `_set_max_lengths_from_model`, which I also fixed.

And I removed a test that I didn't think was necessary anymore, because we don't really use the multiple choice memory network - I didn't want to do the work necessary to make the test pass with this new set up.


nelson-liu:
> And I removed a test that I didn't think was necessary anymore, because we don't really use the multiple choice memory network - I didn't want to do the work necessary to make the test pass with this new set up.

So the multiple choice memory  network is broken, or just the test?

matt-gardner:
The memory network code is at the moment incompatible with the words and characters tokenizer, if you use a bag of words encoder, because of encoding dimension stuff.  It still works with other tokenizers, as the tests I left in show.  I figured it wasn't worth the time to make that old code compatible with words and characters tokenizers.

nelson-liu:
> The memory network code is at the moment incompatible with the words and characters tokenizer, if you use a bag of words encoder, because of encoding dimension stuff. 

ok, that should be mentioned somewhere appropriate...not sure where though

deep_qa/models/multiple_choice_qa/question_answer_similarity.py
```diff
@@ -75,9 +60,15 @@ def _build_model(self):
...
         hidden_input = regularized_encoded_question
-        for layer in self.hidden_layers:
-            hidden_input = layer(hidden_input)
-        projected_input = self.projection_layer(hidden_input)
+        for i in range(self.num_hidden_layers):
+            hidden_layer = Dense(output_dim=self.hidden_layer_width,
+                                 activation=self.hidden_layer_activation,
+                                 name='question_hidden_layer_%d' % i)
+            hidden_input = hidden_layer(hidden_input)
+        projection_layer = Dense(output_dim=K.int_shape(encoded_answers)[-1],
```
nelson-liu: im surprised you can call `K.int_shape` here

matt-gardner: It's just a tensor.  You can't call it on a layer object, but it's the same as calling this on an input tensor inside a layer, because this is what shows up in the input.

If I were doing anything more complicated than getting an `int` out of the tensor, and trying to use it somehow here, there would have been problems.

matt-gardner:
I also just realized I forgot to update all of the example experiments to use the new parameter specification.  I'll do that in another PR, in which I'm also updated `bidaf_squad.json` to use more sane parameters.

PR #263, opened 2017-03-16
Debug output

bsharpataz:
I plan to add debug output for some other layers and also I'd like to print some stats-y stuff.  In particular, I would like to print a "score" that represents how skewed/peaky/flat the distributions are for a given answer_tuple across options after the first noisy-or layer.

I figured, though, that that could all be a next PR? or I can add that stuff here and now - whichever!

### DIFF
--- a/deep_qa/data/instances/tuple_inference_instance.py
+++ b/deep_qa/data/instances/tuple_inference_instance.py
@@ -209,7 +209,7 @@ class TupleInferenceInstance(TextInstance):
                 [question index][tab][all question+answer tuples][tab][background tuples][tab][label]
             Option 2:
                 same as option 1, but [question text][tab] comes immediately after the question index,
-                following the tab. 
+                following the tab.
         The question+answer tuples are formatted as:
             [a_1-tuple_1]$$$...$$$[a_1-tuple_n]###...###[a+_m-tuple_1]$$$...$$$[a_m-tuple_p]
         such that ``$$$`` serves as the tuple delimiter and ``###`` serves as the answer candidate
--- a/deep_qa/data/instances/tuple_inference_instance.py
+++ b/deep_qa/data/instances/tuple_inference_instance.py
@@ -20,13 +20,14 @@ class TextTuple:
         self.location = location
         self.source = source
 
-    def display_string(self):
+    def display_string(self, context_limit=None):
         # NOTE: not currently displaying location, time, and source info.
-        return ("TextTuple( [\n\tsubj: {0} \n\tverb: {1} \n\tobjects: {2} " +
-                "\n\tcontext: {3}\n] )").format(self.subject,
-                                                self.verb,
-                                                ", ".join(self.objects),
-                                                self.context)
+        if context_limit is None:
+            context_limit = len(self.context)
+        return ("(S: {0}, V: {1}, O: {2}, C: {3})").format(self.subject,
+                                                           self.verb,
+                                                           ", ".join(self.objects),
+                                                           self.context[:context_limit])
 
     def to_text_list(self, object_handling: str="collapse", include_context: bool=True):
         '''
@@ -119,6 +120,9 @@ class TupleInferenceInstance(TextInstance):
     background_tuples: List[TextTuple]
         This is a list of background ``TextTuples`` (currently used for all answer candidates).
 
+    question_test: str, default=None
+        The original text of the question, if available.
+
     label: int, default=None
         The class label (i.e. the index of the correct multiple choice answer) -- corresponds to the
         indices in self.answer_tuples.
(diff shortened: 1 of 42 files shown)


deep_qa/data/instances/tuple_inference_instance.py
```diff
@@ -119,6 +120,9 @@ class TupleInferenceInstance(TextInstance):
     background_tuples: List[TextTuple]
         This is a list of background ``TextTuples`` (currently used for all answer candidates).
 
+    question_test: str, default=None
```
matt-gardner: `s/test/text/`

deep_qa/data/instances/tuple_inference_instance.py
```diff
@@ -130,18 +134,25 @@ class TupleInferenceInstance(TextInstance):
...
     def __init__(self, answer_tuples: List[List[TextTuple]], background_tuples: List[TextTuple],
-                 label: int=None, index: int=None):
+                 question_text: str=None, label: int=None, index: int=None):
         super(TupleInferenceInstance, self).__init__(label, index)
         self.answer_tuples = answer_tuples
         self.background_tuples = background_tuples
+        self.question_text = question_text
 
     def display_string(self):
         to_return = 'MultipleTrueFalseTuplesWithBackgroundInstance: \n'
```
matt-gardner: This should be `TupleInferenceInstance: \n`, right?

deep_qa/models/multiple_choice_qa/tuple_inference.py
```diff
@@ -175,3 +177,99 @@ def _build_model(self):
         final_output = MaskedSoftmax()(options_probabilities)
 
         return DeepQaModel(input=[question_input, background_input], output=[final_output])
+
+    def debug_display_instance_tuple_match_scores(self, answer_tuples, background_tuples, tuple_matcher_output,
```
matt-gardner: Minor point, feel free to ignore.  I would probably put this method below the main `_instance_debug_output`.  Also, the method name should probably begin with an underscore, and giving types on the input arguments would help readability here.

deep_qa/models/multiple_choice_qa/tuple_inference.py
```diff
@@ -175,3 +177,99 @@ def _build_model(self):
...
+                                                  instance, num_to_display):
+        num_background = min(self.num_background_tuples, len(instance.background_tuples))
+        result = ""
+        for i, answer_tuple in enumerate(answer_tuples):
+            answer_tuple_string = "\n\t".join(textwrap.wrap(answer_tuple.display_string(), 150))
+            result += "Question (repeated): %s\n" % instance.question_text
+            result += "Answer_tuple_{0} : \n\t{1}\n\n".format(i, answer_tuple_string)
+            result += "Top {0} (out of {1}) highest scoring background tuples:\n\n".format(num_to_display,
+                                                                                           num_background)
+            storage = []
```
matt-gardner: Maybe call this `tuple_match_scores`?

deep_qa/models/multiple_choice_qa/tuple_inference.py
```diff
@@ -175,3 +177,99 @@ def _build_model(self):
...
+            for j, background_tuple in enumerate(background_tuples):
+                tuple_match_score = tuple_matcher_output[i, j]
+                storage.append((tuple_match_score, j, background_tuple))
+            # Sort descending by tuple match score
+            sorted_by_score = sorted(storage, key=lambda tup: tup[0], reverse=True)[:num_to_display]
+            for scored in sorted_by_score:
+                background_tuple_index = scored[1]
+                background_tuple_string = scored[2].display_string()
+                wrapped_tuple = "\n\t".join(textwrap.wrap(background_tuple_string, 150))
```
matt-gardner: You could consider making this a parameter to the class, instead of a hard-coded number here.  Same with `num_to_display = 5` below.

deep_qa/models/multiple_choice_qa/tuple_inference.py
```diff
@@ -175,3 +177,99 @@ def _build_model(self):
...
+        result += "Num tuples per answer option: %s\n" % [len(answer) for answer in instance.answer_tuples]
+        result += "(limiting display to top %s at various levels)\n" % num_to_display
+        result += "====================================================================\n"
+
+        final_scores = []
+        index_of_chosen = None
+        for layer_name in outputs:
+            if "softmax" in layer_name:
+                final_scores = list(enumerate((outputs[layer_name])))
+                sorted_scores = sorted(final_scores, key=lambda tup: tup[1], reverse=True)
```
matt-gardner: I might call these `answer_scores`, so it's more immediately obvious what this is holding.  Took me a minute to realize what this was.  And if you drop `answer` (or `final`) once you sort them, you've lost the key content in the name, so it's hard to remember below what this variable was referring to.

deep_qa/models/multiple_choice_qa/tuple_inference.py
```diff
@@ -175,3 +177,99 @@ def _build_model(self):
...
+            result += "ERROR: no answer chosen\n"
+        elif index_of_chosen == instance.label:
+            result += "  Answered correctly!\n"
+        else:
+            result += "  Answered incorrectly\n"
+        result += "====================================================================\n"
+
+        # Output of the tuple matcher layer:
+        # shape: (num_options, num_question_tuples, num_background_tuples)
+        tuple_matcher_output = outputs.get('timedistributed_3', None)
```
matt-gardner: This probably means you need to give the outer-most `TimeDistributed` layer a name, on lines 74 and 76 above.  Then you don't have to have this complex logic, and you can just use a single name, which is specified by the model above.  These auto-generated names aren't consistent across theano and tensorflow, or even across multiple runs using the same backend (e.g., if you've run another model before this one, you'll get different numbers here).

deep_qa/models/multiple_choice_qa/tuple_inference.py
matt-gardner: Also, this is a nice descriptive name, but it's also a bit long.  Is there a shorter name that's just as good?  Maybe something like `_render_tuple_match_scores`?

deep_qa/models/multiple_choice_qa/tuple_inference.py
bsharpataz: thanks for the tip!

PR #264, opened 2017-03-17
SQuAD Sentence Selection reader + SentenceSelectionInstance

nelson-liu:
This PR adds a "reader" for taking in SQuAD and transforming it into an answer sentence selection task. Note that it's written in python --- we're trying this out.

It also adds a SentenceSelectionInstance for general use.


deep_qa/data/instances/sentence_selection_instance.py
```diff
@@ -0,0 +1,161 @@
...
+import numpy as np
+from overrides import overrides
+
+from .instance import TextInstance, IndexedInstance
+from ..data_indexer import DataIndexer
+
+
+class SentenceSelectionInstance(TextInstance):
+    """
+    A SentenceSelectionInstance an instance for the sentence selection
```
matt-gardner: Missing an "is", I think.

deep_qa/data/instances/sentence_selection_instance.py
```diff
@@ -0,0 +1,161 @@
...
+                           self.sentences]
+        for namespace in words:
+            for sentence_words in sentences_words:
+                words[namespace].extend(sentence_words[namespace])
+        return words
+
+    @overrides
+    def to_indexed_instance(self, data_indexer: DataIndexer):
+        question_indices = self._index_text(self.question_text, data_indexer)
+        sentences_indices = [self._index_text(sentence, data_indexer) for
```
matt-gardner: I think it's more readable to have the `for` on the next line.  Do you think this way is better?

deep_qa/data/instances/sentence_selection_instance.py
```diff
@@ -0,0 +1,161 @@
...
+        Reads a SentenceSelectionInstance object from a line.
+        The format has one of two options:
+
+        (1) [example index][tab][question][tab][list_of_sentences][tab][label]
+        (2) [question][tab][list_of_sentences][tab][label]
+
+        The ``list_of_sentences`` column is assumed formatted as: ``[sentence]###[sentence]###[sentence]...``
+        That is, we split on three hashes (``"###"``).
+
+        ``default_label`` is ignored, but we keep the argument to match the interface.
```
matt-gardner: We should just remove `default_label` from the interface, actually.  I removed the ability to _use_ `default_label` from the trainer code a long time ago.  But you shouldn't worry about that here; I'll open an issue for it.

deep_qa/data/instances/sentence_selection_instance.py
```diff
@@ -0,0 +1,161 @@
...
+
+        if len(fields) == 4:
+            index_string, question, sentences, label_string = fields
+            index = int(index_string)
+        elif len(fields) == 3:
+            question, sentences, label_string = fields
+            index = None
+        else:
+            raise RuntimeError("Unrecognized line format: " + line)
+        # get the answer options
```
matt-gardner: Comment seems unnecessary here.

deep_qa/data/instances/sentence_selection_instance.py
```diff
@@ -0,0 +1,161 @@
...
+        lengths = {}
+
+        # the number of sentences
+        lengths['num_sentences'] = len(self.sentences_indices)
+
+        # the number of words in the question
+        lengths['num_question_words'] = question_lengths['num_sentence_words']
+
+        # the number of words in each passage
+        lengths['num_passage_words'] = max(sentence_lengths['num_sentence_words']
```
matt-gardner: Why the switch to `passage`, where everything else in this PR uses `sentence`?

deep_qa/data/instances/sentence_selection_instance.py
```diff
@@ -0,0 +1,161 @@
...
+        In this function, we pad the questions and sentences (in terms of number
+        of words in each), the number of sentences,
+        as well as the individual words in the questions and sentences themselves.
+        """
+        max_lengths_tmp = max_lengths.copy()
+        # pad the number of words in a question
+        max_lengths_tmp['num_sentence_words'] = max_lengths_tmp['num_question_words']
+        self.question_indices = self.pad_word_sequence(self.question_indices, max_lengths_tmp)
+        # pad the number of sentences
+        num_sentences = max_lengths['num_sentences']
```
matt-gardner: Can't you use `self.pad_sequence_to_length` here?

tests/data/dataset_readers/SQUAD_SAMPLE_FILE.json
```diff
@@ -0,0 +1,34 @@
+{
```
matt-gardner: WHY IS THIS FILE ALL CAPS?  (Sorry, I couldn't resist...)

matt-gardner: Also, I would probably put this file in a separate directory from python code.  Not sure what to call it - `tests/resources/`?  Not sure what is typically done in python.  Another alternative would be to just have this json in the test itself, write it to a file in the temp test directory, then read it from that file.  I'd probably prefer that alternative, actually.  That would make the test file itself much more self-contained.

deep_qa/data/dataset_readers/squad_sentence_selection_reader.py
```diff
@@ -0,0 +1,99 @@
...
+        processed_rows = []
+        logger.info("Reading file at %s", input_filepath)
+        with open(input_filepath) as dataset_file:
+            dataset_json = json.load(dataset_file)
+            dataset = dataset_json['data']
+        for article in tqdm(dataset):
+            for paragraph in article['paragraphs']:
+                context_article = paragraph["context"]
+                # split the context_article into a list of sentences
+                sentences = nltk.sent_tokenize(context_article)
```
nelson-liu: I think it does, but it involves downloading like 1GB of files, and I don't think speed is really an issue here

deep_qa/data/instances/sentence_selection_instance.py
nelson-liu: agreed, thanks

deep_qa/data/instances/sentence_selection_instance.py
nelson-liu: ah, most of the other `get_lengths` functions have standardized to use `num_passage_words` so i was concerned that changing it would break things that interface with the instance. `sentence` should be fine though.

tests/data/dataset_readers/SQUAD_SAMPLE_FILE.json
nelson-liu: I tried putting the json in the test, but i found it to be non-trivial (since JSON doesn't allow line breaks in strings). Also, if you use the python `"""`, it keeps all of the spacing / I haven't found a function like scala's stripmargin. thoughts?

deep_qa/data/dataset_readers/squad_sentence_selection_reader.py
```diff
@@ -0,0 +1,99 @@
...
+                    question_text = question_answer["question"]
+                    answer_start_idx = question_answer["answers"][0]["answer_start"]
+                    # Get the full sentence corresponding to the answer
+                    answer_sentence = None
+                    for span_tup in span_to_sentence_idx:
+                        start_span, end_span = span_tup
+                        if start_span <= answer_start_idx and answer_start_idx < end_span:
+                            answer_sentence = span_to_sentence_idx[span_tup]
+                            break
+                    if answer_sentence is None:
```
nelson-liu: interesting, never knew this was a thing. thanks

tests/data/dataset_readers/SQUAD_SAMPLE_FILE.json
matt-gardner: But the spacing shouldn't matter for json processing, right?  And for really long strings that going into a `"""` block, you can just define them above, and reference them as variables.  Would that work?

deep_qa/data/dataset_readers/squad_sentence_selection_reader.py
```diff
@@ -3,6 +3,8 @@
 import logging
 import os
 import random
+
+import nltk
```
matt-gardner: Why move it back?  Because of pylint?  You could disable the pylint warning.  I guess the real question is whether the slow import at the top matters.  Probably you would be running this script stand-alone, and not importing it, and so the slow import doesn't matter.

deep_qa/data/dataset_readers/squad_sentence_selection_reader.py
```diff
@@ -39,52 +39,52 @@ def read_file(self, input_filepath):
...
-                        start_span, end_span = span_tup
-                        if start_span <= answer_start_idx and answer_start_idx < end_span:
-                            answer_sentence = span_to_sentence_idx[span_tup]
+                    for span_tuple in span_to_sentence_index:
+                        start_span, end_span = span_tuple
+                        if start_span <= answer_start_index and answer_start_index < end_span:
+                            answer_sentence = span_to_sentence_index[span_tuple]
                             break
-                    if answer_sentence is None:
+                    else:
```
matt-gardner: I do think the `  # no break` comment is helpful for reading this, especially for people who aren't familiar with this python quirk.

deep_qa/data/instances/sentence_selection_instance.py
```diff
@@ -134,20 +133,21 @@ def pad(self, max_lengths: Dict[str, int]):
         as well as the individual words in the questions and sentences themselves.
         """
         max_lengths_tmp = max_lengths.copy()
-        # pad the number of words in a question
-        max_lengths_tmp['num_sentence_words'] = max_lengths_tmp['num_question_words']
-        self.question_indices = self.pad_word_sequence(self.question_indices, max_lengths_tmp)
-        # pad the number of sentences
+        # Pad the number of sentences.
         num_sentences = max_lengths['num_sentences']
```
matt-gardner: You didn't want to use `pad_sequence_to_length`?

deep_qa/data/instances/sentence_selection_instance.py
```diff
@@ -134,20 +133,21 @@ def pad(self, max_lengths: Dict[str, int]):
...
         while len(self.sentences_indices) < num_sentences:
             self.sentences_indices.append([])
         self.sentences_indices = self.sentences_indices[:num_sentences]
 
-        # pad the number of words in a sentence
-        max_lengths_tmp['num_sentence_words'] = max_lengths_tmp['num_passage_words']
+        # Pad the number of words in a sentence.
+        # Note that max_lengths_tmp['num_sentence_words'] is set to
+        # ['num_sentence_words'] by default.
```
matt-gardner: I get that you're trying to say that you're using the key defined by `TextTrainer` here, but I don't think the comment makes much sense as-is.

tests/data/dataset_readers/SQUAD_SAMPLE_FILE.json
nelson-liu: yeah, the thing that i was struggling with is that json doesn't allow line breaks within a field, unless you use `\n`, which means that I'd have to have the passage all on one line. i think it works to define them above and reference them as variables, though...

deep_qa/data/dataset_readers/squad_sentence_selection_reader.py
nelson-liu: well having it in the `readFile` method actually doesnt do anything --- it still has to be imported always, since when pylint collects tests it imports everything from every test, and then only runs certain tests.

deep_qa/data/instances/sentence_selection_instance.py
nelson-liu: oops, forgot to change it. thanks for the reminder.

tests/data/instances/sentence_selection_instance_test.py
```diff
@@ -95,18 +95,18 @@ def test_pad_adds_empty_sentences(self):
                            'num_sentence_words': 2,
                            'num_sentences': 4})
         assert self.instance.question_indices == [3, 5, 6]
-        assert self.instance.sentences_indices[0] == [3, 4]
-        assert self.instance.sentences_indices[1] == [5, 1]
-        assert self.instance.sentences_indices[2] == [5, 6]
-        assert self.instance.sentences_indices[3] == [0, 0]
+        assert self.instance.sentences_indices[0] == [0, 0]
```
matt-gardner: I probably would have changed the `truncate` parameter, instead of changing the test, but feel free to leave it if you want.

tests/data/dataset_readers/SQUAD_SAMPLE_FILE.json
nelson-liu: yeah, json doesn't even like linebreaks not within a field...I think leaving it as a separate file is better for now

PR #268, opened 2017-03-20
Baseline SiameseSentenceSelection Model

nelson-liu:
This PR implements a baseline siamese style architecture for sentence selection.

We basically encode the question and each candidate sentence with some encoder, then take the cosine similarity and normalize to get a probability distribution over candidate sentences.

### DIFF
--- a/deep_qa/models/sentence_selection/siamese_sentence_selector.py
+++ b/deep_qa/models/sentence_selection/siamese_sentence_selector.py
@@ -1,0 +1,115 @@
+from typing import Any, Dict
+from overrides import overrides
+from keras.layers import Input
+
+from ...data.instances.sentence_selection_instance import SentenceSelectionInstance
+from ...layers.attention.attention import Attention
+from ...layers.wrappers.encoder_wrapper import EncoderWrapper
+from ...training.text_trainer import TextTrainer
+from ...training.models import DeepQaModel
+
+
+class SiameseSentenceSelector(TextTrainer):
+    def __init__(self, params: Dict[str, Any]):
+        self.max_question_length = params.pop('max_question_length', None)
+        self.max_sentence_length = params.pop('max_sentence_length', None)
+        self.num_sentences = params.pop('num_sentences', None)
+        super(SiameseSentenceSelector, self).__init__(params)
+
+    @overrides
+    def _build_model(self):
+        """
+        The basic outline here is that we'll pass the questions and each
+        sentence in the passage through some sort of encoder. At the simplest,
+        this could be a biLSTM or a biGRU.
+
+        Then, we take the encoded representation of the question and calculate
+        a cosine similarity with the encoded representation of each sentence in
+        the passage, to get a tensor of cosine similarities with shape
+        (batch_size, num_sentences_per_passage). We then normalize for each
+        batch to get a probability distribution over sentences in the passage.
+        """
+        # First we create input layers and pass the inputs through embedding layers.
+        # shape: (batch size, question_length)
+        question_input = Input(shape=self._get_sentence_shape(self.max_question_length),
+                               dtype='int32', name="question_input")
+
+        # shape: (batch size, num_sentences, sentences_length)
+        sentences_input_shape = ((self.num_sentences,) +
+                                 self._get_sentence_shape(self.max_sentence_length))
+        sentences_input = Input(shape=sentences_input_shape,
+                                dtype='int32', name="sentences_input")
+
+        # shape: (batch size, question_length, embedding size)
+        question_embedding = self._embed_input(question_input)
+
+        # shape: (batch size, num_sentences, sentences_length, embedding size)
+        sentences_embedding = self._embed_input(sentences_input)
+
+        # We encode the question embeddings with some encoder.
+        question_encoder = self._get_encoder(name="question_encoder")
+        # shape: (batch size, encoder_output_dimension)
+        encoded_question = question_encoder(question_embedding)
+
+        # We encode the document embeddings with some encoder.
+        sentences_encoder = EncoderWrapper(self._get_encoder(name="sentence_encoder"),
+                                           name="TimeDistributed_sentences_encoder")
+        # shape: (batch size, num_sentences, encoder_output_dimension)
+        encoded_sentences = sentences_encoder(sentences_embedding)
+
+        # Here we use the Attention layer with the cosine similarity function
+        # to get the cosine similarities of each sesntence with the question.
+        # shape: (batch size, num_sentences)
+        attention_name = 'question_sentences_similarity'
+        similarity_params = {"type": "cosine_similarity"}
+        sentence_probabilities = Attention(name=attention_name,
+                                           similarity_function=similarity_params)([encoded_question,
+                                                                                   encoded_sentences])
+
+        return DeepQaModel(input=[question_input, sentences_input],
+                           output=sentence_probabilities)
+
+    @overrides
+    def _instance_type(self):
+        """
+        Return the instance type that the model trains on.
+        """
+        return SentenceSelectionInstance
+
+    @overrides
+    def _get_max_lengths(self) -> Dict[str, int]:
+        """
+        Return a dictionary with the appropriate padding lengths.
+        """
+        max_lengths = super(SiameseSentenceSelector, self)._get_max_lengths()
+        max_lengths['num_question_words'] = self.max_question_length
+        max_lengths['num_sentences'] = self.num_sentences
+        max_lengths['num_sentence_words'] = self.max_sentence_length
+        return max_lengths
+
+    @overrides
+    def _set_max_lengths(self, max_lengths: Dict[str, int]):
+        """
+        Set the padding lengths of the model.
+        """
+        # TODO(nelson): superclass complains that there is no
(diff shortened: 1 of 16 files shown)


deep_qa/models/sentence_selection/siamese_sentence_selector.py
```diff
@@ -0,0 +1,115 @@
...
+from keras.layers import Input
+
+from ...data.instances.sentence_selection_instance import SentenceSelectionInstance
+from ...layers.attention.attention import Attention
+from ...layers.wrappers.encoder_wrapper import EncoderWrapper
+from ...training.text_trainer import TextTrainer
+from ...training.models import DeepQaModel
+
+
+class SiameseSentenceSelector(TextTrainer):
```
matt-gardner: Can you add a class docstring?  And you might note in that docstring that this isn't necessarily a siamese network, because you could have different encoders for the question and sentences, but it's at least similar.

deep_qa/models/sentence_selection/siamese_sentence_selector.py
```diff
@@ -0,0 +1,115 @@
...
+        self.max_sentence_length = params.pop('max_sentence_length', None)
+        self.num_sentences = params.pop('num_sentences', None)
+        super(SiameseSentenceSelector, self).__init__(params)
+
+    @overrides
+    def _build_model(self):
+        """
+        The basic outline here is that we'll pass the questions and each
+        sentence in the passage through some sort of encoder. At the simplest,
+        this could be a biLSTM or a biGRU.
```
matt-gardner: Well, simplest is BOW...

deep_qa/models/sentence_selection/siamese_sentence_selector.py
```diff
@@ -0,0 +1,115 @@
...
+                                dtype='int32', name="sentences_input")
+
+        # shape: (batch size, question_length, embedding size)
+        question_embedding = self._embed_input(question_input)
+
+        # shape: (batch size, num_sentences, sentences_length, embedding size)
+        sentences_embedding = self._embed_input(sentences_input)
+
+        # We encode the question embeddings with some encoder.
+        question_encoder = self._get_encoder(name="question_encoder")
```
matt-gardner: Add a `fallback_behavior` here and below, probably `use_default_encoder` (that way you can have them use the same encoder if you don't specify separate ones).

deep_qa/models/sentence_selection/siamese_sentence_selector.py
```diff
@@ -0,0 +1,115 @@
...
+from ...data.instances.sentence_selection_instance import SentenceSelectionInstance
+from ...layers.attention.attention import Attention
+from ...layers.wrappers.encoder_wrapper import EncoderWrapper
+from ...training.text_trainer import TextTrainer
+from ...training.models import DeepQaModel
+
+
+class SiameseSentenceSelector(TextTrainer):
+    def __init__(self, params: Dict[str, Any]):
+        self.max_question_length = params.pop('max_question_length', None)
```
matt-gardner: Can you use `num_` instead of `max_` for these?

deep_qa/models/sentence_selection/siamese_sentence_selector.py
```diff
@@ -0,0 +1,115 @@
...
+        max_lengths['num_sentences'] = self.num_sentences
+        max_lengths['num_sentence_words'] = self.max_sentence_length
+        return max_lengths
+
+    @overrides
+    def _set_max_lengths(self, max_lengths: Dict[str, int]):
+        """
+        Set the padding lengths of the model.
+        """
+        # TODO(nelson): superclass complains that there is no
```
matt-gardner: Does it really?

deep_qa/models/sentence_selection/siamese_sentence_selector.py
```diff
@@ -0,0 +1,115 @@
...
+
+    @overrides
+    def _get_max_lengths(self) -> Dict[str, int]:
+        """
+        Return a dictionary with the appropriate padding lengths.
+        """
+        max_lengths = super(SiameseSentenceSelector, self)._get_max_lengths()
+        max_lengths['num_question_words'] = self.max_question_length
+        max_lengths['num_sentences'] = self.num_sentences
+        max_lengths['num_sentence_words'] = self.max_sentence_length
```
matt-gardner: Because `num_sentence_words` is actually a key handled by the super class, you don't need it here, or anywhere else in this class.

deep_qa/models/sentence_selection/siamese_sentence_selector.py
```diff
@@ -0,0 +1,115 @@
...
+        # num_sentence_words key, so we set it to None here.
+        # We should probably patch up / organize the API.
+        super(SiameseSentenceSelector, self)._set_max_lengths(max_lengths)
+        self.max_question_length = max_lengths['num_question_words']
+        self.num_sentences = max_lengths['num_sentences']
+        self.max_sentence_length = max_lengths['num_sentence_words']
+
+    @overrides
+    def _set_max_lengths_from_model(self):
+        self.set_text_lengths_from_model_input(self.model.get_input_shape_at(0)[1][1:])
```
matt-gardner: This should be `[2:]` at the end.

tests/common/test_case.py
```diff
@@ -269,23 +269,38 @@ def write_tuple_inference_files(self):
...
+            validation_file.write('1\tquestion 2\tpassage with perhaps the answer\t13,18\n')
         with codecs.open(self.TRAIN_FILE, 'w', 'utf-8') as train_file:
             train_file.write('1\tquestion 5\tpassage5 with answer5 answer3\t14,20\n')
             train_file.write('2\tquestion 6\tpassage6 with answer6\t0,8\n')
             train_file.write('3\tquestion 7\tpassage7 with answer7\t9,13\n')
             train_file.write('4\tquestion 8\tpassage8 with answer8\t14,20\n')
 
+    def write_sentence_selection_files(self):
+        with codecs.open(self.VALIDATION_FILE, 'w', 'utf-8') as validation_file:
+            validation_file.write('1\tWhere is Paris?\tParis is the capital of France.###It '
```
matt-gardner: Can you make this have four sentences, with one of them really long?  Just to be sure padding is set correctly on model load.

deep_qa/models/sentence_selection/siamese_sentence_selector.py
```diff
@@ -0,0 +1,123 @@
...
+    model encodes the question and each sentence with (possibly different)
+    encoders, and then does a cosine similarity and normalizes to get a
+    distribution over the set of sentences.
+
+    Note that in some cases, this may not be exactly "Siamese" because the
+    question and sentences encoders can differ.
+    """
+    def __init__(self, params: Dict[str, Any]):
+        self.num_question_words = params.pop('num_question_words', None)
+        self.num_sentence_words = params.pop('num_sentence_words', None)
```
matt-gardner: This actually won't work.  You're calling this before calling the super class constructor, and the super class will override this with `None`.  You should just remove this line, because the super class handles this parameter.

deep_qa/models/sentence_selection/siamese_sentence_selector.py
```diff
@@ -0,0 +1,123 @@
...
+        batch to get a probability distribution over sentences in the passage.
+        """
+        # First we create input layers and pass the inputs through embedding layers.
+        # shape: (batch size, question_length)
+        question_input = Input(shape=self._get_sentence_shape(self.num_question_words),
+                               dtype='int32', name="question_input")
+
+        # shape: (batch size, num_sentences, sentences_length)
+        sentences_input_shape = ((self.num_sentences,) +
+                                 self._get_sentence_shape(self.num_sentence_words))
```
matt-gardner: No need to have the argument to `self._get_sentence_shape()`, because the default is to use `self.num_sentence_words`.  Also, the shapes in the comments should use names that match the variables.

deep_qa/models/sentence_selection/siamese_sentence_selector.py
```diff
@@ -0,0 +1,123 @@
...
+        super(SiameseSentenceSelector, self)._set_max_lengths(max_lengths)
+        self.num_question_words = max_lengths['num_question_words']
+        self.num_sentences = max_lengths['num_sentences']
+
+    @overrides
+    def _set_max_lengths_from_model(self):
+        self.set_text_lengths_from_model_input(self.model.get_input_shape_at(0)[1][2:])
+        self.num_question_words = self.model.get_input_shape_at(0)[0][1]
+        self.num_sentences = self.model.get_input_shape_at(0)[1][1]
+        self.num_sentence_words = self.model.get_input_shape_at(0)[1][2]
```
matt-gardner: This gets handled with the call to `self.set_text_lengths_from_model_input` - no need to have this line here.

deep_qa/models/sentence_selection/siamese_sentence_selector.py
nelson-liu: oops, keep forgetting that `TextTrainer` handles some of these lengths. thanks

nelson-liu:
merging this now, since it finally passed ci...

PR #277, opened 2017-03-27
Stack seq2seq encoders to give sentence selector more depth

nelson-liu:
Matt suggested to stack some seq2seq encoders to give the model more depth; this does so in a similar manner to the BiDAF code.

There's a funky error going on right now wrt masking the 4D input to the seq2seq encoder, and I can't quite figure it out. It complains that the mask is 2d, and the input itself is 4d...perhaps we need a `EncoderWrapper` specialized for `seq2seq_encoders`?

### DIFF
--- a/deep_qa/models/sentence_selection/siamese_sentence_selector.py
+++ b/deep_qa/models/sentence_selection/siamese_sentence_selector.py
@@ -24,13 +24,16 @@ class SiameseSentenceSelector(TextTrainer):
 
     Parameters
     ----------
-    num_seq2seq_layers : int, optional (default: ``2``)
+    num_hidden_seq2seq_layers : int, optional (default: ``2``)
         We use a few stacked biLSTMs (or similar), to give the model some
         depth.  This parameter controls how many deep layers we should use.
 
+    share_hidden_seq2seq_layers : bool, optional (default: ``False``)
+        Whether or not to encode the sentences and the question with the same
+        hidden seq2seq layers, or have different ones for each.
     """
     def __init__(self, params: Dict[str, Any]):
-        self.num_seq2seq_layers = params.pop('num_hidden_seq2seq_layers', 2)
+        self.num_hidden_seq2seq_layers = params.pop('num_hidden_seq2seq_layers', 2)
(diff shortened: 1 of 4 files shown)


deep_qa/models/sentence_selection/siamese_sentence_selector.py
```diff
@@ -56,18 +65,36 @@ def _build_model(self):
         # shape: (batch size, num_sentences, num_sentence_words, embedding size)
         sentences_embedding = self._embed_input(sentences_input)
 
-        # We encode the question embeddings with some encoder.
+        # We encode the question embedding with some more seq2seq layers
+        modeled_question = question_embedding
+        for i in range(self.num_seq2seq_layers):
+            hidden_layer = self._get_seq2seq_encoder(name="question_seq2seq_{}".format(i),
```
matt-gardner: Note that with these names, it is no longer possible to make this an actual siamese network - there is no way to make these the same encoder with what you've done here.  And I'm not sure how to make it so you can do that just with parameters in the json file...  Probably you'd have to just pass a flag to the model class that says whether to use common names for these hidden layers.

nelson-liu: do you think it's preferable to share weights / use the same encoder, or keep them separate? I agree that passing a flag that indicates whether to share weights would work, but it's hard to make it very fine-grained without making the config overly verbose / cumbersome

matt-gardner: I don't have a good intuition for whether it'd be better to share them or not.  Sharing them means fewer parameters, allowing them to be different lets you encode the query differently than you encode the corpus, which might be beneficial.  Allowing options is probably fine.  And I don't think this particular option would be that bad - just have a flag that called something like `share_hidden_layers`, and if it's true, use `seq2seq_{}` as the name, instead of `question_seq2seq_{}` and `passage_seq2seq_{}`.

matt-gardner:
Hmm, I guess pyhocon is ok with python casing on booleans.  I'd probably stick with valid json where possible, though.

nelson-liu:
I'm happy with this --- want to merge, @matt-gardner ?

PR #281, opened 2017-03-30
IDF in BOW background search

pdasigi:


deep_qa/contrib/background_search/bow_lsh.py
```diff
@@ -99,20 +110,36 @@ def get_word_vector(self, word, random_for_unk=False):
                 self.embeddings[word] = vector
             else:
                 vector = numpy.zeros((self.embedding_dim,))
-            return vector
 
-    def encode_sentence(self, sentence: str, for_background=False):
-        words = [str(w.lower_) for w in self.en_nlp.tokenizer(sentence)]
+        if self.use_idf:
+            # Note that this line causes us to return zero vectors for words questions file that are in the
```
matt-gardner: You're missing some words here: "words questions file".  Not sure what this is supposed to mean.  But shouldn't you use 1 as the default, not 0?  The point is to downweight frequent words - if a word showed up in every document, you'll get `log(N/N) = 0`.  When you give 0 here as the weight for unseen words, you're treating them like they showed up in every document.  Instead, you should give them a weight of `log(N/1)`.  This would be pretty easy to do, by adding an `@UNSEEN@` key to `self.idf_values` when you compute it, and doing `if word in self.idf_values else self.idf_values['@UNSEEN@']`.

pdasigi: Good catch. You're right. Thanks!

PR #282, opened 2017-03-30
SNLI model

matt-gardner:
I'm currently training a model with this, and it's taking ~3.5 hours per epoch.  @matt-peters, you don't need to look at this too closely if you don't want to, but this is the model I'd suggest you modify, and there's a json file here to run the code with.  Loading the data took <2G of RAM with the settings in the json file.  The data is on the S2 EFS drive.


matt-peters:
LGTM, thanks!

PR #283, opened 2017-03-31
Allow use of a data generator for training

matt-peters:
This commit allows `Trainer` subclasses to use a data generator for training and evaluation instead of pre-computed `X` and `y` arrays.

Behavior is controlled with the keys `use_data_generator` and `steps_per_epoch` in the params.  Default behavior is unchanged (don't use a generator).

### DIFF
--- a/deep_qa/training/trainer.py
+++ b/deep_qa/training/trainer.py
@@ -114,6 +114,10 @@ class Trainer:
             pretrainer = concrete_pretrainers[pretrainer_type](self, pretrainer_param)
             self.pretrainers.append(pretrainer)
 
+        # Use a data generator?
+        self.use_data_generator = params.pop('use_data_generator', False)
+        self.steps_per_epoch = params.pop('steps_per_epoch', None)
+
         # We've now processed all of the parameters, and we're the base class, so there should not
         # be anything left.
         if len(params.keys()) != 0:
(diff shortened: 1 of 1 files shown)


matt-gardner:
Have you actually run this?  It'd be nice to have a test in [here](https://github.com/allenai/deep_qa/blob/master/tests/training/text_trainer_test.py) with a model that actually uses a generator, just to make sure this works.

deep_qa/training/trainer.py
```diff
@@ -281,9 +285,13 @@ def train(self):
         train_data, val_data, test_data = self.prepare_data(self.train_files, self.max_training_instances,
                                                             self.validation_files,
                                                             self.test_files)
-        self.training_dataset, self.train_input, self.train_labels = train_data
-        self.validation_dataset, self.validation_input, self.validation_labels = val_data
-        self.test_dataset, self.test_input, self.test_labels = test_data
+        if not self.use_data_generator:
```
nelson-liu: randomly decided to take a look at this (i hope you don't mind...), but I don't quite see the benefit of a data generator here. You still have to compute the `train_data, val_data, test_data` from `self.prepare_data`, right? So I don't quite see how this is any different than just popping things off of a list, i.e. I think that you'd still have to precompute `X` and `y` with this approach, which makes me question the value of a generator.

I've been using a modified version of the dataset code in this repo to structure some personal work, and I (think, haven't actually tested it yet) managed to get a generator working by replacing the equivalent of `self._prepare_data` with a function that returns a generator, vs lists. This involved making the `pad_instances` function in Datasets return a generator, so you essentially pad "on-demand", but I haven't figured out how to get it working here...

matt-peters:
Thanks for the comments.  I have run this and it is working for my purpose.  To get it to work I had to subclass `Trainer` to override `self.prepare_data` to return generators.  This was beneficial in my case for a couple reasons:

* it's necessary to me to change the indexing logic in the datasets to be consistent with another code base
* I'd like to support batches that are dynamically padded for each batch for efficiency reasons

I'll look at adding a test but since it needs a working subclass of Trainer to run it's not super easy to do.

nelson-liu:
Ah, yeah I was just pointing out that this wouldn't work out of the box with one of our texttrainers; thanks for the additional context

matt-gardner:
I can worry about adding a test for this - what I'm thinking of is pretty simple, just overriding the `TrueFalseModel` in the test file itself, just to make sure this code is tested.  I'll do that soon, and push it to the PR before merging (but probably not today).

matt-gardner:
Closing this, as it's superseded by #295.  Thanks for this @matt-peters, it was helpful in figuring out the right way to integrate this functionality into DeepQA.

PR #289, opened 2017-04-09
Improve docs, clean up the base API a bit

matt-gardner:
I talked with Matt Peters about what made DeepQA hard to use, and one of his major suggestions was making the intended API more clear.  This PR does that.  There are three main things in this:

1. I updated the docstrings in `Trainer` and `TextTrainer`, reordered methods in the code to group them by public / abstract / etc., and renamed private methods to actually be private.  This is to make it clear exactly what you're supposed to get out of `TextTrainer`.
2. I removed the pre-training functionality.  I think just loading a model and re-using it is a cleaner way to do the same thing, and getting loading and everything to work right the way we were trying to do things was just messy and didn't ever work well.  Removing the notion of pre-training let me simplify a few other parts of the code.
3. I removed the `api_doc/` section of our documentation, which seemed to be largely just a duplication of what was in the rest of the docs.  If you really want what used to be here, just browse github - it's basically the same thing.  Removing it gives us one less place to keep everything in sync.


matt-gardner:
Just going to merge this, as it's a large, mostly documentation-level change, and so as not to distract people from working on EMNLP papers.  If anyone has comments on this, feel free to post them on here, and I can address them in a future PR.

PR #290, opened 2017-04-11
Data generator prep

matt-gardner:
I'm doing some code cleanup in preparation for switching to data generators, instead of loading all of the data up front.  This will include dynamic padding (which, along with reduced memory consumption, is the main benefit here), but that will be a future PR.

This PR is mainly just some renaming and improving docstrings around loading data.


matt-gardner:
Note that it might be easier to look at this by commit, instead of all at once, because there's a sed command that actually did most of the line changes.

deep_qa/training/text_trainer.py
```diff
@@ -139,23 +139,17 @@ def __init__(self, params: Dict[str, Any]):
...
             logger.info("Fitting data indexer word dictionary.")
             self.data_indexer.fit_word_dictionary(dataset)
         logger.info("Indexing dataset")
         indexed_dataset = dataset.to_indexed_dataset(self.data_indexer)
-        max_lengths = self._get_max_lengths()
-        logger.info("Padding dataset to lengths %s", str(max_lengths))
-        indexed_dataset.pad_instances(max_lengths)
-        if for_train:
-            self._set_max_lengths(indexed_dataset.max_lengths())
+        padding_lengths = self._get_padding_lengths()
```
nelson-liu: i like this change a lot

PR #294, opened 2017-04-13
Added CollapseToBatch and ExpandFromBatch, to handle dynamic padding

matt-gardner:
I'll let the docstrings give more explanation for what's going on here.


deep_qa/layers/backend/collapse_to_batch.py
```diff
@@ -0,0 +1,65 @@
...
+import tensorflow as tf
+
+from ..masked_layer import MaskedLayer
+
+
+class CollapseToBatch(MaskedLayer):
+    """
+    Reshapes a higher order tensor, taking the first ``num_to_collapse`` dimensions and folding
+    them into the batch dimension.  For example, a tensor of shape (2, 4, 5, 3), collapsed with
+    ``num_to_collapse = 2``, would become a tensor of shape (40, 3).  We perform identical
```
pdasigi: You mean `num_to_collapse = 3`?

deep_qa/layers/backend/collapse_to_batch.py
```diff
@@ -0,0 +1,65 @@
...
+    -----
+    This layer requires tensorflow, because of the way we get runtime shapes.
+    """
+    def __init__(self, num_to_collapse: int, **kwargs):
+        self.num_to_collapse = num_to_collapse
+        super(CollapseToBatch, self).__init__(**kwargs)
+
+    @overrides
+    def call(self, inputs, mask=None):
+        new_shape = tf.concat([[-1], tf.shape(inputs)[1 + self.num_to_collapse:]], 0)
```
pdasigi: Can you add a comment describing what this is doing?

deep_qa/layers/backend/collapse_to_batch.py
```diff
@@ -0,0 +1,65 @@
...
+        return tf.reshape(inputs, new_shape)
+
+    @overrides
+    def compute_mask(self, inputs, mask=None):
+        # pylint: disable=unused-argument
+        if mask is None:
+            return None
+        new_shape = tf.concat([[-1], tf.shape(mask)[1 + self.num_to_collapse:]], 0)
```
pdasigi: Comment here as well.

deep_qa/layers/backend/expand_from_batch.py
```diff
@@ -0,0 +1,71 @@
...
+
+    For example, say you had an original tensor of shape (2, 4, 5, 3), then collapsed it with
+    ``CollapseToBatch(2)(tensor)`` to get a tensor with shape (40, 3).  You can then call
+    ``ExpandFromBatch(2)(collapsed, tensor)`` with the result to expand the first two dimensions
+    out of the batch again (presumably after you've done some computation when it was collapsed).
+
+    Inputs:
+        - a tensor that has been collapsed with ``CollapseToBatch(num_to_expand)``.
+        - the original tensor that was used as input to ``CollapseToBatch`` (or one with identical
+          shape in the collapsed dimensions).  We will use this input only to get its shape.
```
pdasigi: Why not just pass the shape?

deep_qa/layers/backend/collapse_to_batch.py
pdasigi: Oh, you mean excluding the original batch size dimension? May be you should mention that.

matt-gardner: Yeah, I'll update the docstring.

deep_qa/layers/backend/expand_from_batch.py
matt-gardner: Because the shape will have `Nones` in it - that's the whole point of using this instead of `TimeDistributed`.  We need a way to get runtime shape, which we do by calling `tf.shape` on the input tensor.  We don't have a good way to get runtime shape in `_build_model`, where we're dealing with Keras layers, so we just take the original tensor to get the shape, so we can expand back the same way.  This should be a bit more clear once I add some comments to `CollapseToBatch`.

matt-gardner:
I addressed your comments @pdasigi; assuming this is good to merge.  I'll merge it soon unless I hear from you.

pdasigi:
LGTM, for the record.

PR #295, opened 2017-04-13
Data generators and dynamic padding

matt-gardner:
This PR allows for what I'll call dynamic padding.  This means that padding only happens at a batch level.  If you sort instances by size and then sample slices from this list, you should be able to dramatically decrease runtime, because the amount of padding you actually need is minimized.

So far I've implemented everything but the sorting, for the simplest model case.  Individual models will have to do a little bit of work, probably, to support this for particular padding dimensions, but the changes should be minimal in most cases.


deep_qa/training/text_trainer.py
```diff
@@ -62,6 +64,22 @@ class TextTrainer(Trainer):
...
         Dropout parameter to apply to the embedding layer
+    use_data_generator: bool, optional (default=False)
+        In :func:`~TextTrainer.create_data_arrays`, should we return one big, padded data array, or
+        should we return a data generator?  Using a data generator allows for dynamic
+        (batch-specific) padding, which can dramatically reduce runtime, but your model must be
+        written correctly to take advantage of that.  See :func:`~TextTrainer._set_padding_lengths`
+        and :func:`~TextTrainer._get_padding_lengths` for more details on that.
+    use_dynamic_padding: bool, optional (default=False)
+        If ``True``, we will set padding lengths based on the data `per batch`, instead of on the
+        whole dataset.  This only works ``use_data_generator`` is ``True``, if your model is
```
DeNeutoy: Missing "if".

deep_qa/training/trainer.py
```diff
@@ -56,6 +57,18 @@ class Trainer:
         Upper limit on the number of validation instances, analogous to ``max_training_instances``.
     max_test_instances: int, optional (default=None)
         Upper limit on the number of test instances, analogous to ``max_training_instances``.
+    train_steps_per_epoch: int, optional (default=None)
```
DeNeutoy: I don't know if the docs require this level of verbosity, but this docstring says the same thing for all three of these arguments.

deep_qa/data/tokenizers/word_and_character_tokenizer.py
```diff
@@ -90,13 +91,18 @@ def embed_input(self,
...
-        for _ in range(3, K.ndim(characters)):
-            word_encoder = EncoderWrapper(word_encoder, name="timedist_" + word_encoder.name)
-        word_encoding = word_encoder(character_embedding)
+        # that you have 0s in the word-level mask, so `Concatenate` further below will do the right
+        # thing.
+
+        # We want to pass a tensor with shape `(batch_size, word_length, embedding_dim)` to the
+        # word encoder.  However, we may not know `num_words` (which we're collapsing) or even
+        # `word_length` at runtime, which messes with Keras' TimeDistributed.  In order to handle
+        # this correctly, we'll use CollapseToBatch and ExpandFromBatch instead of TimeDistributed.
```
DeNeutoy: It was not obvious to me what these functions did - I think it would be better if they were called `RuntimeReshape` or something like that, maybe?

deep_qa/data/tokenizers/word_and_character_tokenizer.py
```diff
@@ -90,13 +91,18 @@ def embed_input(self,
...
-        word_encoder = EncoderWrapper(
-                text_trainer._get_encoder(name="word", fallback_behavior="use default params"))
-        # We might need to TimeDistribute this again, if our input has ndim higher than 3.
-        for _ in range(3, K.ndim(characters)):
-            word_encoder = EncoderWrapper(word_encoder, name="timedist_" + word_encoder.name)
-        word_encoding = word_encoder(character_embedding)
+        # that you have 0s in the word-level mask, so `Concatenate` further below will do the right
+        # thing.
+
+        # We want to pass a tensor with shape `(batch_size, word_length, embedding_dim)` to the
```
DeNeutoy: This would be clearer if you start with what the shape actually is and say what we want it to be, rather than this way around.

deep_qa/data/tokenizers/word_and_character_tokenizer.py
matt-gardner: You mean `CollapseToBatch` and `ExpandFromBatch`?  What if I called them `CollapseDimsToBatch` and `ExpandDimsFromBatch`?  `RuntimeReshapeToBatch` and `RuntimeReshapeFromBatch` also have some appeal, as it makes more it obvious why you would use this instead of `TimeDistributed`...

deep_qa/training/trainer.py
matt-gardner: I guess they'll be right next to each other always... I could just refer to the first one.

matt-gardner:
Note that this handles more than just what `dynamic_rnn` does.  In this particular setting, we're not doing RNNs at all - we're distributing a character-level CNN encoder across an unknown number of words.  This isn't something you would do with `dynamic_rnn`.

I agree that we need to be careful with this, though, which is why there's so much documentation about how you should probably use `TimeDistributed` if it works for you.  This limits the kinds of reshaping you can easily do in your models.  I should probably use the `EncoderWrapper` by default in the word and character tokenizer if `use_dynamic_padding` is `False`, actually, otherwise you just won't be able to build certain kinds of models that use character-level encoders.

deep_qa/data/tokenizers/word_and_character_tokenizer.py
DeNeutoy: I originally thought collapse dims would remove the stated dims, rather than reshape the batch to incorporate them - I appreciate that this initial line of thought might have been a little odd, so feel free to leave it if you are happy.

DeNeutoy:
Yeah sure, you're right. Feel free to merge.

deep_qa/data/tokenizers/word_and_character_tokenizer.py
matt-gardner: I improved the comment, so hopefully it's more immediately understandable now.

deep_qa/data/tokenizers/word_and_character_tokenizer.py
matt-gardner: Yeah, good idea.  Done.

deep_qa/training/text_trainer.py
matt-gardner: Done.

PR #297, opened 2017-04-14
Moved instances into task-specific modules, implemented sorting by padding

matt-gardner:
This PR looks bigger than it really is.  It's just moving the instances around, organizing them by task (which required changing a bunch of imports all over the code), renaming `get_lengths` in `IndexedInstance` to `get_padding_lengths` (which required minor fixes in a bunch of lines in the code), some documentation improvements, and adding one small method to `IndexedDataset`.


deep_qa/data/dataset.py
```diff
@@ -195,9 +165,22 @@ class IndexedDataset(Dataset):
     def __init__(self, instances: List[IndexedInstance]):
         super(IndexedDataset, self).__init__(instances)
 
+    def sort_by_padding(self, sorting_keys: List[str]):
+        """
+        Sorts the ``Instances`` in this ``Dataset`` by their padding lengths, using the keys in
```
DeNeutoy: Can you add: "in the order that the sorting keys are passed" to the docs here, or a comment where you sort with the key as a list?

deep_qa/training/text_trainer.py
```diff
@@ -485,6 +484,20 @@ def _set_padding_lengths_from_model(self):
...
+        we minimize the amount of padding necessary per batch.  This variable sets what exactly
+        gets sorted by.  We'll call
+        :func:`~deep_qa.data.instances.instance.IndexedInstance.get_padding_lengths()` on each
+        instance, pull out these keys, and sort by them in the order specified.  You'll want to
+        override this in your model class if you have more complex models.
+
+        The default implementation is to sort first by ``num_sentence_words``, then by
+        ``num_word_characters``.
+        """
+        return ['num_sentence_words', 'num_word_characters']
```
DeNeutoy: I think this should be just `['num_sentence_words']` - in `instance.get_padding_lengths()` we assume that these keys exist in the length dictionary, but `num_word_characters` only exists if we are also using character embeddings (I think).

DeNeutoy: Having this here is also tying the type of TextTrainer to the type of Instance we are using. Is this a good thing? In order to train a model dynamically padded with a different instance padding type, you would have to subclass TextTrainer. Wouldn't it be better to have this sort of thing as additional args passed to a `Dataset`? 

The `Dataset` classes have a very clean API, but this is becoming at the expense of `TextTrainer`. Perhaps it would be better to push some of the complexity over to `Dataset`, rather than having this here. This is more of a high level thing to possibly fix, interested in your opinion.

matt-gardner: On your first comment: yeah, you're totally right about that.  Good catch.  I'll fix it.

On your second comment: that's kind of the whole point of the TextTrainer subclasses.  Every model is already a TextTrainer subclass, and you have to specify what instance type you're using when you specify `_build_model`.  I don't think there's a way around this; the coupling is required, because the instance type defines your model inputs and outputs.

The alternative would be to have the model come in by composition, not inheritance (e.g., we have a parallel class called 'Model' or something that implements all of the abstract methods that are currently in `TextTrainer`, and `TextTrainer` instantiates one of those, instead of subclassing `TextTrainer`).  The trouble with that is that the TextTrainer is supposed to provide a useful API to the model itself for when you're building the model, and so there would be a ton of back-and-forth between these two separate objects (i.e., you'd have to pass the TextTrainer as an argument to the `Model` constructor, or at least to `_build_model`), which is a pretty strong argument in favor of inheritance instead of composition.

deep_qa/data/dataset.py
matt-gardner: Done.

matt-gardner:
Ok, finally got the merged with other recent PRs, and tests passing.  @DeNeutoy, is this good to merge?

PR #301, opened 2017-04-17
add parameter logging

### LINKED ISSUE #300: Better parameter logging
It'd be nice if every time a parameter was read, there was a logging statement.  This is for reproducibility across code changes, particularly with default parameters.  If the default changes in a version of code, the saved parameter file associated with a model won't give the same result.  If instead we print out the actual value used for each parameter, default or not, we'd have a complete picture of what parameters were used.

I'd probably do this by adding a method to `common.params`, something like `get_with_default`, that logs the value that actually gets returned, with a common prefix.  Then you can grep the output for these statements, and rebuild a json file that has all of the parameters, including default values.

DeNeutoy:
fixes #300


deep_qa/common/params.py
```diff
@@ -1,9 +1,28 @@
...
 
 from .checks import ConfigurationError
 
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+PARAMETER = 60
+logging.addLevelName(PARAMETER, "PARAM")
+
+
+def param(self, message, *args, **kws):
```
matt-gardner: Can you make this a private method?  `def __param()`

deep_qa/common/params.py
```diff
@@ -12,6 +31,28 @@ def _get_choice_error_message(value: Any, choices: List[Any], name: str=None) ->
         return '%s not in acceptable choices: %s' % (value, str(choices))
 
 
+def pop_with_logging(params: Dict[str, Any], key: str):
```
matt-gardner: We actually need more detail than this.  We need a `name`, here, which tells you which class, or (even better) which path gave us this parameter.  Ideally, you could handle this like namespaces are handled in tensorflow, just adding a new namespace when you start reading parameters for an object you're creating, so that things are automatically nested and you can basically read off the entire nested json structure.  Does this make sense?

deep_qa/common/params.py
```diff
@@ -12,6 +31,28 @@ def _get_choice_error_message(value: Any, choices: List[Any], name: str=None) ->
...
+    Performs the functionality associated with dict.pop(key) but with parameter
+    logging. This is required because pop_with_default may receive a default value
+    of None, which means we can't check for it not being passed.
+    """
+    value = params.pop(key)
+    logger.param(key + " : " + str(value))
+    return value
+
+
+def pop_with_default(params: Dict[str, Any], key: str, default: Any):
```
matt-gardner: What if we just merge this method and the above, making `default: Any=None`?  Then you just call it `get_param`, or `pop_param`, or something, and all parameter reads will just use the same function.

DeNeutoy:
@matt-gardner This now works except I am having some issues with model serialisation - I think you might know more about that than me. Basically, Keras can't serialise the `Param` objects which are arguments to layers....

deep_qa/common/params.py
```diff
@@ -21,74 +23,119 @@ def param(self, message, *args, **kws):
...
+        return self.__check_is_dict(key, value)
+
+    @overrides
+    def get(self, key: str, default: Any=DEFAULT) -> Union["Params", Any]:
+        """
+        Performs the functionality associated with dict.pop(key) but with parameter
+        logging. This is required because pop_with_default may receive a default value
+        of None, which means we can't check for it not being passed.
+        """
+        print(default)
```
DeNeutoy: Ah sorry this should implement the `get` method of a dict, I just copied it over from `pop` and forgot to change it, thanks for spotting

deep_qa/common/params.py
```diff
@@ -44,34 +44,32 @@ def __init__(self, params: Dict[str, Any], history: str=""):
...
             value = self.params.pop(key)
         else:
             value = self.params.pop(key, default)
-        logger.param(key + " : " + str(value))
+        logger.param(self.history + " : " + key + " : " + str(value))
         return self.__check_is_dict(key, value)
 
     @overrides
-    def get(self, key: str, default: Any=DEFAULT) -> Union["Params", Any]:
+    def get(self, key: str, default: Any=DEFAULT):
```
matt-gardner: Do we have to implement this method at all?  If it's here, people might be tempted to use it, and that messes with parameter validation.  I don't think we actually use this anywhere.

deep_qa/common/params.py
```diff
@@ -83,42 +81,41 @@ def pop_choice(self, key: str, choices: List[Any], name: str=None) -> Union["Par
...
+                raise ConfigurationError(_get_choice_error_message(default, choices, self.history))
 
             return self.__check_is_dict(key, default)
 
+    def as_dict(self):
+        """
+        Sometimes we need to just represent the parameters as a dict, for instance when we pass
+        them to a Keras layer(so that they can be serialised).
+        """
+        return self.params
```
matt-gardner: Can you do a recursive `logger.param()` on everything in here, before returning it, as well as an initial statement like `logger.param("Converting param object to dictionary; logging of default values won't work with the result")`?

DeNeutoy:
@matt-gardner this now has all tests passing - can you just check you are happy with what I have done for the `recurrence` classes - in order to not use dicts and pop off arguments but retain the same interface, I've had to allow passing arguments which aren't used. Don't know what you think about that, but I think changing it will be ugly because we'd then have to have conditions in the `_get_recurrence_mode()` method of the `MemoryNetwork`.

deep_qa/common/params.py
```diff
@@ -21,74 +23,119 @@ def param(self, message, *args, **kws):
...
+        Note that this _pops_ the key from params, modifying the dictionary, consistent with how
+        parameters are processed in this codebase.
+        """
+
+        value = self.pop(key)
+        if value not in choices:
+            raise ConfigurationError(self._get_choice_error_message(value, choices, name))
+        return self.__check_is_dict(key, value)
+
+    def pop_choice_with_default(self,
```
matt-gardner: Am I missing something about why this might be hard, or have you just not gotten to this one yet?  It'd make the API cleaner to just have one function to call for popping a choice.

deep_qa/layers/adaptive_recurrence.py
```diff
@@ -31,7 +31,8 @@ class AdaptiveRecurrence:
     can create one copy of the weights and although the number of iterations of the memory step is
     non-deterministic, the computational graph can be statically defined - we just loop over a part of it.
     '''
-    def __init__(self, memory_network, params: "Params"):
+    # pylint: disable=dangerous-default-value, unused-argument
```
matt-gardner: I think you don't need either of these pylint disablings - the first one is important, and we shouldn't disable it, and the second one shouldn't apply (unless it's for kwargs?).  I would do this instead like this:

```python
def __init__(self, memory_network, adaptive_step_params: Dict[str, Any]=None):
    if adaptive_step_params is None:
        adaptive_step_params = {}
    self.adaptive_step_params = adaptive_step_params
```

You could also consider just calling the argument `adaptive_step` instead of `adaptive_step_params`, letting the type convey the information that it's a set of parameters.

deep_qa/layers/adaptive_recurrence.py
```diff
@@ -65,7 +66,7 @@ class AdaptiveStep(MaskedLayer):
     probabilities, states and outputs from a timestep t's contribution if they have already reached
     1 - epsilon at a timestep s < t.
     '''
-    def __init__(self, memory_network, params: "Params",
+    def __init__(self, memory_network, params,
```
matt-gardner: Instead of having params here, you should instead have `epsilon`, `max_computation`, etc., as named arguments to `__init__`, just like `initialization` and `name` are.  Though, `name` doesn't really matter here, because it gets picked up by the super class already.  If you notice the `TODO` down in `get_config`, it's because of exactly this issue.  The keys there need to match the named arguments to `__init__`, in order to conform to the Keras layer API.

deep_qa/layers/adaptive_recurrence.py
matt-gardner: Just saw your comment on the PR, too - if you have to have `**kwargs` here to make the API work out with different layer classes, that's fine, and you can keep the `unused-argument` pylint thing.

deep_qa/layers/attention/attention.py
```diff
@@ -32,16 +33,19 @@ class Attention(MaskedLayer):
 
     Parameters
     ----------
-    similarity_function_params: "Params", optional (default={})
+    similarity_function_params: Dict, optional (default={})
```
matt-gardner: `Dict[str, Any]`

deep_qa/layers/attention/attention.py
```diff
@@ -32,16 +33,19 @@ class Attention(MaskedLayer):
...
         These parameters get passed to a similarity function (see
         :mod:`deep_qa.tensors.similarity_functions` for more info on what's acceptable).  The
         default similarity function with no parameters is a simple dot product.
     """
-    def __init__(self, similarity_function: "Params"=Params({}), **kwargs):
+    # pylint: disable=dangerous-default-value
+    def __init__(self, similarity_function: Dict={}, **kwargs):
```
matt-gardner: Yeah, this should just go back to how it was, without the pylint warning, and with a default value of `None`.

deep_qa/layers/attention/attention.py
```diff
@@ -32,16 +33,19 @@ class Attention(MaskedLayer):
...
         super(Attention, self).__init__(**kwargs)
         self.similarity_function_params = deepcopy(similarity_function)
-        sim_function_choice = similarity_function.pop_choice_with_default('type',
-                                                                          list(similarity_functions.keys()))
+        sim_function_choice = pop_choice_with_default(similarity_function,
+                                                      'type', list(similarity_functions.keys()))
+        # TODO(mark): The default value here is mutable so this will introduce naming bugs
```
matt-gardner: Once it goes back to how it was, I think this TODO is no longer necessary.

deep_qa/layers/attention/matrix_attention.py
```diff
@@ -1,10 +1,11 @@
 from copy import deepcopy
+from typing import Dict
 
 from keras import backend as K
 from overrides import overrides
 
 from ..masked_layer import MaskedLayer
-from ...common.params import Params
+from ...common.params import pop_choice_with_default
```
matt-gardner: You can probably change `pop_choice_with_default` to just `pop_choice` here, too, using the same `DEFAULT = object()` line.

deep_qa/layers/attention/matrix_attention.py
```diff
@@ -35,17 +36,18 @@ class MatrixAttention(MaskedLayer):
 
     Parameters
     ----------
-    similarity_function_params: "Params", default={}
+    similarity_function_params: Params, default={}
```
matt-gardner: `s/Params/Dict[str, Any]/`

deep_qa/layers/attention/matrix_attention.py
```diff
@@ -35,17 +36,18 @@ class MatrixAttention(MaskedLayer):
...
         These parameters get passed to a similarity function (see
         :mod:`deep_qa.tensors.similarity_functions` for more info on what's acceptable).  The
         default similarity function with no parameters is a simple dot product.
     '''
-    def __init__(self, similarity_function: "Params"=Params({}), **kwargs):
+    # pylint: disable=dangerous-default-value
+    def __init__(self, similarity_function: Dict={}, **kwargs):
```
matt-gardner: Same here, with putting the default value back to how it was.

deep_qa/layers/entailment_models/encoded_sentence.py
```diff
@@ -26,10 +25,10 @@ class TrueFalseEntailmentModel:
     a single vector, then pass that vector through an MLP.  How we actually merge the three inputs
     is specified by one of the combiners below.
     """
-    def __init__(self, params: "Params"):
-        self.num_hidden_layers = params.pop('num_hidden_layers', 1)
-        self.hidden_layer_width = params.pop('hidden_layer_width', 50)
-        self.hidden_layer_activation = params.pop('hidden_layer_activation', 'relu')
+    def __init__(self, num_hidden_layers=1, hidden_layer_width=50, hidden_layer_activation="relu"):
```
matt-gardner: Can you add types to these?

deep_qa/layers/entailment_models/encoded_sentence.py
```diff
@@ -68,11 +67,11 @@ class QuestionAnswerEntailmentModel:
     into the same dimension as the answer encoding, does a dot product between the combined input
     encoding and the answer encoding, and does a final softmax over those similar scores.
     """
-    def __init__(self, params: "Params"):
-        self.num_hidden_layers = params.pop('num_hidden_layers', 1)
-        self.hidden_layer_width = params.pop('hidden_layer_width', 50)
-        self.hidden_layer_activation = params.pop('hidden_layer_activation', 'relu')
-        self.answer_dim = params.pop('answer_dim')
+    def __init__(self, answer_dim, num_hidden_layers=1, hidden_layer_width=50, hidden_layer_activation="relu"):
```
matt-gardner: Types here too would be nice.

deep_qa/layers/entailment_models/encoded_sentence.py
```diff
@@ -118,10 +117,10 @@ class MultipleChoiceEntailmentModel:
     a single vector, then pass that vector through an MLP, once for each of the multiple choices,
     then have a final softmax over answer options.
     """
-    def __init__(self, params: "Params"):
-        self.num_hidden_layers = params.pop('num_hidden_layers', 1)
-        self.hidden_layer_width = params.pop('hidden_layer_width', 50)
-        self.hidden_layer_activation = params.pop('hidden_layer_activation', 'relu')
+    def __init__(self, num_hidden_layers=1, hidden_layer_width=50, hidden_layer_activation="relu"):
```
matt-gardner: And here...

deep_qa/layers/recurrence_modes.py
```diff
@@ -10,9 +10,9 @@ class FixedRecurrence:
     generated by the knowledge_selector and knowledge_combiner layers (the simplest
     case being a weighted sum).
     '''
-    def __init__(self, memory_network, params: "Params"):
+    def __init__(self, memory_network, num_memory_layers):
```
matt-gardner: Type and default value here.

deep_qa/layers/tuple_matchers/embedded_tuple_matcher.py
```diff
@@ -17,15 +18,15 @@ class EmbeddedTupleMatcher:
         A reference to the TextTrainer object this TupleMatcher belongs to, so we can call
         ``embed_input`` with it.
 
-    tuple_matcher_params: "Params", default={}
+    tuple_matcher_params: Dict, default={}
```
matt-gardner: `Dict[str, Any]`, here and below.

deep_qa/layers/tuple_matchers/encoded_tuple_matcher.py
```diff
@@ -19,17 +20,18 @@ class EncodedTupleMatcher:
         A reference to the TextTrainer object this TupleMatcher belongs to, so we can call
         ``embed_input`` and ``get_encoder`` with it.
 
-    tuple_matcher_params: "Params", default={}
+    tuple_matcher_params: Params, default={}
```
matt-gardner: `Dict[str, Any]`, and default value needs to be fixed below.

deep_qa/layers/tuple_matchers/slot_similarity_tuple_matcher.py
```diff
@@ -35,7 +36,7 @@ class SlotSimilarityTupleMatcher(MaskedLayer):
 
     Parameters
     ----------
-    - similarity_function_params: "Params", default={}
+    - similarity_function_params: Dict, default={}
```
matt-gardner: `Dict[str, Any]`

deep_qa/layers/tuple_matchers/threshold_tuple_matcher.py
```diff
@@ -52,7 +53,7 @@ class ThresholdTupleMatcher(MaskedLayer):
 
     Parameters
     ----------
-    similarity_function_params: "Params", default={}
+    similarity_function_params: Dict, default={}
```
matt-gardner: `Dict[str, Any]`

deep_qa/models/memory_networks/differentiable_search.py
```diff
@@ -3,14 +3,15 @@
 import pickle
 
 from itertools import zip_longest
-from typing import Dict, List  # pylint disable: unused-import
+from typing import Dict, List  # pylint: disable=unused-import
```
matt-gardner: If this isn't actually necessary, just remove the import.

deep_qa/layers/attention/matrix_attention.py
DeNeutoy: This is slightly different though - if we do that, to use the first value of `choices` as the default parameters you would have to explicitly pass `None`. Is that what you want?

deep_qa/common/params.py
DeNeutoy: Yes - we do use it in several places and __getitem__ is only used for the `dict[key]` syntax and `MutableMapping` doesn't implement it, so without this you can't use `.get` on `Params`. I was trying to make using `Params` as near to using a `dict` as possible.

deep_qa/layers/attention/matrix_attention.py
matt-gardner: Yeah, that's a good point.  One suggestion:

```python
def pop_choice(params, key, choices, default: Any=None, allow_default: bool=False):
  # in docstring, say that this pops the key, ensures that the key is in `choices`,
  # and uses a default value if the key is not present in the dictionary (and you want
  # that behavior).  If you pass a non-None default, `allow_default` is implied, and
  # thus not necessary.  If `default` is `None` and `allow_default` is `True`, we'll use
  # choices[0] as the default.
  if key not in params:
    if default is not None:
      if default not in choices:
        raise ConfigurationError
      return default
    elif allow_default:
      return choices[0]
    else:
      raise ConfigurationError("required key not specified; or some similar message")
  value = params.pop(key)
  if value not in choices:
    raise ConfigurationError()
  return value

File: deep_qa/common/params.py (hunk 5 above)
Comment (matt-gardner) [inline deep_qa/common/params.py]: Oh, yeah, you're right.  I guess our tests should catch if you're using this wrong, anyway, because we check that the parameters are empty in `Trainer` - if you used `get` instead of `pop`, it'll get caught.

File: deep_qa/layers/attention/matrix_attention.py (hunk 13 above)
Comment (DeNeutoy) [inline deep_qa/layers/attention/matrix_attention.py]: You still have the same problem though, right - you have to explicitly set `allow_default` to be True if you want to be able to use choices[0]. With two methods, the usage is implied by the name, whereas here you would(I would) definitely have to check.

Comment (matt-gardner) [inline deep_qa/layers/attention/matrix_attention.py]: You would have to check when reading the code or when writing the code?  I don't think it's particularly obvious that `pop_choice_with_default("type", choices)` will return the first choice if "type" is not a key in the dict - you have to look at the docstring or know the method to know that.  And if that's true, is it much of a stretch to go from there to `pop_choice("type", choices, allow_default=True)`?  One of the first names I thought of for `allow_default` was `use_first_choice_as_default`, which is a lot more clear, but quite long.  What do you think?

File: deep_qa/common/params.py [hunk 25]
```diff
@@ -1,52 +1,184 @@
...
+            value = self.params.pop(key)
+        else:
+            value = self.params.pop(key, default)
+        logger.param(self.history + " : " + key + " : " + str(value))
+        return self.__check_is_dict(key, value)
+
+    @overrides
+    def get(self, key: str, default: Any=DEFAULT):
+        """
+        Performs the functionality associated with dict.pop(key) but with parameter
```
matt-gardner: Looks like docstring got copy-pasted.

deep_qa/common/params.py
```diff
@@ -1,52 +1,184 @@
...
+        """
+        if default is self.DEFAULT:
+            value = self.params.get(key)
+        else:
+            value = self.params.get(key, default)
+        return self.__check_is_dict(key, value)
+
+    def pop_choice(self, key: str, choices: List[Any]):
+        """
+        Gets the value of `key` in the `params` dictionary, ensuring that the value is one of the given
```
matt-gardner: docstrings get rendered using rST now, instead of markdown, so these need double backticks.

deep_qa/common/params.py
```diff
@@ -1,52 +1,184 @@
...
+        choices.  `name` is an optional description for where a configuration error happened, if there
+        is one.
+
+        Note that this _pops_ the key from params, modifying the dictionary, consistent with how
```
matt-gardner: And this needs single backticks.

deep_qa/common/params.py
```diff
@@ -1,52 +1,184 @@
...
+        Sometimes we need to just represent the parameters as a dict, for instance when we pass
+        them to a Keras layer(so that they can be serialised).
+        """
+        def log_recursively(parameters, history):
+            for key, value in parameters.items():
+                if isinstance(value, dict):
+                    new_local_history = value if history == "" else history + "." + key
+                    log_recursively(value, new_local_history)
+                else:
+                    logger.param(self.history + self.history + " : " + key + " : " + str(value))
```
matt-gardner: Looks like you've duplicated `self.history` here.

deep_qa/layers/attention/matrix_attention.py
DeNeutoy: Ok I agree that I would have to look, but when I do look, the current docstrings are immediately obvious - I don't have to think at all to understand what they do, whereas in the combined method, I would have to enumerate all 4 combinations in my head before I understand how to use it. Maybe i'm being picky but I think it's easier.

deep_qa/common/params.py
```diff
@@ -1,52 +1,184 @@
...
+    def pop(self, key: str, default: Any=DEFAULT):
+        """
+        Performs the functionality associated with dict.pop(key), along with checking for
+        returned dictionaries, replacing them with Param objects with an updated history.
+        """
+        if default is self.DEFAULT:
+            value = self.params.pop(key)
+        else:
+            value = self.params.pop(key, default)
+        logger.param(self.history + " : " + key + " : " + str(value))
```
matt-gardner: I'd probably change this format slightly, here and below.  If you just do `history + "." + key`, this is directly readable with pyhocon, because, e.g., `embeddings_dim.words = 100` is valid hocon.  Doing it this way also suggests that you can just have `history` always end with a `.`, so you only need to do `history + key` here, and it will just work when `history` is empty, too.

deep_qa/common/params.py
```diff
@@ -1,52 +1,184 @@
...
+        del self.params[key]
+
+    def __iter__(self):
+        return iter(self.params)
+
+    def __len__(self):
+        return len(self.params)
+
+
+def _get_choice_error_message(value: Any, choices: List[Any], name: str=None) -> str:
```
matt-gardner: Double underscore here at the beginning of the method (I know I originally wrote it this way; I've come to think that actually making these methods private with double underscores is probably better, if you really don't intend them to be used anywhere else).

deep_qa/common/params.py
```diff
@@ -1,52 +1,184 @@
...
+        if default is self.DEFAULT:
+            value = self.params.get(key)
+        else:
+            value = self.params.get(key, default)
+        return self.__check_is_dict(key, value)
+
+    def pop_choice(self, key: str, choices: List[Any]):
+        """
+        Gets the value of `key` in the `params` dictionary, ensuring that the value is one of the given
+        choices.  `name` is an optional description for where a configuration error happened, if there
```
matt-gardner: `name` isn't an argument anymore.

deep_qa/common/params.py
```diff
@@ -1,52 +1,184 @@
...
+def pop_choice_with_default(params: Dict,
                             key: str,
                             choices: List[Any],
                             default: Any=None,
-                            name: str=None):
+                            name: str=None) -> Any:
     """
-    Like get_choice, but with a default value.  If `default` is None, we use the first item in
-    `choices` as the default.
+    Performs the same function as the pop_with_default_method of Params,
```
matt-gardner: ```Performs the same function as :func:`Params.pop_with_default`, ...```

deep_qa/common/params.py
```diff
@@ -1,52 +1,184 @@
...
+    but is required in order to deal with places that the Params object is not
+    welcome, such as inside Keras layers.
     """
     try:
-        return get_choice(params, key, choices, name)
```
matt-gardner: Is there no more `pop_choice` method at all?

deep_qa/common/params.py
```diff
@@ -1,52 +1,184 @@
...
+        raise ConfigurationError(_get_choice_error_message(default, choices, name))
+
+    logger.param("UNSCOPED PARAMETER: Retrieve default arguments manually."
+                 + " : " + key + " : " + str(value))
+
+    return value
 
 
-def assert_params_empty(params: Dict[str, Any], class_name: str):
+def assert_params_empty(params: Union["Params", Dict], class_name: str):
```
matt-gardner: This shouldn't need quotes around `Params`, because `Params` has already been defined at this point.

deep_qa/contrib/background_search/vector_based_retrieval.py
```diff
@@ -1,10 +1,10 @@
 import gzip
 import logging
-from typing import Any, Dict, List, Tuple
+from typing import List, Tuple
 
 import numpy
 
-from ...common.params import assert_params_empty, get_choice_with_default
+from ...common.params import assert_params_empty, Params  # pylint: disable=unused-import
```
matt-gardner: You can remove the unused import directive.

deep_qa/layers/tuple_matchers/embedded_tuple_matcher.py
```diff
@@ -18,17 +18,14 @@ class EmbeddedTupleMatcher:
...
         then pass the rest of the parameters on to that class as ``kwargs``.
     """
-    def __init__(self, text_trainer, tuple_matcher_params: Dict[str, Any]=None):
+    def __init__(self, text_trainer, tuple_matcher_params: Dict[str, Any]):
         self.text_trainer = text_trainer
-        if tuple_matcher_params is None:
-            tuple_matcher_params = {}
-        tuple_matcher_choice = get_choice_with_default(tuple_matcher_params,
-                                                       "embedded_matcher_type",
+        tuple_matcher_choice = pop_choice_with_default(tuple_matcher_params, "embedded_matcher_type",
```
matt-gardner: I think the only change necessary in this file is changing `get` to `pop`, and you can just revert the rest of them.  This doesn't currently allow a default value for `tuple_matcher_params`, and it would crash if you gave it `None`.

deep_qa/layers/tuple_matchers/slot_similarity_tuple_matcher.py
```diff
@@ -19,7 +19,7 @@ class SlotSimilarityTupleMatcher(MaskedLayer):
     to determine entailment we find the cosine similarity between these encoded slot representations,
     i.e., the similarity between the first slot in each, then the second slot in each, etc.
     This generates a set of similarity features equal to the number of slots in the tuples, which are
-    then fed to a shallow NN with output of size one.  The output of this NN is considered to be the
+    then fed to  a shallow NN with output of size one.  The output of this NN is considered to be the
```
matt-gardner: Extra space added here.

deep_qa/layers/tuple_matchers/slot_similarity_tuple_matcher.py
```diff
@@ -1,11 +1,11 @@
 from copy import deepcopy
-from typing import Any, Dict
+from typing import Dict, Any
```
matt-gardner: Nitpicking here, but alphabetizing these is better than not.  The line should stay how it was.

matt-gardner: In fact, the whole file should stay how it was, except for the change from `get` to `pop`.

deep_qa/layers/tuple_matchers/threshold_tuple_matcher.py
```diff
@@ -1,12 +1,12 @@
 from copy import deepcopy
-from typing import Any, Dict
```
matt-gardner: Same with this file.

deep_qa/models/memory_networks/differentiable_search.py
```diff
@@ -3,14 +3,15 @@
 import pickle
 
 from itertools import zip_longest
-from typing import Any, Dict, List
+from typing import Dict, List  # pylint: disable=unused-import
```
matt-gardner: Just remove the import instead of disabling pylint.

deep_qa/training/text_trainer.py
```diff
@@ -137,9 +137,11 @@ def __init__(self, params: Dict[str, Any]):
...
         self.encoder_params = params.pop('encoder', {'default': {}})
         self.encoder_fallback_behavior = params.pop('encoder_fallback_behavior', 'crash')
-        self.seq2seq_encoder_params = params.pop('seq2seq_encoder', {'default': {"encoder_params": {},
-                                                                                 "wrapper_params": {}}})
-        self.seq2seq_encoder_fallback_behavior = params.pop('seq2seq_encoder_fallback_behavior', 'crash')
+        self.seq2seq_encoder_params = params.pop('seq2seq_encoder',
+                                                 {'default': {"encoder_params": {},
+                                                              "wrapper_params": {}}})
+        self.seq2seq_encoder_fallback_behavior = \
+            params.pop_choice_with_default('seq2seq_encoder_fallback_behavior', 'crash')
```
matt-gardner: Looks like this is wrong - you're not passing a list of choices here, so this will actually return `'c'` as the default.

tests/layers/attention/matrix_attention_test.py
```diff
@@ -6,6 +6,7 @@
 
 from deep_qa.layers.attention.matrix_attention import MatrixAttention
 from deep_qa.layers.wrappers.output_mask import OutputMask
+from deep_qa.common.params import Params
```
matt-gardner: We should probably just leave out `Params` for all of these layer tests, because we're not using `Params` for layers, anyway.  It's good for the rest of the tests, though.

deep_qa/common/params.py
DeNeutoy: Union doesn't like it without the quotes for some reason, otherwise I would have....

deep_qa/layers/attention/matrix_attention.py
matt-gardner: Being nitpicky is ok - I'm nitpicky pretty often myself, as you have experienced...  If we can find something we're both happy with, it'll lead to a better API.  Here's a second try:

```python
def pop_choice(params, key, choices, default_to_first_choice: bool=False, other_default: Any=None):
  """
  Pops ``key`` from ``params``, ensuring that the popped value is in ``choices``.
  If ``key`` is not in ``params``, you can optionally default to the first item in
  ``choices``, or you can pass in a default value (which must also be in ``choices``).
  """
  if key not in params:
    if default is not None:
      if default not in choices:
        raise ConfigurationError
      return default
    elif allow_default:
      return choices[0]
    else:
      raise ConfigurationError("required key not specified; or some similar message")
  value = params.pop(key)
  if value not in choices:
    raise ConfigurationError()
  return value
```

deep_qa/common/params.py
DeNeutoy: This is actually a good example of why we shouldn't do this: using double underscores in function names and then using them within methods of a class doesn't work, because when you use a double underscore _anywhere_ in a class, all that happens is that `__function_name` gets replaced with `_classname_function_name`. However, if `__function_name` is not a method on the class, it won't be defined.

matt-gardner: Oh, hmm...  Ok, then, I stand corrected =).  Keep it as it is.

DeNeutoy:
Bah, I rebased instead of merging. @matt-gardner is this ok, given that the commits will be squashed anyway?

Issue #315, opened 2017-04-20
Allow for non-Keras optimizers


matt-gardner:
Using plain tensorflow optimizers instead of Keras optimizers is beneficial in some instances (particularly when you have a large embedding matrix, as Matt Peters has discovered).  We should split out the actual training loop into something configurable, to allow for different means of optimizing the same computation graph.  You can still use `_build_model` just like we normally do, you just pull out the inputs and outputs and pass them directly into a tensorflow optimizer instead of calling `model.fit()`.

matt-gardner:
Note that if we do this, you could also conceivably use something like PyTorch to build your model, and just have a PyTorch optimizer.  The only real benefit there would be to re-use the data processing code, though - you couldn't use any of the layers we have, or the higher-level `TextTrainer` API.

PR #317, opened 2017-04-21
simple switch to tensorflow optimizers

DeNeutoy:
#315 

This is a very simple change which allows us to use tensorflow optimisers directly. However, this comes with a few caveats:

1. It won't work with Keras `Constraints`. This is because tensorflow returns an operation which does the gradient updates for all variables simultaneously and doesn't return anything, whereas the Keras optimisers apply them individually and return the updated tensor, to which the constraint is applied. It's probably not worth our time getting around this problem, given this is a short term thing until we switch to full TF Keras.

2. If you want to pass arguments to the TF optimiser class you need to use a dictionary of keyword arguments which correspond to their arguments, not the Keras ones. I think this is reasonable given the way we pass arguments to `DeepQAModels` anyway. 

This should fix the `IndexedSlices` problem in that gradient updates for embeddings will be sparse, but doesn't address the training loop separation.


deep_qa/training/optimizers.py
```diff
@@ -15,18 +15,114 @@
...
+            if constraints:
+                raise ValueError('TF optimizers do not support '
+                                 'weights constraints. Either remove '
+                                 'all weights constraints in your model, '
+                                 'or use a Keras optimizer.')
+            grads = self.optimizer.compute_gradients(loss, params)
+
+            if self.clip_norm is not None:
+                grads = [clip_by_norm(grad, self.clip_norm) for grad in grads]
+            if self.clip_value is not None:
```
matt-peters: I don't think that tf.clip_by_value handles sparse gradients and converts them to dense values.
tf.clip_by_global_norm does the right thing with IndexedSlices.

deep_qa/training/optimizers.py
```diff
@@ -15,18 +15,114 @@
...
+        def get_updates(self, params, constraints, loss):
+            if constraints:
+                raise ValueError('TF optimizers do not support '
+                                 'weights constraints. Either remove '
+                                 'all weights constraints in your model, '
+                                 'or use a Keras optimizer.')
+            grads = self.optimizer.compute_gradients(loss, params)
+
+            if self.clip_norm is not None:
+                grads = [clip_by_norm(grad, self.clip_norm) for grad in grads]
```
matt-peters: I'm not sure how keras implements this... but it's generally recommended to use tf.clip_by_global_norm (see https://arxiv.org/abs/1211.5063)

deep_qa/training/optimizers.py
```diff
@@ -15,18 +15,114 @@
...
+            raise NotImplementedError
+
+    def callable_wrapper(optimiser: optimizer.Optimizer, name: str):
+        """
+        This allows arguments to be unpacked into the tensorflow
+        optimiser if required, but first splits off the "clipnorm" and "clipvalue"
+        optional arguments and passes those to the ``TFOptimizer`` wrapper class.
+        This makes the interface to Tensorflow optimisers identical to Keras ones.
+
+        :param optimiser: An instance of a tensorflow optimiser.
```
matt-gardner: We're using numpydoc to generate our docs, so this should be formatted like:

```
Parameters
----------
optimiser: optimizer.Optimizer
    comment
name: str
    comment

Returns
-------
comment

**Comment by matt-gardner** [comment]
I'm confused as to why this is showing up as not having code coverage in our tests - it makes it look like these optimizers aren't actually being used somehow, even though the default optimizer is `"adam"`.  Any idea what's going on here?

And have you actually tried running this?  Does it help runtimes?

File: deep_qa/training/optimizers.py (hunk 1 above)
Comment (DeNeutoy) [inline deep_qa/training/optimizers.py]: Huh, thanks! I hadn't realised this was a problem with tensorflow.... I wonder why they do this - there must be a reason, that is a pretty big bug if it's not meant to be like this.

**Comment by DeNeutoy** [comment]
Ok so using the TF optimisers gives speedups which look linear in the size of the vocabulary and constant(ish, but still significant) in the size of the word embeddings. Measurements in seconds:

Just an embedding layer: 

|Vocabulary|: 7000
TF: 0.028
Keras: 0.042
_2x speedup_
|Vocabulary|: 70,000
TF: 0.02929
Keras: 0.1470
_7x speedup_

|Vocabulary|: 700,000
TF: 0.0297
Keras: 1.4926
_70x speedup_
Varying only the Embedding size, holding the vocabulary fixed at 70,000:

Embedding size: 128
Tensorflow: 0.00904
Keras: 0.03917
_3x speedup_

Embedding size: 512
TF: 0.02929
Keras: 0.1470
_7x speedup_

Embedding size: 1024
TF: 0.0514
Keras: 0.2659
_5x speedup_

Now for an LSTM with 512 inputs, using 512 dim embeddings with 70,000 vocab size:
TF: 0.64
Keras: 0.71

Now for an LSTM with 512 inputs, using 512 dim embeddings with 700,000 vocab size:

TF: 0.63
Keras: 1.70
_2x speedup_

Now for an LSTM with 128 inputs, using 512 dim embeddings with 700,000 vocab size:

TF: 0.13
Keras: 1.20
_9x speedup_

Overall this certainly looks worthwhile. I'm trying to profile it on one of our actual models, but it's a bit tricky because we don't control the training loop execution, Keras does, but I'll try that this afternoon. 
Finally, the TF optimisers really don't seem to be playing well with how Keras represents sparse tensors - I am still getting errors when using gradient clipping (of any type). Additionally, we have certain places in our code such as the Keras learning rate decay callback which rely on using the Keras optimisers.

**Comment by DeNeutoy** [comment]
Hmm, ok maybe this isn't a straightforward switch - for the GA reader on sciQ, using 300 dim word embeddings, we have:
Keras: 5.2 secs/batch
Tensorflow 5.6 secs/batch.

although this will only have vocab size equal to sciq's vocab, which i'm guessing isn't very large. I was trying it on who did what but my machine ran out of memory. I'll try again tomorrow.

**Comment by DeNeutoy** [comment]
Ok so using Bidaf, we have:

100 dim embeddings:
TF: 1.05 sec/batch average over 20 batches
Keras 1.24 sec/batch average over 20 batches

800 dim embeddings:
TF: 2.5 sec/batch average over 20 batches
Keras 2.9 sec/batch average over 20 batches

**Comment by matt-gardner** [comment]
A free 20% speedup is really good.  I'm not going to complain about that.  I'm still confused about what's going on with the tests, though - it doesn't seem like it should be 41% coverage on the diff.

**Comment by nelson-liu** [comment]
> I'm still confused about what's going on with the tests, though - it doesn't seem like it should be 41% coverage on the diff.

Agreed. I briefly poked at it, and the coverage report definitely seems correct; inserting a `import pdb; pdb.set_trace()` in any of the uncovered regions doesn't make the tests break, so the code definitely isn't being run. As to why, not too sure yet...

**Comment by nelson-liu** [comment]
nice, i found it. the default value of optimizer in [`trainer.py`](https://github.com/allenai/deep_qa/blob/00d36306759cb1c232489f68844371fb727ce2c8/deep_qa/training/trainer.py#L155) is just a string `"adam"`. In the [`optimizer_from_params` function](https://github.com/allenai/deep_qa/pull/317/files#diff-1a42dc6fd9aea80b94b72cdc133a6d9eR147), there's a condition to just return the input params if its a string, without doing anything to the `optimizer` dict. As a result, in this case it's bypassing the dicts completely and thus doesn't instantiate/test any of the wrapper classes.

a quick fix is to just change the default value of the `optimizer` param in `Trainer.py` linked above to `{"type": "adam"}`, as opposed to just `"adam"`, I think. I ran the tests and got some failures, but the coverage on the `optimizers.py` should be reported correctly once those are fixed...

**Comment by matt-gardner** [comment]
Nice, thanks Nelson!

I'd probably just change `optimizer_from_params` to look like this:

```python
if isinstance(params, str):
    optimizer = params
    params = {}
else:
    optimizer = params.pop_choice("type", optimizers.keys())
return optimizers[optimizer](**params)
```

That way you can still just pass in `"adam"` as the optimizer param to `Trainer`.

DeNeutoy:
@matt-gardner I've cleaned up a few things here - the big issue is incorporating learning rate callbacks. The problem is that all the keras optimisers have a variable as the learning rate, which is updated by the callback. However, I can't bind the tensorflow ones to the same class variable because 1) they are not tensors and 2) different optimisers have different variable names for the learning rate, which makes doing this kind of tricky....

Also, there is a test in `text_trainer` which I cannot figure out for the life of me why it isn't passing. It's throwing an error when it trains because we are managing to pass it a tensor with a negative first dimension 😭 . However, all the other tests don't fail with this and are much more complex(in theory).

matt-gardner:
I'm not sure what's going on with the test, but I can add some context - it's failing when `model.fit` is called a second time on the same model.  Can you reproduce that in a repl with a simple model?  I don't know _why_ this might be happening, and probably won't have time to figure it out today, but that's my guess for what is going on here.

deep_qa/training/models.py
```diff
@@ -28,6 +35,51 @@ def _summary_with_mask_info(self):
...
+                inputs += [K.learning_phase()]
+
+            # Here we override Keras to use tensorflow optimizers directly.
+            self.global_step = K.variable(0., name='global_step')
+            grads = self.optimizer.compute_gradients(self.total_loss, self._collected_trainable_weights)
+            if self.gradient_clipping is not None:
+                clip_type = self.gradient_clipping.pop("type")
+                clip_value = self.gradient_clipping.pop("value")
+                if clip_type == 'clip_by_norm':
+                    grads, _ = tensorflow.clip_by_global_norm(grads, clip_value)
```
matt-peters: `grads`, as returned from `self.optimizer.compute_gradients` is a list of `[gradient_tensor, variable]` pairs. `tf.clip_by_global_norm` just takes a list of tensors.  So you need to do something like:

```python
grad_tensors = [g for g, v in grads]
grad_vars = [v for g, v in grads]
clipped_grad_tensors = tf.clip_by_global_norm(t_list)
clipped_grads_and_vars = list(zip(clipped_grad_tensors, grad_vars))
...
training_updates = self.optimizer.apply_gradients(clipped_grads_and_vars, global_step=self.global_step)
```

matt-peters: Also -- there is one more detail to worry about.  Sometimes, the gradient tensor can be `None` for a particular variable that doesn't have a gradient.  So you also need to filter them out of the list of `grad_tensors` before passing to `tf.clip_by_global_norm`.

DeNeutoy: Ah thanks a lot, i'd always used tf.gradients and I hadn't realised the signatures would be different. Looks like the `None` problem has been fixed in TF 1.0 too 👍

DeNeutoy:
@matt-gardner - this now works - the only annoying thing is that I had to remove the learning rate scheduling in the GA Reader. The Keras LR callback is not going to work with the tensorflow optimisers as the learning rate is bound to different attributes for different optimisers - however, there are automatic learning rate decay functions in TF which we could use easily. I wanted to get this merged before it became too big though. I think it's a good first step towards exposing the actual session.run calls and I now have a better understanding of how that actually works in Keras, so it should be easier to do other cool stuff like arbitrary tensorboard logging within models.

deep_qa/training/models.py
```diff
@@ -28,6 +35,62 @@ def _summary_with_mask_info(self):
...
+        our tensorflow optimiser in a keras wrapper, which we don't want. We override the
+        only method in ``Model`` which uses this attribute, ``_make_train_function``, which
+        raises an error if compile is not called first.
+
+        As we move towards using a Tensorflow first optimisation loop, more things will be
+        added here which add functionality to the way Keras runs tensorflow Session calls.
+
+        """
+        optimizer = params.get('optimizer')
+        self.gradient_clipping = params.pop("gradient_clipping", None)
```
matt-gardner: Did this used to be inside of the optimizer params?  And now it's not, because tf does things differently?

deep_qa/training/models.py
```diff
@@ -28,6 +35,62 @@ def _summary_with_mask_info(self):
...
+        if not hasattr(self, 'train_function'):
+            raise RuntimeError('You must compile your model before using it.')
+        if self.train_function is None:
+            inputs = self._feed_inputs + self._feed_targets + self._feed_sample_weights
+            if self.uses_learning_phase and not isinstance(K.learning_phase(), int):
+                inputs += [K.learning_phase()]
+
+            # Here we override Keras to use tensorflow optimizers directly.
+            self.global_step = K.variable(0., name='global_step')
+            grads = tensorflow.gradients(self.total_loss, self._collected_trainable_weights)
```
matt-gardner: `s/grads/gradients/`

deep_qa/training/models.py
```diff
@@ -28,6 +35,62 @@ def _summary_with_mask_info(self):
...
+            if self.gradient_clipping is not None:
+                # Don't pop from the gradient clipping dict here as
+                # when we load models we need it to still be there.
```
matt-gardner: You mean if you call `fit()` twice, you need it to still be there.  The issue with this is that the parameter won't get logged, though.  Can you call `self.gradient_clipping.as_dict()`, then read from that?

deep_qa/training/models.py
```diff
@@ -28,6 +35,62 @@ def _summary_with_mask_info(self):
...
+                clip_value = self.gradient_clipping.get("value")
+                if clip_type == 'clip_by_norm':
+                    grads, _ = tensorflow.clip_by_global_norm(grads, clip_value)
+                elif clip_type == 'clip_by_value':
+                    grads = [tensorflow.clip_by_value(x, -clip_value, clip_value) for x in grads]
+                else:
+                    raise ConfigurationError("{} is not a supported type of gradient clipping.".format(clip_type))
+
+            zipped_grads_with_weights = zip(grads, self._collected_trainable_weights)
+            # pylint: disable=no-member
```
matt-gardner: What's this check for?  `self.optimizer.apply_gradients`?  Any way to make it so this is not necessary?  And if the answer is no, can you re-enable it after the offending line?

deep_qa/training/optimizers.py
```diff
@@ -14,18 +14,27 @@
 \* I should also note that Keras is an incredibly useful library that does a lot
 of things really well. It just has a few quirks...
 """
+import logging
 from typing import Union
-from keras.optimizers import SGD, RMSprop, Adagrad, Adadelta, Adam, Adamax, Nadam
+
+# pylint: disable=no-name-in-module
```
matt-gardner: I think you can get around this by adding tensorflow here: https://github.com/allenai/deep_qa/blob/82534a7d6340ee986e3c9be28f30f8eb11cf8a19/.pylintrc#L35

deep_qa/training/trainer.py
```diff
@@ -684,8 +685,10 @@ def __compile_kwargs(self):
         member variables that we use to set arguments for model.compile(), we group those arguments
         together here, to only specify them once.
         """
-        return {
+        # TODO(mark): factor all compile kwargs into a single dict in the json config.
+        return Params({
+                'gradient_clipping': self.gradient_clipping,
                 'loss': self.loss,
                 'optimizer': self.optimizer,
-                'metrics': self.metrics,
```
matt-gardner: Keeping the last comma is just fine in python, and I think it's actually encouraged, because it makes the code easier to modify later.

deep_qa/training/models.py
DeNeutoy: Yes + yes, not really a way we can avoid this without writing wrappers for the tf optimisers.

deep_qa/training/models.py
DeNeutoy: The problem is that model assigns `optimiser` in `compile`, but then we override it with a tensorflow optimiser. I didn't override compile, as it would require copying loads of code - this is probably something we could do eventually but for the moment this was considerably simpler, if a bit hacky. I'll add the enable after that line.

deep_qa/training/models.py
DeNeutoy: Ah good point, thanks.

deep_qa/training/optimizers.py
DeNeutoy: This didn't work locally and googling didn't yield much useful stuff....

matt-gardner: Hmm, ok.  Tensorflow does its own crazy stuff, so it's not too surprising that it didn't work.

DeNeutoy: good to merge?

matt-gardner: Yes.

PR #318, opened 2017-04-22
Cleaning up the parameter API a bit

matt-gardner:
@DeNeutoy, I had an idea that I _think_ you'll be happy with.  I realized that we never used the `default` argument to `pop_choice_with_default`, so we didn't need it.  That made it so there's only _one_ default option to `pop_choice`, and the API is now simpler and cleaner.

While I was at it, I improved the documentation a bit, and cleaned up a few other methods dealing with parameters.


deep_qa/common/params.py
```diff
@@ -118,15 +141,21 @@ def log_recursively(parameters, history):
...
+        an argument so that the error message gives some idea of where an error happened, if there
+        was one.  ``class_name`` should be the name of the `calling` class, the one that got extra
+        parameters (if there are any).
+        """
+        if len(self.params) != 0:
+            raise ConfigurationError("Extra parameters passed to {}: {}".format(class_name, self.params))
 
     def __getitem__(self, key):
-        return self.__check_is_dict(key, self.params[key])
+        if key in self.params:
```
DeNeutoy: Won't the dictionary raise this error if the key is not present anyway?

deep_qa/common/params.py
```diff
@@ -140,47 +169,31 @@ def __iter__(self):
...
+
+    This method adds a ``history`` parameter, in the off-chance that you know it, so that we can
+    reproduce :func:`Params.pop_choice` exactly.  We default to using "?." if you don't know the
+    history, so you'll have to fix that in the log if you want to actually recover the logged
+    parameters.
     """
-    if len(params) != 0:
-        raise ConfigurationError("Extra parameters passed to {}: {}".format(class_name, params))
+    value = Params(params, history).pop_choice(key, choices, default_to_first_choice)
+    return value
```
DeNeutoy: Can Keras serialise Layer attributes which are `Param` objects? Is it only the arguments to the layer that matter?

deep_qa/common/params.py
matt-gardner: Because it's actually a pyhocon object, it raises a different error, which breaks the typical dictionary API.  Confusing and unfortunate...

deep_qa/common/params.py
matt-gardner: It only serializes stuff that shows up in `get_config`.  So we save all of the layer parameters to `self` as we got them (as `dicts`), and return those in `get_config`, so it doesn't matter what we do with them after that.

PR #319, opened 2017-04-22
Got dynamic padding actually working for sequence tagger

matt-gardner:
The main issue was with the CNN encoder used to get word encodings.  I needed to make it agnostic to the input length, which it now is.

This led to a ~4x speedup in runtime, going from over 4000s per epoch to ~1000s.  And that's an automatic gain, with a couple of flags, for any model that's compatible.  I'm pretty happy about this.  @matt-peters, thanks for the idea!  The only issue is that some of our more complex models will be hard to make compatible with dynamic padding...


deep_qa/training/trainer.py
```diff
@@ -319,7 +319,7 @@ def train(self):
         # so you may have left it above zero on accident.
         if self.validation_arrays is not None:
             kwargs['validation_data'] = self.validation_arrays
-        elif self.validation_split > 0.0:
+        elif self.validation_split > 0.0 and isinstance(self.training_arrays, tuple):
```
DeNeutoy: Instead of these tuple checks everywhere in `Trainer`, can we use `self.use_data_generator == True` instead? It would be easier to follow the logic.

matt-gardner: Hmm, yeah, you're right.  Trying to remember why I didn't do that in the first place...

Oh, it was because `self.use_data_generator` is only defined in `TextTrainer`, not here.  I thought it wasn't a great idea to have a parameter to `Trainer` that `Trainer` can't actually do anything about - the `Trainer` class doesn't know anything about how data is generated, and you could imagine a subclass that doesn't implement data generators - how should `Trainer` handle the `use_data_generator` parameter in that case?

Anyway, I think the right way to fix this is to add a field here, not a parameter, like `uses_data_generator`, and check that instead of having type checks for `tuple`.  You're right that this would be _way_ more readable.  Subclasses will have to set that field in their constructor, after the call to `super().__init__`.  I'll make this change before merging.

DeNeutoy: Can't it be in `Trainer` with a default value of false? That way you'd have to pass it explicitly in the parameter file, and presumably you'd know whether or not your subclass can handle it. But yeah, a field would also be fine.

matt-gardner: Yeah, it will still be a parameter to `TextTrainer`, but a field in `Trainer`, that gets set to `True` when the parameter in `TextTrainer` is `True`.

DeNeutoy: oh right gotcha 👍

DeNeutoy: Hmm actually i've re-thought about this and it seems to break quite heavily with the parameter popping idea and i'm not so sure it's a good idea - having parameters interact across classes is quickly going to get complicated.

matt-gardner:
@DeNeutoy, to address a comment that (I think) you deleted, you're right.  This is similar to the `sorting_keys` stuff I added a little while ago - I originally had that as a field on `TextTrainer` that subclasses would override, but decided it was easier to document and resulted in a cleaner API to make it a method instead.  I think that's appropriate to do here, as well, just so we're consistent about what kinds of things you have to override in subclasses.  It is indeed really complicated if you have to be careful to pop a parameter before calling the superclass, then set a field _after_ calling the superclass.  Too much magic.  I'll switch it to a method, so parameter passing will be more sane, and so that it shows up in the docs.

matt-gardner:
@DeNeutoy, here's another try.  Does this look better?

DeNeutoy:
Much better ☺️

PR #320, opened 2017-04-23
Got dynamic padding working for BiDAF

matt-gardner:
I was actually surprised that this worked; I didn't think `MatrixAttention` was going to be compatible, but it worked, with some modification.  This depends on a couple of previous PRs, so we should wait to review it until those other two are merged, but I was excited that I got this working, so I'm opening a PR now.  This again gives a 4-5x speed up in running time, plus a much faster start-up time, because you don't have to do all of the padding up front.  Before this, I was seeing ~37-38 hours per epoch, and with this, it's 7-8 hours per epoch.  Still pretty slow, but _much_ better than before.  I'm hoping that with TF optimizers and some more profiling work, we can get this down lower, to be more comparable with Min's implementation.  I think it's still ~3x slower than Min's, maybe as much as 5x.


matt-gardner:
I just tried increasing the batch size to 60, and it's now ~1.5 hours per epoch, which is comparable to Min's.  I'm afraid it might run out of GPU memory when it gets to some of the longer instances, though - we'll see.

matt-gardner:
And.... it ran out of memory.  Too bad.  But at least that hints that if we can bring down memory usage a bit, somehow, we should have an implementation that's just as fast as Min's, but a whole lot cleaner and more extensible.

nelson-liu:
> I'm afraid it might run out of GPU memory when it gets to some of the longer instances, though - we'll see.

What max length are you using? or are you not truncating?

matt-peters:
Nice to see this and those are some impressive speedups!

RE: OOM.  A few easy things to do to reduce memory pressure:

* put embedding layers on CPU.  Typically these are most of the model parameters and since they are only used for lookups they aren't needed on GPU.
* use tf based optimizers.  The dense gradient updates in keras use a lot of memory
* smaller batch sizes

matt-gardner:
I wasn't truncating at all, because the code isn't set up yet to do that well with dynamic padding.

But, here's a nice idea: use adaptive batch sizes.  When the instances are small, increase the batch size, and when they are large, decrease the batch size.  We could do this really easily with some simple heuristics right now, but doing it well, and automatically, seems pretty challenging.

nelson-liu:
> I wasn't truncating at all, because the code isn't set up yet to do that well with dynamic padding.

I'd guess that this (the original BiDAF code truncates aggressively) + making sure embeddings are on the CPU, at least, should solve our memory issues...

deep_qa/layers/backend/repeat_like.py
```diff
@@ -0,0 +1,63 @@
...
+        - The input tensor repeated along one of the dimensions.
+
+    Parameters
+    ----------
+    axis: int
+        We will add a dimension to the input tensor at this axis.
+    copy_from_axis: int
+        We will copy the dimension from the second tensor at this axis.
+    """
+    def __init__(self, axis: int, copy_from_axis: int, **kwargs):
```
DeNeutoy: Could these be optional parameters to `Repeat` instead? Something like 
```
if copy_from_axis is not None and copy_tensor is not None:
    RepeatLike.call(...)
else:
    Repeat.call(...)

File: deep_qa/layers/attention/weighted_sum.py [hunk 2]
```diff
@@ -71,39 +71,51 @@ def compute_output_shape(self, input_shapes):
...
         if self.use_masking and matrix_mask is not None:
-            matrix_mask = self._expand_matrix_if_necessary(matrix_mask, matrix_shape[:-1], attention_vector)
+            matrix_mask = self._expand_matrix_if_necessary(matrix_mask, attention_vector, K.ndim(matrix) - 1)
             # Doing a multiply here instead of a `switch` to avoid allocating another large tensor.
             matrix = K.cast(K.expand_dims(matrix_mask), 'float32') * matrix
         return K.sum(K.expand_dims(attention_vector, axis=-1) * matrix, -2)
 
     @staticmethod
-    def _expand_matrix_if_necessary(matrix, matrix_shape, attention_vector):
+    def _expand_matrix_if_necessary(matrix, attention_vector, num_matrix_dims: int):
```
matt-gardner: You mean do `K.ndim(matrix) - 1`?  The trouble is that `K.ndim(matrix)` and `K.ndim(mask)` are different, so I need some way to distinguish them inside the function.  Even though it says `matrix`, something this method will be passed the `mask`.

deep_qa/layers/backend/repeat_like.py
matt-gardner: This would mean that `Repeat` can take as arguments either a single tensor or two tensors.  What does it mean to `Repeat` two arguments?  It's not intuitive what should happen in that case.  I was going for similarity to `K.ones()` and `K.ones_like()`.  Yes, it means there are two classes with similar functionality, but I think the use is distinct enough that having two classes is warranted.

I realize this is the opposite decision from `pop_choice` - I think there, the difference between the two methods was just what you do when the key isn't in the dictionary.  Should you crash, or use the first valid choice as default?  Not much is changing between the two methods, so it made more sense to merge them.  Here, the difference seems a lot bigger to me.

deep_qa/layers/attention/weighted_sum.py
```diff
@@ -71,39 +71,51 @@ def compute_output_shape(self, input_shapes):
 
     @overrides
     def call(self, inputs, mask=None):
+        # pylint: disable=redefined-variable-type
         matrix, attention_vector = inputs
-        matrix_shape = K.int_shape(matrix)
-        matrix = self._expand_matrix_if_necessary(matrix, matrix_shape[:-1], attention_vector)
+        matrix = self._expand_matrix_if_necessary(matrix, attention_vector, K.ndim(matrix) - 1)
```
matt-peters: In what case would you have to expand the dimensions of the matrix?  It should always have one more dimension then the attention.

deep_qa/layers/attention/weighted_sum.py
```diff
@@ -71,39 +71,51 @@ def compute_output_shape(self, input_shapes):
...
-            # the end of the attention_shape exactly.
-            assert matrix_shape[0] == attention_shape[0], "somehow batch sizes don't match"
-            matrix_shape = matrix_shape[1:]
-            attention_shape = attention_shape[1:]
-            assert attention_shape[-len(matrix_shape):] == matrix_shape, ("matrix_shape must be "
-                                                                          "subset of attention_shape")
-            for i in range(len(attention_shape) - len(matrix_shape)):
-                matrix = K.expand_dims(matrix, axis=i+1)  # +1 to account for batch_size
-                matrix = K.repeat_elements(matrix, attention_shape[i], axis=i+1)
+        num_attention_dims = K.ndim(attention_vector)
```
matt-peters: Given that you are just multiplying the expanded matrix above, the tiling isn't necessary.  tf will broadcast the single dimensions to the full size matrix (and depending on how it's implemented may be more efficient).

deep_qa/layers/attention/weighted_sum.py
matt-gardner: This is a fancy weighted sum =). We actually need this in BiDAF, because we're doing a weighted sum of the question, for each passage word.  The inputs [here](https://github.com/allenai/deep_qa/blob/67ce4ebc172ac36d329840b5641806b565982225/deep_qa/models/reading_comprehension/bidirectional_attention.py#L128) are of shape `(batch_size, question_length, encoding_dim)` (for the "matrix"), and `(batch_size, passage_length, question_length)`, for the "attention vector".  We want the result to be `(batch_size, passage_length, encoding_dim)`, summing over the question length.  To do this, we need to expand the "matrix" first.

See also this docstring: 
https://github.com/allenai/deep_qa/blob/67ce4ebc172ac36d329840b5641806b565982225/deep_qa/layers/attention/weighted_sum.py#L37-L53.

deep_qa/layers/attention/weighted_sum.py
matt-gardner: Oh, yeah, that's a good point.  Thanks!  I think I might also be able to do this with `tf.einsum`, which might be more efficient still.  I'll look into that.

matt-gardner: Yeah, that dramatically simplified that piece of code.

deep_qa/layers/attention/weighted_sum.py
matt-peters: Got it

deep_qa/layers/backend/repeat_like.py
DeNeutoy: I guess I was thinking that the tensor you want to tile to would be passed to the init method of `Repeat`, but go for this way, sure.

matt-gardner: Oh, I see.  Yeah, that wouldn't work with how Keras layers are structured - you need to be able to initialize layers from the `get_config` parameters, and you can't put tensors in there.

DeNeutoy: ah of course, sorry.

PR #326, opened 2017-04-27
Multi gpu support

DeNeutoy:
This PR adds data parallelism which allows batch updates to be split across any number of GPUs with synchronous gradient updates performed on the CPU. 

Additionally, I've added a tensorflow device scope to the `_embed_input` function, which should locate all embedding variables on the CPU. 

There are a few things to decide here:
1) Default behaviour when the number of available GPUS is not what it is in the model parameters. At the moment (and by default in Keras), we enable `soft_device_placement=True` in the session configuration, which means that tensorflow will try to put things on specified GPUs, but won't complain if they aren't available and will just allocate them some other way if the devices aren't there.

2) Should we keep the device_placement logging/make it a parameter? It will print all of the operations in the graph and what device they are on, which is a lot of stuff, but also extremely helpful for debugging etc.

### DIFF
--- a/deep_qa/training/multi_gpu.py
+++ b/deep_qa/training/multi_gpu.py
@@ -1,0 +1,47 @@
+from keras.layers import merge
+from keras.layers.core import Lambda
+from keras.models import Model
+
+import tensorflow as tf
+
+
+def make_parallel(model, gpu_count):
+    def get_slice(data, idx, parts):
+        shape = tf.shape(data)
+        size = tf.concat([shape[:1] // parts, shape[1:]], axis=0)
+        stride = tf.concat([shape[:1] // parts, shape[1:] * 0], axis=0)
+        start = stride * idx
+        return tf.slice(data, start, size)
+
+    outputs_all = []
+    for i in range(len(model.outputs)):
+        outputs_all.append([])
+
+    # Place a copy of the model on each GPU, each getting a slice of the batch
+    for i in range(gpu_count):
+        with tf.device('/gpu:%d' % i):
+            with tf.name_scope('tower_%d' % i) as scope:
+
+                inputs = []
+                # Slice each input into a piece for processing on this GPU
+                for x in model.inputs:
+                    input_shape = tuple(x.get_shape().as_list())[1:]
+                    slice_n = Lambda(get_slice, output_shape=input_shape, arguments={'idx': i, 'parts': gpu_count})(x)
+                    inputs.append(slice_n)
+
+                outputs = model(inputs)
+
+                if not isinstance(outputs, list):
+                    outputs = [outputs]
+
+                # Save all the outputs for merging back together later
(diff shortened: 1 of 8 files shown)


matt-peters:
Does this method work and give performance speedups?  E.g. if you run on multiple GPUs and manually inspect nvidia-smi, are all GPUs working equally hard?  Or how much speedup do you get vs just using a larger batch on a single GPU?

It looks like you are creating only one copy of each variable/weight on a single device.  Then multiple input and output tensors separated across devices.  I'm curious what tensorflow would do in this situation:

* copy weights to each GPU to do computation
* copy the input tensors to the device with the weight matrices and do computation there
* something else?

DeNeutoy:
Yeah - I think this is copying weights back and forth to the different GPUs, with the speedup being that this can be done in parallel - however, I need to look into the difference between this and using one of tensorflow's parameter servers, as that seems to be how they want people to do this distributed stuff...

deep_qa/training/text_trainer.py
```diff
@@ -305,7 +306,13 @@ def _embed_input(self, input_layer: Layer, embedding_name: str="embedding"):
         uses both words and characters, we need to run the character encoder and concatenate the
         result with a word embedding).
         """
-        return self.tokenizer.embed_input(input_layer, self.__get_embedded_input, self, embedding_name)
+        with tensorflow.device("/cpu:0"):
```
matt-gardner: I think this is the wrong place for this.  If you're using a `word and characters` tokenizer, `tokenizer.embed_input` gets a word embedding and also does a character-level CNN, then concatenates them.  You only want the embedding lookups to happen on the CPU, not the CNN.  I also am thinking that MattP's pre-trained language model stuff will fit into DeepQA in the call to `self._embed_input`, so this could also get a lot more complex in some scenarios.  I think the right place to put the device scope is [here](https://github.com/allenai/deep_qa/blob/00d36306759cb1c232489f68844371fb727ce2c8/deep_qa/training/text_trainer.py#L602).

deep_qa/training/multi_gpu.py
```diff
@@ -0,0 +1,57 @@
+from keras.layers import merge
+from keras.layers.core import Lambda
+from .models import DeepQaModel
+import tensorflow as tf
+
+
+def make_parallel(model: DeepQaModel, gpu_count: int) -> DeepQaModel:
+
+    """
+    :param model: An instance of a DeepQaModel.
```
matt-gardner: We're using a different docstring format.  This won't work with our doc tools (and even if I'm wrong and it does, we should still just standardize on a single docstring format, for consistency and readability across the library).

deep_qa/training/text_trainer.py
DeNeutoy: Ah yep, I forgot about the cnn encoder stuff. Thanks.

PR #327, opened 2017-04-28
Peek across

bsharpataz:
Peek-across model.  I'll say that line 38 of subtract_minimum particularly bothers me (the arbitrary high number...) and I think there is likely a much more elegant way to accomplish the whole process...


deep_qa/layers/subtract_minimum.py
```diff
@@ -0,0 +1,45 @@
...
+    '''
+    def __init__(self, axis, **kwargs):
+        self.axis = axis
+        super(SubtractMinimum, self).__init__(**kwargs)
+
+    def build(self, input_shape):
+        # Add the trainable weight variable for the noise parameter.
+        super(SubtractMinimum, self).build(input_shape)
+
+    def get_output_shape_for(self, input_shape): # pylint: disable=no-self-use
```
matt-gardner: This should be `compute_output_shape`.  Also, can you add `@overrides` tags to all of the methods here that override superclass methods?  That would have caught this issue.

deep_qa/layers/subtract_minimum.py
```diff
@@ -0,0 +1,45 @@
...
+class SubtractMinimum(MaskedLayer):
+    '''
+    This layer is used to normalize across a tensor axis.  Normalization is done by finding the minimum value
+    across the specified axis, and then subtracting that value from all values (again, across the spcified axis).
+    '''
+    def __init__(self, axis, **kwargs):
+        self.axis = axis
+        super(SubtractMinimum, self).__init__(**kwargs)
+
+    def build(self, input_shape):
```
matt-gardner: You don't need this method at all.

deep_qa/layers/subtract_minimum.py
```diff
@@ -0,0 +1,45 @@
+from keras import backend as K
+
+from deep_qa.layers.masked_layer import MaskedLayer
+
+class SubtractMinimum(MaskedLayer):
+    '''
+    This layer is used to normalize across a tensor axis.  Normalization is done by finding the minimum value
+    across the specified axis, and then subtracting that value from all values (again, across the spcified axis).
```
matt-gardner: Can you add [input/output/parameter docs](https://github.com/allenai/deep_qa/blob/3a8a98b4fe91ab63e11a801896f65c3e2ae6fdd3/deep_qa/layers/backend/repeat_like.py#L17-L29) to this?  I know this layer is pretty simple, so it's not as necessary, but it's helpful for someone reading the docs to have everything consistent.

deep_qa/layers/subtract_minimum.py
```diff
@@ -0,0 +1,45 @@
...
+            #     mask_tile_dimensions = [1] * K.ndim(inputs)
+            #     mask_tile_dimensions[-1] = K.int_shape(inputs)[-1]
+            #     mask = K.tile(mask, mask_tile_dimensions)
+            # # Make all masked values very large
+            mask_flipped_and_scaled = K.cast(K.equal(mask, K.zeros_like(mask)), "float32") * 1000000.0
+            dim_1_minimums = K.min(inputs + mask_flipped_and_scaled, axis=self.axis, keepdims=True)
+        else:
+            dim_1_minimums = K.min(inputs, axis=self.axis, keepdims=True)
+
+        minimums = K.tile(dim_1_minimums, tile_dimensions)
```
matt-gardner: You don't need this tile - tensorflow will broadcast the subtraction, so it still works.  That means you can remove a bunch of the stuff above here, too.

deep_qa/layers/subtract_minimum.py
```diff
@@ -0,0 +1,45 @@
...
+    def compute_mask(self, inputs, mask=None):
+        return mask
+
+    def call(self, inputs, mask=None):
+        axis_dimension = K.int_shape(inputs)[self.axis]
+        tile_dimensions = [1] * K.ndim(inputs)
+        tile_dimensions[self.axis] = axis_dimension
+
+        if mask is not None:
+            # # Handle the case where the mask is one-dimension smaller than the inputs to mask all values for that
```
matt-gardner: Remove commented out code.

matt-gardner:
Pretty surprised that the layer test passed, but it's not working when I put it in an actual model.  I'll look into what's going on with this.

PR #331, opened 2017-05-01
Pinned Keras version, updated README

matt-gardner:
@nelson-liu, does this wording sound ok?  I'm thinking of putting the same wording as a template for issues, so that people read that before opening a new issue.  Is that too much?


nelson-liu:
hmm, I think its ok; the wording is a bit strong but it gets the point across. I don't think its necessary to put this in an issue template, but i'm in general +1 for an issue template so people make sure to include relevant versions, steps to reproduce, stack trace, etc.

PR #332, opened 2017-05-01
Adaptive batch size

matt-gardner:
Still very much in progress, just opening the PR to track things.  If anyone cares at this point, this at least has my basic plan for implementing this in the docstrings, though I haven't done any of the actual work yet.


deep_qa/training/text_trainer.py
```diff
@@ -80,6 +80,20 @@ class TextTrainer(Trainer):
         or not.  It's handled correctly for ``num_sentence_words`` and ``num_word_characters`` in
         this class, but you need to be sure to implement it correctly in subclasses for this to
         work.
```
DeNeutoy: Should we perhaps try to avoid parameters which depend on other parameters? `use_dynamic_padding` could be a dict of parameters instead.

DeNeutoy: Otherwise this looks great :1st_place_medal:

matt-gardner: Yeah, that's a good thought.  There are a bunch of related parameters here, and grouping them all together is probably a good idea.

matt-gardner:
Preliminary result: looks like I can get BiDAF down to less than 45 minutes per epoch using this technique (@nelson-liu should be happy about that =) ).  This is a ~10x speedup, better than the 4-5x from my initial estimate.  It's possible it'll still crash on me before finishing the first epoch, but I'm optimistic.  The largest batches have somewhere around 400 instances in them.

nelson-liu:
wow, that's amazing! Great work @matt-gardner

matt-gardner:
It worked!  Final epoch time: ~34 minutes!

matt-gardner:
Ok, this is close to done.  I still need to add a test for the adaptive grouping, but other than that I think it's ready for review.

deep_qa/training/text_trainer.py
```diff
@@ -64,22 +60,15 @@ class TextTrainer(Trainer):
...
-        If ``True``, we will set padding lengths based on the data `per batch`, instead of on the
-        whole dataset.  This only works if ``use_data_generator`` is ``True``, if your model is
-        structured to allow variable-length sequences (typically using ``None`` for specific
-        dimensions when you build your model), and if you don't set padding values in
-        :func:`~TextTrainer._set_padding_lengths`.  This flag specifically is read in
-        :func:`~TextTrainer._set_padding_lengths` to know if we should set certain padding values
-        or not.  It's handled correctly for ``num_sentence_words`` and ``num_word_characters`` in
-        this class, but you need to be sure to implement it correctly in subclasses for this to
-        work.
+    data_generator: Dict[str, Any], optional (default=None)
```
DeNeutoy: Can we add a fairly prominent warning here that using this will make your batches non-iid if your labels are correlated with input length and that this problem is compounded with smaller amounts of data.

deep_qa/models/reading_comprehension/bidirectional_attention.py
```diff
@@ -222,10 +225,18 @@ def _set_padding_lengths_from_model(self):
         # self.num_sentence_words.
         self._set_text_lengths_from_model_input(self.model.get_input_shape_at(0)[1][1:])
 
-    def _get_instance_sorting_keys(self) -> List[str]:  # pylint: disable=no-self-use
+    @overrides
+    def get_instance_sorting_keys(self) -> List[str]:  # pylint: disable=no-self-use
         return ['num_passage_words', 'num_question_words']
 
+    @overrides
+    def get_padding_memory_scaling(self, padding_lengths: Dict[str, int]) -> int:
```
DeNeutoy: Is dynamic padding only working with Bidaf at the moment? We should probably try and implement it for the other models reasonably soon because it's quite a big feature to have only for a single model. That's definitely a separate issue though.

deep_qa/training/text_trainer.py
```diff
@@ -491,10 +472,12 @@ def _set_padding_lengths_from_model(self):
 
     ########################
     # Model-specific methods - if you do anything complicated, you probably need to override these,
-    # but simple models might be able to get by with just the default implementation
+    # but simple models might be able to get by with just the default implementation.  Some of
```
DeNeutoy: Maybe we should just make all of these methods not private as they are used in subclasses (I know this is not the standard definition of private) and make all the ones which are currently `__func` private. The `__func` does weird class name substitution things and I don't like it - would you agree that this usage is pretty unusual to find in Python code in general?

deep_qa/models/reading_comprehension/bidirectional_attention.py
matt-gardner: Yesterday I updated a couple more; I'll submit PRs for them today, after this one's merged.

deep_qa/training/text_trainer.py
matt-gardner: `__func` is how you actually implement private methods in python (see [here](http://radek.io/2011/07/21/private-protected-and-public-in-python/), [here](http://stackoverflow.com/a/1641236/1467533), and [here](http://stackoverflow.com/questions/1641219/does-python-have-private-variables-in-classes#comment3435444_1641305)).  `_func` is "protected", by convention, which is why pylint complains about "protected-access" (_not_ "private-access") when you try to use them outside of the class where they were defined.

As we're building something that's actually going to be released, being really clear in the code and in the documentation about what's public, what's protected, and what's private is pretty important.  So, yes, I think we should definitely make use of python's conventions for all three.

deep_qa/training/text_trainer.py
matt-gardner: I added a note discussing this.  I'm also going to add some more parameters in another PR today, to do more shuffling, set a maximum batch size, and other stuff, to try to fix some of the bad learning behaviors that can result from this.

DeNeutoy:
LGTM

matt-gardner:
Just a note for anyone who was following this thread:

You might not want to use as large a batch size as possible - this is kind of the whole point of SGD, anyway - you use minibatches, and the variance this causes in your gradient is actually helpful in avoiding local minima.  I ran BiDAF yesterday, fully optimizing GPU memory usage, and some batches had upwards of 400 instances in them.  This gave a running time of 33-34 minutes per epoch, but it also had poorer learning behavior, and only got to ~53% exact match on the dev set, where it should be reaching closer to ~60% (this is using one randomly picked dev annotation, not all three, so numbers are lower than the official evaluation).

I'm running it again, capping batch size at 60 (batches are still smaller for the largest instances, so it fits on the GPU without any truncation), and it's taking ~1.1 hours per epoch, but the learning behavior seems to be a bit better.  It hasn't finished yet, though.  I'll update this thread again when it does.

matt-gardner:
So the run with batch size 60 got to about 55% accuracy on span begin, and 58% on span end, which we could be optimistic and say translates to ~55% exact match, which is still a few percent below Min's implementation (which gets ~60% on this metric).  We haven't tuned any of the learning parameters, or dropout, or regularization, or anything, though.  The point is that the running time of this code is now a little faster than Min's, the code is much cleaner, and if someone wants to, you could tune it to get comparable performance.

Issue #333, opened 2017-05-01
Get random seed into the parameter file


matt-gardner:
The plan for actually implementing model ensembling (#306) is to write a class that takes several model parameter files, loads them all, calls `score_dataset` on all of them, and averages the result.  In order for this to work well, we really should have the random seed as part of the parameter file.

The only real trouble here is that the random seeds have to be set [before keras is imported](https://github.com/allenai/deep_qa/blob/0b6a735176b0b5b7105da0bc60a6353ff2ca83c3/scripts/run_model.py#L6-L12), so this has to be implemented carefully.

PR #334, opened 2017-05-01
move run_model into deep_qa

DeNeutoy:
Just shifted the run_model function into the module and allows you to set random seeds in the parameter files. One thing that caught my eye - at the moment if we aren't training a model, we just load it instead, but don't actually do anything with it. Perhaps we should split this into 3 functions, `train` which does training, `eval`, which loads a model from the file path and runs evaluation on a dataset, and `load` which just loads and returns a model to the user. Thoughts?


DeNeutoy:
#334 #333

matt-gardner:
This looks great, thanks!  Yes, I agree with adding two more functions, like you said.  Also, can you add a test for this?

deep_qa/training/run_model.py
```diff
@@ -0,0 +1,117 @@
+
```
matt-gardner: Extra blank line at the top.  Also, what do you think about calling this `run.py`, instead of `run_model.py`?

deep_qa/training/run_model.py
```diff
@@ -0,0 +1,117 @@
...
+import shutil
+import os
+
+import random
+import pyhocon
+import numpy
+
+# pylint: disable=wrong-import-position
+sys.path.append(os.path.join(os.path.dirname(__file__), ".."))
+from deep_qa.common.params import Params, replace_none, ConfigurationError
```
matt-gardner: These can be relative imports, can't they?  And you shouldn't need the `sys.path.append` stuff.

deep_qa/training/run_model.py
```diff
@@ -0,0 +1,117 @@
...
+        seed = None
+        numpy_seed = None
+    if seed is not None:
+        random.seed(seed)
+    if numpy_seed is not None:
+        numpy.random.seed(numpy_seed)
+
+
+def run_model(param_path: str):
+
```
matt-gardner: Extra blank line here.

deep_qa/training/run_model.py
matt-gardner: And I might just put it at the top level: `deep_qa/run.py`.  And in `deep_qa/__init__.py`, you put the imports, so you can just do `from deep_qa import run_model, evaluate_model, load_model`.

deep_qa/training/run_model.py
```diff
@@ -0,0 +1,117 @@
...
+    model_type = params.pop_choice('model_class', concrete_models.keys())
+    model_class = concrete_models[model_type]
+    model = model_class(params)
+
+    if model.can_train():
+        logger.info("Training model")
+        model.train()
+        K.clear_session()
+    else:
+        raise ConfigurationError("The supplied model does not have enough training inputs.")
```
matt-gardner: `model.train()` already runs tests if you give test data, right?  Maybe we should split that out, but that's something for another PR, if we want to do it.

deep_qa/training/run_model.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+def load_model(param_path: str):
```
matt-gardner: This should have a return type, in the declaration and in the docstring.

deep_qa/training/run_model.py
```diff
@@ -0,0 +1,117 @@
...
+    params = Params(replace_none(param_dict))
+    prepare_environment(params)
+
+    # These have to be imported _after_ we set the random seed,
+    # because keras uses the numpy random seed.
+    from deep_qa.common.checks import log_keras_version_info
+    from deep_qa.models import concrete_models
+    from keras import backend as K
+
+    log_keras_version_info()
```
matt-gardner: Maybe put this at the end of `prepare_environment()`?

deep_qa/training/run_model.py
```diff
@@ -0,0 +1,117 @@
...
+    """
+    param_dict = pyhocon.ConfigFactory.parse_file(param_path)
+    params = Params(replace_none(param_dict))
+    prepare_environment(params)
+
+    from deep_qa.models import concrete_models
+    model_type = params.pop_choice('model_class', concrete_models.keys())
+    model_class = concrete_models[model_type]
+    model = model_class(params)
+    # TODO(matt): figure out a way to specify which epoch you want to load a model from.
```
matt-gardner: You can just remove this TODO, I think.

deep_qa/training/run_model.py
```diff
@@ -0,0 +1,117 @@
...
+    model.load_model()
+    return model
+
+
+def evaluate_model(param_path: str, test: str=False):
```
matt-gardner: This also needs a return type in the declaration and the docstring.

scripts/run_model.py
```diff
@@ -1,63 +1,18 @@
...
-import pyhocon
-
-sys.path.append(os.path.join(os.path.dirname(__file__), ".."))
-from deep_qa.common.checks import ensure_pythonhashseed_set, log_keras_version_info
-from deep_qa.common.params import Params, replace_none
-from deep_qa.common.tee_logger import TeeLogger
-from deep_qa.models import concrete_models
-from keras import backend as K
+from deep_qa.common.checks import ensure_pythonhashseed_set
+from deep_qa.training import run_model
```
matt-gardner: This is actually going to import keras, before `prepare_environment` is called, because you're running `deep_qa/training/__init__.py`, which imports `Trainer`, which imports keras.  I guess it's not just for cosmetic reasons that we need to move this to `deep_qa/run.py` - we _have_ to in order to get the import order right.

DeNeutoy:
@matt-gardner I've added the ability for a user to specify a new `DeepQaModel` subclass when calling train/evaluate/load to get around the problem of having to add the model to `concrete_models`. I think this is a good solution - if a user has built there own model, we should probably trust that they have done it correctly. What do you reckon?

scripts/run_model.py
```diff
@@ -1,63 +1,18 @@
...
+from deep_qa import run_model
+from deep_qa.common.checks import ensure_pythonhashseed_set
 
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
+
 def main():
     if len(sys.argv) != 2:
-        print('USAGE: run_model.py [param_file]')
+        print('USAGE: run.py [param_file]')
```
matt-gardner: You didn't change the name of this script.  I'd probably actually keep this one as `run_model.py`, matching `run.run_model`.  But maybe that's too confusing...  Feel free to pick one, but make sure you've fixed it everywhere - various places in this PR have this as both `run.py` and `run_model.py`.

deep_qa/run.py
```diff
@@ -0,0 +1,137 @@
...
+    if seed is not None:
+        random.seed(seed)
+    if numpy_seed is not None:
+        numpy.random.seed(numpy_seed)
+
+    from deep_qa.common.checks import log_keras_version_info
+    log_keras_version_info()
+
+
+def train_model(param_path: str, model_class=None):
```
matt-gardner: I actually think this should stay as `run_model`, because it trains _and_ evaluates a model, depending on what's in the parameters you're passing.  We should also note this after the first sentence of the docstring.

deep_qa/run.py
```diff
@@ -0,0 +1,137 @@
...
+    """
+    This function is the normal entry point to DeepQA. Use this to run a DeepQA model in
+    your project. Note that if you care about exactly reproducible experiments,
+    you should avoid importing Keras before you import and use this function, as
+    Keras relies on random seeds which can be set in this function via a
+    JSON specification file.
+
+    Parameters
+    ----------
+    param_path: str, required. A json file specifying a DeepQaModel.
```
matt-gardner: The description needs to be indented, on the next line.

deep_qa/run.py
```diff
@@ -0,0 +1,137 @@
...
+    model_class: A DeepQaModel, optional. This option is useful if you have
```
matt-gardner: Same here, and it should just be: `model_class: DeepQaModel, optional (default=None)` on the first line.

deep_qa/run.py
```diff
@@ -0,0 +1,137 @@
...
+    implemented a new model class which is not one of the ones implemented
+    in this library.
+    returns: None.
```
matt-gardner: Just omit this if there's no return value.

deep_qa/run.py
```diff
@@ -0,0 +1,137 @@
...
+    """
+    Loads and returns a model.
+
+    Parameters
+    ----------
+    param_path: str, required. A json file specifying a DeepQaModel.
+    model_class: A DeepQaModel, optional. This option is useful if you have
+    implemented a new model class which is not one of the ones implemented
+    in this library.
+    returns: A DeepQaModel instance.
```
matt-gardner: This should be formatted as

```
Returns
-------
A ``DeepQaModel`` instance
```

matt-gardner:
Also, you can see the generated docs here: https://750-47793051-gh.circle-artifacts.com/0/home/ubuntu/deep_qa/doc/_build/html/training/misc.html.  Looks like `deep_qa/run.py` doesn't show up anywhere yet.  We should add it somewhere in there.  Where do you think it should go?

README.md
```diff
@@ -46,6 +46,29 @@ data processing code directly into DeepQA, so that DeepQA Experiments is not nec
...
 
+## Running a Model
+Running a model in DeepQA is very straightforward. Once you have specified model parameters in a
+json file you can run the following:
+
+```
+from deep_qa import run_model, evaluate_model, load_model
+
+# Train a model given a json specification
+train_model("/path/to/json/parameter/file")
```
matt-gardner: This should now be `run_model`, because you changed it back.

README.md
```diff
@@ -56,12 +79,25 @@ section.  For a simple example of a fully functional model, see the [simple sequ
...
-One snag is that if you're doing a new task, or a new variant of a task with a different
+In order to train, load and evaluate models which you have written yourself, simply pass an additional
+argument to the functions above and remove the "model_class" parameter from your json specification.
+For example:
+```
+from deep_qa import run_model
+from .local_project import MyGreatModel
+
+# Train a model given a json specification (without a "model_class" attribute).
+train_model("/path/to/json/parameter/file", model_class=MyGreatModel)
```
matt-gardner: `run_model` again here.

deep_qa/contrib/background_search/retrieval_encoders.py
```diff
@@ -149,7 +149,7 @@ class SentenceSelectionRetrievalEncoder(RetrievalEncoder):
     Parameters
     ----------
     model_param_file: str
-        This is the parameter file used to train the sentence selection model with ``run_model.py``.
+        This is the parameter file used to train the sentence selection model with ``run.py``.
```
matt-gardner: Can you change ``` ``run.py`` ``` to ```:func:`~deep_qa.run.run_model()` ```?

README.md
```diff
@@ -33,7 +33,7 @@ code for these systems is typically 50 lines or less.
 ## Using DeepQA
 
 To train or evaluate a model using DeepQA, the recommended entry point is to use the
-[`run_model.py`](./scripts/run_model.py) script.  That script takes one argument, which is a
+[`run.py`](./scripts/run.py) script.  That script takes one argument, which is a
```
matt-gardner: What did you decide on `run.py` vs. `run_model.py`?  If you want it to be `run.py`, there are lots of places below that should be switched.  I think this is the only place that currently has `run.py`.

README.md
```diff
@@ -56,12 +79,25 @@ section.  For a simple example of a fully functional model, see the [simple sequ
 tagger](./deep_qa/models/sequence_tagging/simple_tagger.py), which has about 20 lines of actual
 implementation code.
 
-One snag is that if you're doing a new task, or a new variant of a task with a different
+In order to train, load and evaluate models which you have written yourself, simply pass an additional
+argument to the functions above and remove the "model_class" parameter from your json specification.
```
matt-gardner: Backticks around "model_class", I think.

deep_qa/run.py
```diff
@@ -0,0 +1,160 @@
...
+    if numpy_seed is not None:
+        numpy.random.seed(numpy_seed)
+
+    from deep_qa.common.checks import log_keras_version_info
+    log_keras_version_info()
+
+
+def run_model(param_path: str, model_class=None):
+    """
+    This function is the normal entry point to DeepQA. Use this to run a DeepQA model in
```
matt-gardner: How about this for the first two sentences: "This function is the normal entry point to DeepQA.  It typically trains, validates, and tests the model specified in the parameters in `param_path`, but can do a subset of that, depending on the files included in the parameters."

And now that I look more at `model.train()`, the more I think that calling this `train_model` actually does make sense...  Sorry, I'm going back and forth here.  I say we should keep it like this for now, but maybe refactor things in the future, so that the testing functionality is moved out of `model.train()`, and then this can be called `train_model`.

matt-gardner: Oh, I just got to the second paragraph and noticed that you already have this in there.  Maybe having it down below is fine...  Feel free to ignore all of this rambling if you want.

deep_qa/run.py
```diff
@@ -0,0 +1,160 @@
...
+
+    Returns
+    -------
+    Numpy arrays of model predictions in the format of model.outputs.
+
+    """
+    model = load_model(param_path, model_class=model_class)
+    dataset_files = model.test_files if test else model.validation_files
+    dataset = model.load_dataset_from_files(dataset_files)
+    indexed_dataset = dataset.to_indexed_dataset(model.data_indexer)
```
matt-gardner: This shouldn't be the way this works.  If we have to do it this way, we should change `model.score_dataset`.  I was thinking you could use `model.load_data_arrays()`, and that might work, except then you need a different scoring function that just calls `model.evaluate()` (or `model.evaluate_generator()`...).  Probably the right way to do this is just to put the `dataset.to_indexed_dataset` call inside of `model.score_dataset`, so it takes an un-indexed dataset as input.

deep_qa/training/trainer.py
```diff
@@ -31,7 +31,7 @@ class Trainer:
     The intended use of this class is that you construct a subclass that defines a model,
     overriding the abstract methods and (optionally) some of the protected methods in this class.
     Thus there are four kinds of methods in this class: (1) public methods, that are typically only
-    used by ``scripts/run_model.py`` (or some other driver that you create), (2) abstract methods
+    used by ``scripts/run.py`` (or some other driver that you create), (2) abstract methods
```
matt-gardner: Here's another place where you need to be consistent between `run_model.py` and `run.py`

deep_qa/run.py
```diff
@@ -0,0 +1,160 @@
...
+import sys
+import logging
+import shutil
+
+import random
+import pyhocon
+import numpy
+
+# pylint: disable=wrong-import-position
+from deep_qa.common.params import Params, replace_none, ConfigurationError
```
matt-gardner: This should be `from .common.params import ...`, and similarly on the next line.

deep_qa/run.py
DeNeutoy: Agree

DeNeutoy: Ah, slight issue here: `Trainer` doesn't define a `data_indexer`, `TextTrainer` does, so we can't index the dataset in `Trainer`. Are you happy with this going in `TextTrainer` instead? It feels a bit weird given we do the training etc  in `Trainer`...

matt-gardner: Yeah, `score_dataset` should be abstract in `Trainer`, and the implementation should be in `TextTrainer`.

Issue #303, opened 2017-04-20
Switch `run_on_aws.sh` to use ECR instead of bintray


matt-gardner:
See email from Jesse, which links to [this document](https://github.com/allenai/infrastructure/blob/master/docker/amazon-ecr.md).  This _should_ just mean changing a couple of lines in that script, but there might also be some other setup required.  In general, we should probably have a README somewhere with what the requirements are to use that script.

PR #337, opened 2017-05-04
Move to ECR for docker images

DeNeutoy:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM allenai-docker-private-docker.bintray.io/cuda:8
+FROM 896129387501.dkr.ecr.us-west-2.amazonaws.com/infrastructure/aristo/cuda:8
```
matt-gardner: Have you checked with aristo-eng or techops that this name is good to put in the repo?

matt-gardner: I'm guessing it's fine, but we should probably check, just to be sure.

scripts/run_on_aws.sh
```diff
@@ -13,10 +13,22 @@ fi
 
 set -e
 
-docker pull allenai-docker-private-docker.bintray.io/cuda:8
-
-docker build -t allenai-docker-private-docker.bintray.io/$CONTAINER_NAME . --build-arg PARAM_FILE=$PARAM_FILE
+# Get temporary ecr login.
+eval $(aws --region=us-west-2 ecr get-login)
+
+docker pull 896129387501.dkr.ecr.us-west-2.amazonaws.com/infrastructure/aristo/cuda:8
```
matt-gardner: This long crazy string is repeated a bunch of times; can you put it in a variable up above?

matt-gardner:
Fixes #303.

DeNeutoy:
I'm getting an error when running the actual script which isn't due to this change - I think it was to do with removing the paths when I moved run_model into deep_qa:
```
Traceback (most recent call last):
  File "scripts/run_model.py", line 4, in <module>
    from deep_qa import run_model
ImportError: No module named 'deep_qa'

```

matt-gardner:
Yeah, I'm about to fix that in another PR, that also adds a bit of functionality.  I'll submit it now.

Dockerfile
DeNeutoy: Yep it's fine.

PR #339, opened 2017-05-04
Making the DataGenerator more configurable

matt-gardner:


deep_qa/data/data_generator.py
```diff
@@ -86,45 +112,63 @@ def create_generator(self, dataset: IndexedDataset):
...
+                    groups = grouped_instances
+                for group in groups:
+                    batch = IndexedDataset(group)
+                    batch.pad_instances(self.text_trainer.get_padding_lengths(), verbose=False)
+                    yield batch.as_training_data()
         return generator()
 
+    def __create_batches(self, instances: List[IndexedInstance]) -> List[List[IndexedInstance]]:
+        for i in range(1, len(instances)):
+            if random.random() < self.permute_probability:
```
DeNeutoy: if `self.permute_probability == 1.` then `instances = instances[1:] + instances[0]`, right? This is a bit counter-intuitive as you would expect the ordering to become more random as `permute_probability` increases.

deep_qa/data/data_generator.py
```diff
@@ -46,6 +46,18 @@ class DataGenerator:
         certain padding values or not.  It's handled correctly for ``num_sentence_words`` and
         ``num_word_characters`` in :class:`~deep_qa.training.TextTrainer`, but you need to be sure
         to implement it correctly in subclasses for this to work.
+    permute_probability: double, optional (default=.2)
```
DeNeutoy: I'm not sure that this is really the best way to do this - imagine a dataset with binary labels where all sentences of length > 10 are positive and all sentences of length < 10 are negative (obviously an exaggerated case but the point holds). Using this method of shuffling only one instance with a negative label will cross over from being in a batch with all negative labels to being in a batch with all positive labels and even then this will only happen with probability `permute_probability`. 

What about doing something based off making the distribution of labels per batch more uniform past some threshold, say 60-70%? This is a hard problem to tackle.

deep_qa/data/data_generator.py
```diff
@@ -59,7 +71,18 @@ class DataGenerator:
...
+        without running out of memory.  Then (2) turn on ``adaptive_batch_sizes``, and set this
+        parameter so that you get the right batch size for your biggest instances.  If you set the
+        log level to ``DEBUG`` in ``scripts/run_model.py``, you can see the batch sizes that are
+        computed.
+    maximum_batch_size: int, optional (default=1000000)
+        If we're using adaptive batch sizes, you can use this to be sure you do not create batches
+        larger than this, even if you have enough memory to handle it on your GPU.  You might
+        choose to do this to keep smaller batches because you like the noisier gradient estimates
+        that come from smaller batches, for instance.
     biggest_batch_first: bool, optional (default=False)
```
DeNeutoy: As we start using this for different models, we should document good defaults for different GPUs in the models (just a random thought I had whilst reviewing, not really related).

deep_qa/data/data_generator.py
matt-gardner: You're right that if your data is adversarial, this will perform poorly.  I'm not too concerned about it in the typical case, though.

deep_qa/data/data_generator.py
matt-gardner: Yeah, I'm not sure how to document that, other than in the example files.  It definitely depends on the model you're using.

deep_qa/data/data_generator.py
matt-gardner: Yeah, that's what would happen if the probability is 1.  I was just trying to come up with a simple way to get a slight permutation of the sorted ordering.  Can you think of a better way to do it?

DeNeutoy: Maybe shuffling `self.permute_probabilities` percent of instances within a batch_size + eps window? 

```
for i in range(0, len(instances) - batch_size -1):
    indices_to_shuffle = [random.randint(0, batch_size) for _ in range(int(self.permute_probability * batch_size))]
    instances_to_shuffle = [(instances[x], x) for x in indices_to_shuffle]
    shuffled = shuffle(instances_to_shuffle)
    for (original, new) in (instances_to_shuffle, shuffled):
        instances[original] = new
```

matt-peters: Another option is to add a little noise to the padding lengths before sorting.  For example, if your average sentence length is 20 tokens, then adding random uniform [-2, 2] would effectively shuffle the batches a little on each epoch.

matt-gardner: Hmm, I like @matt-peters's suggestion - pretty simple to implement, doesn't change big-O runtime.  I'll do that.

deep_qa/common/util.py
```diff
@@ -15,3 +16,17 @@ def group_by_count(iterable: List[Any], count: int, default_value: Any) -> List[
     make a function out of it.
     """
     return [list(l) for l in zip_longest(*[iter(iterable)] * count, fillvalue=default_value)]
+
+
+def add_noise_to_dict_values(dictionary: Dict[Any, float], noise_param: float):
```
DeNeutoy: `-> Dict[Any, float]`

matt-gardner: Ah, thanks.

PR #341, opened 2017-05-04
Made TupleInferenceModel compatible with adaptive batch sizes

matt-gardner:
And cleaned up the code a bit while I was at it.


deep_qa/layers/tuple_matchers/word_token_tuple_matcher.py
```diff
@@ -0,0 +1,60 @@
...
+    This class gets two inputs of shape ``(batch_size, num_options, num_question_tuples,
+    num_background_tuples, num_slots, num_words)``, collapses them down to the right shape using
+    ``TimeDistributed`` (or equivalent) and passes them through the actual tuple matcher, returning
+    a tensor of tuple matches of shape ``(batch_size, num_options, num_question_tuples,
+    num_background_tuples)``.
+
+    Parameters
+    ----------
+    text_trainer: TextTrainer
+        A reference to the TextTrainer object this TupleMatcher belongs to, so we can call
```
DeNeutoy: ``TextTrainer`` and ``TupleMatcher``

deep_qa/layers/tuple_matchers/word_token_tuple_matcher.py
```diff
@@ -0,0 +1,60 @@
...
+                'ExpandFromBatch': ExpandFromBatch,
+                'Squeeze': Squeeze,
+                'WordOverlapTupleMatcher': WordOverlapTupleMatcher,
+                }
+
+
+
+# The first item added here will be used as the default in some cases.
+word_token_tuple_matchers = OrderedDict()  # pylint: disable=invalid-name
+word_token_tuple_matchers['word_overlap'] = WordOverlapTupleMatcher
```
DeNeutoy: Should the `WordTokenTupleMatcher` be in this dict?

deep_qa/layers/tuple_matchers/word_token_tuple_matcher.py
```diff
@@ -0,0 +1,60 @@
...
+        A reference to the TextTrainer object this TupleMatcher belongs to, so we can call
+        ``embed_input`` with it.
+
+    tuple_matcher_params: Dict[str, Any], default={}
+        Parameters for constructing an underlying TupleMatcher that operates on word token tuples.
+        We only read the "type" key here, which indexes a class in ``word_token_tuple_matchers``,
+        and then pass the rest of the parameters on to that class as ``kwargs``.
+    """
+    def __init__(self, text_trainer, tuple_matcher_params: Dict[str, Any]=None):
+        self.text_trainer = text_trainer
```
DeNeutoy: Where is this attribute used?

DeNeutoy: Ah is this so the api is the same for the other tuple matchers which do embed things?

deep_qa/layers/tuple_matchers/word_token_tuple_matcher.py
DeNeutoy: Seems like this happens in several of the tuple matchers - why?

deep_qa/layers/tuple_matchers/word_token_tuple_matcher.py
matt-gardner: Oh, yeah, I didn't even realize I had that copy-paste issue.  I'll add a comment above saying that this is for API consistency.

deep_qa/layers/tuple_matchers/word_token_tuple_matcher.py
matt-gardner: The way we did this is a little bit under-handed, and we should probably be more clear, or use better names, or something.  We essentially have two layers of nesting here: (1) what kind of input does your tuple matcher operate on, and (2) what exactly does the computation look like?  For (1), we have "word token", "embedded", and "encoded" input options, and then each of those has one or more concrete matching implementations.  So this dictionary is the second layer of nesting for the "word token" option.

DeNeutoy: Gotcha.

PR #342, opened 2017-05-04
Ecr docker v2

DeNeutoy:


DeNeutoy:
Switch to using tags rather than registries to distinguish between images.

matt-gardner:
Looks like you kept on working from your local branch, instead of creating a new branch off of master?

PR #345, opened 2017-05-05
Pinned pandas version

matt-gardner:
Newer version makes our build fail.

In general, I think I prefer pinning versions wherever possible, so that someone else upgrading their software doesn't break our code.  We can upgrade to a new pinned version when there's a compelling reason to do so.


matt-gardner:
Well, the issue with pinning versions is conflicting versions in transitive dependencies...  I guess the policy should be, if there's a version upgrade that breaks our build, we pin the version, then open an issue to remove the pin at some point?

Issue #306, opened 2017-04-20
Provide an easy model ensembling API


matt-gardner:
Not sure what this looks like, exactly, but it should be doable without much work.

matt-gardner:
This should just require adding a new class, `Ensembler`, and a script to run it.  The constructor takes a list of model config files (or param objects), loads all of the models, evaluates all of them on some dataset, and combines the results.  This should be pretty simple.

matt-gardner:
I'm starting on this now.

matt-gardner:
The initial implementation of this is in and working, but it could be much cleaner and more extensible.  That's a new, lower-priority issue, though, so I'm closing this one.

PR #349, opened 2017-05-08
Model ensembling

matt-gardner:
#306 

This provides a simple way to do model ensembling, just loading a list of saved models and averaging their predictions.

I also updated some docs a bit, so there are also some somewhat unrelated changes in this PR.

I'm planning on writing a simple test for this, that running an ensemble with 1 item in it returns the same result as `score_dataset` by itself, but the basic design is ready to be reviewed, I think.


deep_qa/run.py
```diff
@@ -183,3 +188,62 @@ def evaluate_model(param_path: str, dataset_files: List[str]=None, model_class=N
...
+        logger.info("Scoring model %d of %d", i + 1, len(models))
+        dataset = model.load_dataset_from_files(dataset_files)
+        model_predictions, labels = model.score_dataset(dataset)
+        predictions.append(model_predictions)
+        if labels_to_return is None:
+            labels_to_return = labels
+    logger.info("Averaging model predictions")
+    all_predictions = numpy.stack(predictions)
+    averaged = numpy.mean(all_predictions, axis=0)
+    return averaged, labels_to_return
```
DeNeutoy: Maybe return the individual model predictions as well? I could imagine a situation where you might want the actual values, such as to compute the variance of predictions or something, or if the models you are evaluating are actually designed to deal with different types of inputs.

deep_qa/training/text_trainer.py
```diff
@@ -168,6 +168,39 @@ def create_data_arrays(self, dataset: IndexedDataset):
...
+        """
+        # TODO(matt): for some reason the reference to the super class docs above isn't getting
+        # linked properly.  I'm guessing it's because of an indexing issue in sphinx, but I
+        # couldn't figure it out.  Once that works, it can be changed to "See :func:`the superclass
+        # docs <Trainer.score_dataset>` for usage info").
+        # TODO(matt): it _should_ be as easy as iterating over the data again to pull out the
+        # labels, so we can still use data generators, but I'm waiting on implementing that.
+        indexed_dataset = dataset.to_indexed_dataset(self.data_indexer)
+        data_generator = self.data_generator
+        self.data_generator = None
```
DeNeutoy: Add a comment here saying that this links up to the bit where you say we aren't using data generators, as it's not super obvious this is what it does atm.

deep_qa/run.py
```diff
@@ -150,13 +151,17 @@ def score_dataset(param_path: str, test: str=False, model_class=None):
 
     Returns
     -------
-    Numpy arrays of model predictions in the format of model.outputs.
-
+    predictions: numpy.array
+        Numpy array of model predictions in the format of model.outputs (typically one array, but
+        could be List[numpy.array] if your model has multiple outputs).
+    dataset: Dataset
```
matt-gardner: Because of how sphinx / numpydoc processes parameters, you actually don't put double backticks around parameter types.  See http://deep-qa.readthedocs.io/en/latest/run.html.

deep_qa/run.py
matt-gardner: I'm thinking that actually this should be an `Ensembler` class, so that you can easily have options on _how_ to do the ensembling, and you can specify a parameter file for the ensemble, etc.  That should make a lot of things easier down the road, when we're integrating with the Aristo Lab stuff that aristo-eng is doing.  So I'm going to wait on this suggestion until we refactor this for a better ensembling API.

deep_qa/training/text_trainer.py
matt-gardner: Done.

deep_qa/run.py
DeNeutoy: Great that sounds good.

deep_qa/run.py
DeNeutoy: Ah, my bad.

PR #353, opened 2017-05-11
Ranking loss in sentence selection

matt-gardner:
This isn't working yet; I'm running into a funny error with actually using this loss function.  Once I figure out that error, though, this should be good to use.


matt-gardner:
Turns out the error was because tensorflow couldn't compute a gradient for variables when there was a `switch` in the loss function.  This is actually kind of troubling, because it might mean all of our models and layers that use `switch` might be broken...  Replacing the `switch` with an addition fixed things, though, so this is now ready to review.

matt-gardner:
@haniesedghi, FYI.

deep_qa/layers/attention/attention.py
```diff
@@ -28,17 +27,22 @@ class Attention(MaskedLayer):
...
-    - attention: shape ``(batch_size, num_rows)``, no mask (masked input rows have value 0 in the
-      output)
+    - attention: shape ``(batch_size, num_rows)``.  If ``normalize`` is ``True``, we return no
+      mask, as we've already applied it (masked input rows have value 0 in the output).  If
+      ``normalize`` is ``False``, we return the matrix mask, if there was one.
 
     Parameters
     ----------
-    similarity_function_params: Dict[str, Any], optional (default={})
+    similarity_function_params : ``Dict[str, Any]``, optional (default: ``{}``)
```
matt-gardner: @DeNeutoy, I noticed that I was actually wrong earlier about the correct formatting of these things.  The backticks doing the wrong thing was because there needs to be a space before the colon. TIL...

You can see a working example [here](http://deep-qa.readthedocs.io/en/latest/models/reading_comprehension.html#module-deep_qa.models.reading_comprehension.bidirectional_attention).

PR #356, opened 2017-05-15
Multi gpu 2

DeNeutoy:


DeNeutoy:
This is just a clean version of the other PR without the experimental bits and leaving out some irrelevant keras.compile refactoring.

DeNeutoy:
@matt-gardner I have added a ModelCheckpoint class for the multi-gpu models which will only save the actual `DeepQaModel` layer. In terms of model loading, what do you think the default behavior should be? presumably when you are loading a model, you are doing some sort of evaluation so you probably don't want it to be parallelised? What do you think the default should be?

matt-gardner:
Yeah, load a model onto a single GPU (or CPU).  If we need to get more complicated than that later, we'll worry about it then.

matt-gardner:
Not sure what the deal is with the sphinx warning; you might want to try building the docs locally to see if it helps: `cd doc; make html-strict`.

deep_qa/training/callbacks.py
```diff
@@ -0,0 +1,68 @@
...
+        num_lambda_layers = len(self.model.outputs)
+
+        self.epochs_since_last_save += 1
+        if self.epochs_since_last_save >= self.period:
+            self.epochs_since_last_save = 0
+            filepath = self.filepath.format(epoch=epoch, **logs)
+            if self.save_best_only:
+                current = logs.get(self.monitor)
+                if current is None:
+                    warnings.warn('Can save best model only with %s available, '
```
matt-gardner: Why not use `logger.warn`?

deep_qa/training/callbacks.py
```diff
@@ -0,0 +1,68 @@
...
+                                  'skipping.' % (self.monitor), RuntimeWarning)
+                else:
+                    if self.monitor_op(current, self.best):
+                        if self.verbose > 0:
+                            print('Epoch %05d: %s improved from %0.5f to %0.5f,'
```
matt-gardner: These print statements too; I'd probably just change them to `logger.debug`.

deep_qa/training/multi_gpu.py
```diff
@@ -0,0 +1,89 @@
+from keras.layers import merge
+from keras.layers.core import Lambda
+import keras.backend as K
+import tensorflow
+from .models import DeepQaModel
+
+
+def pin_variable_device_scope(device, parameter_device="/cpu:0"):
+
```
matt-gardner: There's an extra blank line here.

deep_qa/training/multi_gpu.py
```diff
@@ -0,0 +1,89 @@
+from keras.layers import merge
+from keras.layers.core import Lambda
+import keras.backend as K
+import tensorflow
+from .models import DeepQaModel
```
matt-gardner: I think having a blank line separating import groups is helpful (one block for standard python imports, one block for third-party imports, and one block for first-party imports).

deep_qa/training/multi_gpu.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def pin_variable_device_scope(device, parameter_device="/cpu:0"):
```
matt-gardner: You use `variable` in the method name, but `parameter` in the argument name.  These should be consistent.

deep_qa/training/multi_gpu.py
```diff
@@ -0,0 +1,89 @@
...
+        node_def = graph_op if isinstance(graph_op, tensorflow.NodeDef) else graph_op.node_def
+        if node_def.op in ['Variable', 'VariableV2']:
+            return parameter_device
+        else:
+            return device
+    return _assign
+
+
+def make_parallel(model: DeepQaModel, gpu_count: int) -> DeepQaModel:
+
```
matt-gardner: There's an extra blank line here too.

deep_qa/training/multi_gpu.py
```diff
@@ -0,0 +1,89 @@
...
+                                         arguments={'index': gpu_index, 'parts': gpu_count})
+                    slice_n = slice_layer(model_input)
+                    inputs.append(slice_n)
+
+                outputs = model(inputs)
+                if not isinstance(outputs, list):
+                    outputs = [outputs]
+
+                # Save all the outputs for merging back together later
+                for i in range(len(outputs)):  # pylint: disable=consider-using-enumerate
```
matt-gardner:
```python
for i, output in enumerate(outputs):
    all_outputs[i].append(output)
```

deep_qa/training/multi_gpu.py
```diff
@@ -0,0 +1,89 @@
...
+                    all_outputs[i].append(outputs[i])
+
+    # merge outputs on CPU.
+    with tensorflow.device('/cpu:0'):
+        merged = []
+        for outputs in all_outputs:
+            merged.append(merge(outputs, mode='concat', concat_axis=0))
```
matt-gardner: `merge` is deprecated in keras 2.  You should just use `Concatenate`.

deep_qa/training/trainer.py
```diff
@@ -3,16 +3,18 @@
 from typing import Any, Dict, List, Tuple
 
 import numpy
+import tensorflow
 from keras.models import model_from_json
 from keras.callbacks import LambdaCallback, TensorBoard, EarlyStopping, CallbackList, ModelCheckpoint
-
+from ..training.callbacks import ReplicaModelCheckpoint
```
matt-gardner: Can you sort these?

deep_qa/training/trainer.py
```diff
@@ -148,8 +150,15 @@ def __init__(self, params: Params):
             os.makedirs(parent_directory, exist_ok=True)
 
         # `model.fit()` parameters.
+        self.num_gpus = params.pop("num_gpus", 1)
```
matt-gardner: This needs to be added to the docstring, with some usage info (like that we will just use batch parallelism, and that it's not currently compatible with adaptive batch sizes, and increases padding lengths with dynamic padding).

deep_qa/training/trainer.py
```diff
@@ -282,6 +291,10 @@ def train(self):
         # Then we build the model and compile it.
         logger.info("Building the model")
         self.model = self._build_model()
+        if self.num_gpus > 1:
+            with tensorflow.device("/cpu:0"):
```
matt-gardner: Why is this necessary?  The device scopes you have inside of `make_parallel` aren't enough?

deep_qa/training/trainer.py
```diff
@@ -591,7 +606,11 @@ def _save_auxiliary_files(self):
         Called after training. If you have some auxiliary object, such as an object storing
         the vocabulary of your model, you can save it here. The model config is saved by default.
         """
-        model_config = self.model.to_json()
+        if self.num_gpus > 1:
+            lambda_layers = len(self.model.outputs)
```
matt-gardner: `num_lambda_layers` - when I see a variable named `lambda_layers`, I think it's a list.

tests/training/multi_gpu_test.py
```diff
@@ -0,0 +1,68 @@
...
+import keras.backend as K
+import tensorflow
+from deep_qa.training.multi_gpu import pin_variable_device_scope
+from deep_qa.common.params import Params
+from deep_qa.models.text_classification import ClassificationModel
+from ..common.test_case import DeepQaTestCase
+
+
+class TestMultiGpu(DeepQaTestCase):
+    # pylint: disable=protected-access
```
matt-gardner: I don't think you need this.

tests/training/multi_gpu_test.py
```diff
@@ -0,0 +1,68 @@
+# pylint: disable=no-self-use,invalid-name
+
+
+from copy import deepcopy
```
matt-gardner: Separating these imports into the three standard blocks would be nice.  And I think zero or one blank line between the top comment and the first import is enough.

tests/training/multi_gpu_test.py
```diff
@@ -0,0 +1,68 @@
...
+                'num_gpus': 2,
+                'save_models': True,
+                'show_summary_with_masking_info': True,
+        })
+
+    def test_model_can_train_and_load(self):
+        self.ensure_model_trains_and_loads(ClassificationModel, self.args)
+
+    def test_pinned_scope_correctly_allocates_ops(self):
+
```
matt-gardner: These blank lines at the beginning of methods aren't necessary.

tests/training/multi_gpu_test.py
```diff
@@ -0,0 +1,68 @@
...
+        self.args = Params({
+                'embedding_dim': {'words': 4, 'characters': 2},
+                'batch_size': 8,
+                'num_gpus': 2,
+                'save_models': True,
+                'show_summary_with_masking_info': True,
+        })
+
+    def test_model_can_train_and_load(self):
+        self.ensure_model_trains_and_loads(ClassificationModel, self.args)
```
matt-gardner: What does this do when run on a machine with no gpu?

tests/training/multi_gpu_test.py
```diff
@@ -0,0 +1,68 @@
...
+            # slightly differently from their scopes ... (i.e != "/cpu:0")
+            assert variable.device == "/device:CPU:0" or variable.device == ""
+
+    def test_multi_gpu_shares_variables(self):
+
+        multi_gpu_model = self.get_model(ClassificationModel, self.args)
+
+        single_gpu_args = deepcopy(self.args)
+        single_gpu_args["num_gpus"] = 1
+        single_gpu_model = self.get_model(ClassificationModel, self.args)
```
matt-gardner: Did you mean to pass in `single_gpu_args` here?

tests/training/multi_gpu_test.py
```diff
@@ -0,0 +1,68 @@
...
+
+        multi_gpu_model.train()
+        multi_gpu_variables = [x.name for x in multi_gpu_model.model.trainable_weights]
+
+        K.clear_session()
+        single_gpu_model.train()
+        single_gpu_variables = [x.name for x in single_gpu_model.model.trainable_weights]
+
+        assert single_gpu_variables == multi_gpu_variables
```
matt-gardner: If you weren't sharing variables, how would this fail?  There would be more variables in the multi-gpu case, because Keras had allocated more of them?

DeNeutoy: Yeah if the scopes weren't working, there would be a new copy of the variables, eg variable, variable_1 ...

tests/training/multi_gpu_test.py
DeNeutoy: Yes! Thanks.

tests/training/multi_gpu_test.py
DeNeutoy: If you set `soft_device_placement=True` in the tensorflow session, tf won't complain if it can't allocate things exactly on the device you specify, so in this case, we'd be building 2 models on the cpu. If we remove this option, it would crash. I think generally it's recommended to allow tensorflow this flexibility, as it does some optimisiation in terms of device placements, like putting embedding lookups on the same device as the embedding etc.

deep_qa/training/trainer.py
DeNeutoy: Yeah this was from when I was playing around with using a global cpu scope before I realised I could use that conditional scope thing - I am not sure if we need to wrap when we actually call `build_model`, just in case some of our layers are actually build when we call this (rather than being built in model.compile). Regardless, this is in the wrong place even if we do need to do this.

DeNeutoy:
@matt-peters, I think you no longer need to use your Keras fork to do multi-gpu stuff (with some caveats). Please let me know if you don't agree with the below as it effects this PR:

This (line in this commit)[https://github.com/fchollet/keras/commit/82f9e2358c35c5bbff8714280174617ba29cf92b#diff-fe3ca734e8421a17729d0b0ea4c800baR445] added name scoping into building keras layers(although not to address this problem, just to make better layer scopes for tensorboard).

Now, this shouldn't change anything for variable creation, because tf.name_scope only scopes non-variable ops. Right? Wrong - it effects all ops with the exception of calls to `tf.get_variable`, but includes `tf.Variable()`, which is what keras uses in all of it's backend tensorflow variable creation methods. 
example:
```
with tf.name_scope("test"):
    var1 = tf.Variable([], name="var1")

assert var1.name == 'test/test_var:0' # True

with tf.name_scope("test"):
    var2 = tf.get_variable("var2", [])

assert var2.name == "var2:0"
```
This means that if we do something like: 

```
model = KerasModel(inputs, outputs)
outputs = []
for x in range(3):
with tf.device("/gpu:{}".format(x)):
     outputs[x] = model(inputs)

new_model = KerasModel(inputs, concat(outputs))
model.compile(new_model)
```
Now we are reusing the model as a layer, it will have it's own scope when it is built and hence the layers will be reused, rather than built again - the same argument as why to create a shared lstm or something in Keras, you would do:
```
lstm = LSTM(64)

# Shared weights
lstm(input1)
lstm(input2)
```

Do you agree?

matt-peters:
In your second snippet, I agree that this will re-use the variables since the layers won't be rebuilt, in the same way that the third snippet will re-use them.  One sanity check that I do is printing the results of `tf.global_variables()` after the final model is created.  This will quickly show if there are duplicated variables that should be re-used.

The thing that I'm not sure about with this approach is how tensorflow will handle moving ops and variables from device to device and how that impacts efficiency.  The only way I know to resolve this is to benchmark - if you run on say 2 GPUs with reasonably large batch size (so that computing gradients is slow relative to updating them) what is the speedup vs 1 GPU?  And 4 GPUs vs 1 GPUs?

DeNeutoy:
Ok great, I have a test which does exactly the `global_variables` thing, I was just checking that you also thought it was sufficient. 

With regard to the efficiency thing, I think I have fixed that by pinning all the variables to the cpu using this [scope function](https://github.com/allenai/deep_qa/pull/356/files#diff-e32b8ddb852508eafd1b287df0bcb4a1R9), which allocates new variables to the cpu even if they are created within a gpu scope. When I tested this, 2 and 3 gpus were no slower than using 1.

matt-peters:
I'm confused by your last statement about efficiency - with linear scaling, 2 GPUs should be twice as fast as 1 GPU (for the same equivalent batch size), not about the same speed?

DeNeutoy:
Oh I see - I think I must have a grip of how people talk about scaling wrong because Matt G also was confused by my measurements - basically 2 GPUs with 32 batch size on each (64 effective batch size) is as fast as 1 GPU with a 32 batch size, and the same with 3GPUs (96 effective batch size). Is that what you meant?

matt-gardner:
Yeah, I'd call that linear scaling.  You're right that it's not necessarily a linear speedup in learning time, because that interacts in interesting ways with the number of steps you take, and whatnot.  But it's a linear speedup in how fast you can go through the data.

matt-peters:
Ah yes, that's what I'd call linear scaling.  Just to confirm, your timings are per batch, right (not per epoch)?  We should expect the epoch time to decrease by half with 2 GPUs and 2x effective batch size.

Sounds like this is as efficient as we could hope for then, nice work!

DeNeutoy:
Ah, I need to run these again as the comparisons were with an approximate batch size. Thanks for pointing that out!

DeNeutoy:
Ok this isn't working properly, the time for a full epoch over the training data for 2 GPUs is only 50secs faster with 2 GPUs, I was thinking about the measurements wrong. I think the problem is the gradient computation - it needs to per gpu, with just the aggregation happening on the cpu. So frustrating.

DeNeutoy:
Merging this, with the objective of adding more of tower functionality for the gradients to see if that speeds things up. @matt-peters do you have benchmarks for how fast this should be? What speedup did you see from running the multi-gpu stuff for your language models?

matt-peters:
After running some benchmarks on my code, I'm thinking that this approach is only useful for taking larger batches then can fit on a GPU.  For an SNLI model with LM (high ratio of computation to number of parameters) I'm seeing approximately linear speedup with number of GPUs.  However, I can also get substantial speedup on a single GPU by just increasing batch size until it runs OOM.

Here are some benchmarks:

```
1 GPU, batch_size = 32, 24.6 examples / s
1 GPU, batch_size = 64, 45.7 examples / s
1 GPU, batch_size = 128: OOM

2 GPUs, batch_size = 32 x 2, 46.6 examples / s
2 GPUS, batch_size = 64 x 2, 86.9 examples / s

4 GPUs, batch_size = 64 x 4, 174.6 examples / s
```

matt-peters:
Relevant paper:

https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/45187.pdf

PR #357, opened 2017-05-15
Update sentence selection model to use dynamic padding

matt-gardner:
Currently getting a weird error on model loading, but after that's fixed, this will hopefully just work as-is.


matt-gardner:
This is now ready to review.  It's updating the sentence selection model to be compatible with the new dynamic padding stuff that I've been working on.  This should make training these models a lot faster.  @haniesedghi, FYI.

deep_qa/layers/wrappers/add_encoder_mask.py
```diff
@@ -0,0 +1,55 @@
...
+    you want to wrap an encoder using :class:`~..backend.CollapseToBatch` and
+    :class:`~..backend.ExpandFromBatch`, you need a way to get the mask back into the right form
+    after running your encoder.  This is an issue because Keras' encoders don't return masks when
+    they output single vectors.
+
+    For example, say you have a list of sentences, like [[5, 2, 1, 0], [2, 3, 1, 1], [0, 0, 0, 0]]
+    (using word indices instead of embeddings for simplicity), which has been padded to be three
+    sentences, even though only two of them are actually used.  After passing it though an encoder,
+    you'll have something like [[vector], [vector], [vector]], and you want a mask that looks like
+    [1, 1, 0].  Keras' LSTMs and such won't give this to you.  This method adds it back.
```
pdasigi: You may want to say that you are assuming the encoder does not return sequences.

deep_qa/layers/wrappers/add_encoder_mask.py
```diff
@@ -0,0 +1,55 @@
...
+          to that encoder.  The mask for this tensor must have shape ``(batch_size, ...,
+          num_words)``.
+
+    Output:
+        - The first input tensor, with a mask computed from the second input tensor.  The
+          computation is just ``K.any()`` on the last dimension.
+    """
+    def __init__(self, **kwargs):
+        # The base layer class overwrites self.supports_masking,
+        # so we add it afterwards.
```
pdasigi: Not sure what you meant by this comment. Did you mean to change the `supports_masking` field?

tests/layers/wrappers/add_encoder_mask_test.py
```diff
@@ -0,0 +1,37 @@
...
+                        [0, 1, 2],
+                        [1, 0, 0],
+                        [0, 0, 0],
+                        [0, 1, 0],
+                        [1, 1, 1],
+                ]
+        ])
+        expected_mask = numpy.asarray([[0, 1, 0, 1, 1, 0, 1, 1]])
+        actual_mask = model.predict([test_background])
+        numpy.testing.assert_array_almost_equal(expected_mask, actual_mask)
```
pdasigi: Why not `assert_array_equal`?

deep_qa/layers/wrappers/add_encoder_mask.py
matt-gardner: I updated the docstring in a couple of places to hopefully make this more clear.

deep_qa/layers/wrappers/add_encoder_mask.py
matt-gardner: Thanks for catching that.  It's a copy-paste comment that was still in `OutputMask` because I forgot to remove it when I made it inherit from `MaskedLayer`.  Removed here and in `OutputMask`.

tests/layers/wrappers/add_encoder_mask_test.py
matt-gardner: Yeah, that's a good point.  These are boolean tensors, so equal should work.

matt-gardner:
Ok, I fixed those comments, @pdasigi, and I've run some timing benchmarks.  Without dynamic padding, I was truncating to 40 words per sentence.  The max sentence length is ~260 words, though I don't know the average sentence length.  One epoch with this truncation took about 45 minutes.  When I switched to dynamic padding I removed all truncation.  It was on track to finish in about 1.25 hours per epoch, but it ran out of memory on a batch with really long sentences.

If we assume a linear relationship between sentence length and runtime (which may not be accurate), we'd get maybe ~3 hours per epoch without any truncation (though it also would run out of memory, so we'd have to decrease the batch size, which would make it slower).  Dynamic padding, then, gives us a ~2-3x speedup in this model.

PR #358, opened 2017-05-15
Added Verb semantics instance and corresponding test cases

bhavanadalvi:


deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
```diff
@@ -0,0 +1,173 @@
...
+from overrides import overrides
+
+from ..instance import TextInstance, IndexedInstance
+from ...data_indexer import DataIndexer
+
+
+class VerbSemanticsInstance(TextInstance):
+    """
+    A VerbSemanticsInstance is a :class:`TextInstance` that is a single sentence of text,
+    where that sentence has some associated categorical label.
```
DeNeutoy: Perhaps add a few more descriptions of the verb, entity, arg1, arg2 and state change inputs - that would help someone work out the format of this instance.

deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
```diff
@@ -0,0 +1,173 @@
...
+    """
+    def __init__(self, sentence: List[str], verb: (int, int), entity: (int, int), state_change: str, arg1: (int, int), arg2: (int, int), index: int=None):
+        """
+        sentence: a sentence from which we want to predict verb semantics.
+        A sentence can mention multiple verbs and entities, a data instance
+        is talking about a particular (verb, entity) pair in a given sentence
+        verb: span of the verb indicating the event being referred to in the query
+        entity: entity indicating 
+        """
+        """ TODO: initialize all members of the class"""
```
DeNeutoy: I think you've done this TODO already, just remove it 😄

deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
```diff
@@ -0,0 +1,173 @@
...
+    A VerbSemanticsInstance is a :class:`TextInstance` that is a single sentence of text,
+    where that sentence has some associated categorical label.
+    """
+    def __init__(self, sentence: List[str], verb: (int, int), entity: (int, int), state_change: str, arg1: (int, int), arg2: (int, int), index: int=None):
+        """
+        sentence: a sentence from which we want to predict verb semantics.
+        A sentence can mention multiple verbs and entities, a data instance
+        is talking about a particular (verb, entity) pair in a given sentence
+        verb: span of the verb indicating the event being referred to in the query
+        entity: entity indicating
```
DeNeutoy: Finish/add descriptions for the rest of the arguments.

deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
```diff
@@ -0,0 +1,173 @@
...
+        Convert (arg1, arg2) into a tag sequence over the sentence
+        :param arg1: startIndex,endIndex of arg1 in the sentence 
+        :param arg2: startIndex,endIndex of arg2 in the sentence
+        :return: tag sequence
+        E.g. sentence = "Roots absorb water from the soil"
+            arg1 = (4,5)  // the soil
+            arg2 = (0,0)  // roots
+            tag sequence = (B-ARG2, O, O, O, B-ARG1, I-ARG1)
+        """
+        tags = ["O"] * self.sentence.__len__()
```
DeNeutoy: Generally you should use `len(self.sentence)` as `__x__()` methods are designed to be called by Python, rather than the user(in general).

deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
```diff
@@ -0,0 +1,173 @@
...
+        if arg1[0] != -1:
+            tags[arg1[0]] = "B-" + "ARG1"
+            for index1 in range(arg1[0]+1, arg1[1]):
+                tags[arg1[0]] = "I-" + "ARG1"
+        if arg2[0] != -1:
```
DeNeutoy: I presume that it's ok for both arg1 and arg2 not to be present? If not, you should catch this here.

deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
```diff
@@ -0,0 +1,173 @@
...
+        """
+        words = self._words_from_text(self.sentence)
+        words['state_changes'] = [self.label[0]]
+        words['tags'] = self.label[1]
+        return words
+
+    @overrides
+    def to_indexed_instance(self, data_indexer: DataIndexer):
+        sentence_indices = self._index_text(self.sentence, data_indexer)
+        verb_indices = sentence_indices[self.verb[0]:self.verb[1]]
```
DeNeutoy: You could add a comment here that the verb and entity indices are just spans within the sentence indices.

tests/data/instances/sequence_tagging/verb_semantics_instance_test.py
```diff
@@ -0,0 +1,57 @@
...
+from deep_qa.data.instances.instance import TextInstance
+from deep_qa.data.instances.sequence_tagging.pretokenized_tagging_instance import PreTokenizedTaggingInstance
+from deep_qa.data.instances.sequence_tagging.verb_semantics_instance import VerbSemanticsInstance
+from deep_qa.data.tokenizers import tokenizers
+from deep_qa.common.params import Params
+from ....common.test_case import DeepQaTestCase
+
+
+class TestVerbSemanticsInstance(DeepQaTestCase):
+    def setUp(self):
```
DeNeutoy: If you aren't adding any functionality to a superclass, there's no need to implement it, so you can just remove setUp and tearDown - if you look in DeepQaTestCase they are already implemented.

tests/data/instances/sequence_tagging/verb_semantics_instance_test.py
```diff
@@ -0,0 +1,57 @@
...
+        verb = (1, 1)
+        entity = (2, 2)
+        state_change = "MOVE"
+        tags = ["B-ARG2", "O", "O"]
+
+        assert instance.sentence == sentence
+        assert instance.verb == verb
+        assert instance.entity == entity
+        assert instance.label[0] == state_change
+        assert instance.label[1] == tags
```
DeNeutoy: Add an empty new line at the end of the file.

deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
```diff
@@ -0,0 +1,173 @@
...
+        verb: startIndex,endIndex of verb in the sentence
+        entity: startIndex,endIndex of entity in the sentence
+        state_change_label: one label from {CREATE, DESTROY, MOVE}
+        arg1: startIndex,endIndex of arg1 in the sentence
+        arg2: startIndex,endIndex of arg2 in the sentence
+        """
+        fields = line.split("\t")
+
+        if len(fields) == 6:
+            sentenceStr, verb_str, entity_str, state_change_label, arg1_str, arg2_str = fields
```
DeNeutoy: Change these variable names to be sentence_string, verb_string. Use full names in variables so that they are easy to read.

deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
```diff
@@ -0,0 +1,173 @@
...
+            sentence = sentenceStr.split("####")
+            v_parts = verb_str.split(",")
```
DeNeutoy: same for these - use `verb_parts` instead of `v_parts`, etc.

tests/data/instances/sequence_tagging/verb_semantics_instance_test.py
```diff
@@ -0,0 +1,57 @@
+# pylint: disable=no-self-use,invalid-name
+from typing import List
```
DeNeutoy: Some of the imports below are unused - see the pylint tests here: https://travis-ci.org/allenai/deep_qa/jobs/232628802

Just remove the ones you don't need.

You can run the pylint tests locally by running: 
`pylint -d locally-disabled -f colorized DIRECTORY`

where DIRECTORY is the directory which you want pylint to check. You should run this on deep_qa/ and tests/ as you have added code to both of these - it should help to get the formatting right.

tests/data/instances/sequence_tagging/verb_semantics_instance_test.py
```diff
@@ -0,0 +1,57 @@
...
+from deep_qa.data.data_indexer import DataIndexer
+from deep_qa.data.instances.instance import TextInstance
+from deep_qa.data.instances.sequence_tagging.pretokenized_tagging_instance import PreTokenizedTaggingInstance
+from deep_qa.data.instances.sequence_tagging.verb_semantics_instance import VerbSemanticsInstance
+from deep_qa.data.tokenizers import tokenizers
+from deep_qa.common.params import Params
+from ....common.test_case import DeepQaTestCase
+
+
+class TestVerbSemanticsInstance(DeepQaTestCase):
```
DeNeutoy: You might want to test that when you call `to_indexed_instance`, everything functions as you expect - you can have a look at an example of that here: 
https://github.com/allenai/deep_qa/blob/master/tests/data/instances/sequence_tagging/test_tagging_instance.py#L24

deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
bhavanadalvi: yes it is fine for both arguments to be empty.

tests/data/instances/sequence_tagging/verb_semantics_instance_test.py
bhavanadalvi: Thanks, this is a cool tool, I started using it.

deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
```diff
@@ -0,0 +1,173 @@
...
+from ..instance import TextInstance, IndexedInstance
+from ...data_indexer import DataIndexer
+
+
+class VerbSemanticsInstance(TextInstance):
+    """
+    A VerbSemanticsInstance is a :class:`TextInstance` that is a single sentence of text,
+    where that sentence has some associated categorical label.
+    """
+    def __init__(self, sentence: List[str], verb: (int, int), entity: (int, int), state_change: str, arg1: (int, int), arg2: (int, int), index: int=None):
```
matt-gardner: Hmm, I didn't think that `(int, int)` was a valid type declaration, but pylint doesn't appear to be complaining about it.  That's good to know.

deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
```diff
@@ -0,0 +1,173 @@
...
+        """ TODO: initialize all members of the class"""
+        self.sentence = sentence
+        self.verb = verb
+        self.entity = entity
+        tags = self.__make_tag_sequence(arg1, arg2)
+        super(VerbSemanticsInstance, self).__init__((state_change, tags), index)
+
+
+    def __str__(self):
+        return 'VerbSemanticsInstance(' + self.sentence + ', ' + str(self.label) + ')'
```
matt-gardner: Not a huge deal, but this is for debug output.  You might want to also include the verb and entity in here somehow, too.

deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def __make_tag_sequence(self, arg1: (int, int), arg2: (int, int)) -> List[str]:
+        """
+        Convert (arg1, arg2) into a tag sequence over the sentence
+        :param arg1: startIndex,endIndex of arg1 in the sentence
```
matt-gardner: Again here, because we use sphinx + numpydoc to generate our documentation, this should be formatted slightly differently, like this: https://github.com/allenai/deep_qa/blob/6a0039822f9f2c727151b6143e2214d331bdc111/deep_qa/tensors/backend.py#L153-L167.

deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
matt-gardner: And make sure you note that the sentence is _pretokenized_.

deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
matt-gardner: You probably don't want to do this this way.  Probably you want a binary vector that's the length of the sentence, which can be concatenated with the word embedding.  So you'd compute this very similarly to how you compute the tag sequence, starting with a 0 vector and replacing the span with 1s.

deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
```diff
@@ -0,0 +1,173 @@
...
+    def to_indexed_instance(self, data_indexer: DataIndexer):
+        sentence_indices = self._index_text(self.sentence, data_indexer)
+        verb_indices = sentence_indices[self.verb[0]:self.verb[1]]
+        entity_indices = sentence_indices[self.entity[0]:self.entity[1]]
+
+        indexed_label = self._index_label()
+        return IndexedVerbSemanticsInstance(sentence_indices, verb_indices,
+                                            entity_indices, indexed_label, self.index)
+
+    def _index_label(self, data_indexer: DataIndexer) -> (int, List[int]):
```
matt-gardner: This is a separate method in `TaggingInstance` so that it can be easily overridden.  There isn't a whole lot of reason to make it a separate method here; it'd probably be fine to just move this code into `to_indexed_instance` directly.

deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
```diff
@@ -0,0 +1,173 @@
...
+            sentence = sentenceStr.split("####")
+            v_parts = verb_str.split(",")
+            verb = (int(v_parts[0]), int(v_parts[1]))
+            e_parts = entity_str.split(",")
+            entity = (int(e_parts[0]), int(e_parts[1]))
+            a1_parts = arg1_str.split(",")
+            arg1 = (int(a1_parts[0]), int(a1_parts[1]))
+            a2_parts = arg2_str.split(",")
+            arg2 = (int(a2_parts[0]), int(a2_parts[1]))
+            return cls(sentence, verb, entity, state_change_label, arg1, arg2)
```
matt-gardner: It looks like this logic will crash if `arg1_str` or `arg2_str` is empty, which I'm pretty sure you expect to happen.

deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
```diff
@@ -0,0 +1,173 @@
...
+    @overrides
+    def get_padding_lengths(self) -> Dict[str, int]:
+        """
+        This simple IndexedInstance only has one padding dimension: word_indices.
+        """
+        return self._get_word_sequence_lengths(self.sentence)
+
+    @overrides
+    def pad(self, padding_lengths: Dict[str, int]):
+        self.sentence = self.pad_word_sequence(self.sentence, padding_lengths)
```
matt-gardner: If you represent the verb and entity the way that I suggested above, you'll also need to pad them correctly here, and you need to pad the label.  You'd do all of those with something like [this](https://github.com/allenai/deep_qa/blob/6a0039822f9f2c727151b6143e2214d331bdc111/deep_qa/data/instances/sequence_tagging/tagging_instance.py#L74-L77).

tests/data/instances/sequence_tagging/verb_semantics_instance_test.py
matt-gardner: You can also run `./scripts/pylint.sh` to run it on all of the code at once, instead of running it separately for `deep_qa/` and `tests/`.

tests/data/instances/sequence_tagging/verb_semantics_instance_test.py
```diff
@@ -0,0 +1,57 @@
...
+    def setUp(self):
+        super(TestVerbSemanticsInstance, self).setUp()
+
+    def tearDown(self):
+        super(TestVerbSemanticsInstance, self).tearDown()
+
+    def test_read_from_arguments(self):
+        """sentence: List[str], verb: (int, int), entity: (int, int), state_change: str, arg1: (int, int), arg2: (int, int)"""
```
matt-gardner: I think there's not really a need to have a docstring on a test method like this.  We can see the format from looking at the instance class docstrings.

tests/data/instances/sequence_tagging/verb_semantics_instance_test.py
```diff
@@ -0,0 +1,57 @@
...
+        tags = ["B-ARG2", "O", "O"]
+
+        assert instance.sentence == sentence
+        assert instance.verb == verb
+        assert instance.entity == entity
+        assert instance.label[0] == state_change
+        assert instance.label[1] == tags
+
+    def test_read_from_line(self):
+        """sentence: List[str], verb: (int, int), entity: (int, int), state_change: str, arg1: (int, int), arg2: (int, int)"""
```
matt-gardner: Similarly here, this docstring probably isn't needed.

deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
DeNeutoy: We talked about an attention style model which uses the verb/entity as a query, so we would want it as the actual word index. However, I guess that method would work too, as in the model you could just use the binary vector to extract the indices from the sentence anyway.

matt-gardner: Even in that case, though, you probably don't need to duplicate the word ids - you just pull them out again after they've been embedded, using the binary vector.  If you do it this way, you need to worry about padding the verb and entity indices to their own lengths, and you need to run them through embeddings / encoders separately.

matt-gardner: I guess, if you envision a model that takes arbitrary entities and verbs as queries, that are not necessarily in the sentence, you might need something like what you're suggesting.  But because these are literally indices into the sentence, you probably don't need to do this yet.

bhavanadalvi:
I incorporated all your comments. I could not fix following test-case: test_convert_instance_to_indexed_instance 

I can see that both values in line #90 are same, still the following check fails:
        assert_array_almost_equal(train_labels, expected_label)

Can you please check the code for as_training_data, and test_convert_instance_to_indexed_instance?

===========================================================
>       assert_array_almost_equal(train_labels, expected_label)
E       ValueError: 
E       error during assertion:
E       
E       Traceback (most recent call last):
E         File "/Users/bhavanad/anaconda/lib/python3.5/site-packages/numpy/testing/utils.py", line 720, in assert_array_compare
E           val = safe_comparison(x, y)
E         File "/Users/bhavanad/anaconda/lib/python3.5/site-packages/numpy/testing/utils.py", line 665, in safe_comparison
E           return comparison(*args, **kwargs)
E         File "/Users/bhavanad/anaconda/lib/python3.5/site-packages/numpy/testing/utils.py", line 910, in compare
E           z = z.astype(float_)  # handle object arrays
E       ValueError: setting an array element with a sequence.
E       
E       
E       Arrays are not almost equal to 6 decimals
E        x: array([array([1], dtype=int32),
E              array([[0, 1],
E              [1, 0],
E              [1, 0]], dtype=int32)], dtype=object)
E        y: array([array([1], dtype=int32),
E              array([[0, 1],
E              [1, 0],
E              [1, 0]], dtype=int32)], dtype=object)

tests/data/instances/sequence_tagging/verb_semantics_instance_test.py:90: ValueError

DeNeutoy:
Nice - the error you are getting is because numpy is trying to convert your data format into an array when it does `assert_array_almost_equal`, but it can't because it isn't square. Instead, just check the two elements individually - eg:

```
assert_array_almost_equal(train_labels[0], expected_label[0])
assert_array_almost_equal(train_labels[1], expected_label[1])

```

bhavanadalvi:
arghhh....
Thanks Mark for pointing that out, will fix this :P

bhavanadalvi:
All test cases are running fine now, and zero pylon errors. Please check if the PR looks fine :-)

bhavanadalvi:
I "meant zero pylint errors"

deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
```diff
@@ -0,0 +1,208 @@
...
+        # verb indices, entity indices are one-hot vectors representing spans within the sentence
+        verb_indices = numpy.zeros(len(sentence_indices))
+        for entity_id in range(self.verb[0], self.verb[1] + 1):
+            verb_indices[entity_id] = 1
+
+        entity_indices = numpy.zeros(len(sentence_indices))
+        for entity_id in range(self.entity[0], self.entity[1] + 1):
+            entity_indices[entity_id] = 1
+
+        # indexed_label = (state_change, tags)
```
matt-gardner: This looks too much like commented out code - can you make it more obvious it's a comment, by adding some English words to it?

deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
```diff
@@ -0,0 +1,208 @@
...
+    """
+    def __init__(self, sentence: List[str], verb: (int, int), entity: (int, int),
+                 state_change: str, arg1: (int, int), arg2: (int, int), index: int=None):
+        """
+        sentence: List[str]
+            A pre-tokenized sentence from which we want to predict verb semantics.
+            A sentence can mention multiple verbs and entities, a data instance
+            is talking about a particular (verb, entity) pair in a given sentence.
+        verb: (int, int)
+            Span of the verb indicating the event being referred to in the query
```
DeNeutoy: Full stops and capitals on comments here.

deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+    def __str__(self):
+        return 'VerbSemanticsInstance(' + self.sentence + ', ' + self.verb + ', ' + \
+               self.entity + str(self.label) + ')'
+
+    def __make_tag_sequence(self, arg1: (int, int), arg2: (int, int)) -> List[str]:
+        """
+        Convert (arg1, arg2) into a tag sequence over the sentence
+        arg1: (int, int)
```
DeNeutoy: Full stops and and spaces after commas in the comments here.

deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
```diff
@@ -0,0 +1,208 @@
...
+                                            entity_indices, indexed_label, self.index)
+
+    @classmethod
+    @overrides
+    def read_from_line(cls, line: str):
+        """
+        Reads a VerbSemanticsInstance object from a line.  The format is as follows:
+        [sentence][tab][verb][tab][entity][tab][stateChangeLabel][tab][arg1][tab][arg2]
+
+        sentence: this is pre-tokenized with tokens separated by "####"
```
DeNeutoy: Capitals and full stops on comments.

tests/data/instances/sequence_tagging/verb_semantics_instance_test.py
```diff
@@ -0,0 +1,121 @@
...
+        expected_verb_indices = [0, 1, 0]
+        expected_entity_indices = [0, 0, 1]
+
+        assert_array_almost_equal(indexed_instance.sentence, expected_sentence_indices)
+        assert_array_almost_equal(indexed_instance.verb, expected_verb_indices)
+        assert_array_almost_equal(indexed_instance.entity, expected_entity_indices)
+
+        b_array = numpy.asarray([0, 1], dtype='int32')
+        o_array1 = numpy.asarray([1, 0], dtype='int32')
+        o_array2 = numpy.asarray([1, 0], dtype='int32')
```
matt-gardner: You don't need two `o_arrays` here, you could just have one, and use it twice on line 73 below.

tests/data/instances/sequence_tagging/verb_semantics_instance_test.py
```diff
@@ -0,0 +1,121 @@
...
+
+        assert_array_almost_equal(train_labels[0], expected_label[0])
+        assert_array_almost_equal(train_labels[1], expected_label[1])
+
+    def test_padding(self):
+        line_record = "roots####absorb####water\t1,1\t2,2\tMOVE\t-1,-1\t0,0"
+        data_indexer = DataIndexer()
+
+        instance = VerbSemanticsInstance.read_from_line(line_record)
+        roots_index = data_indexer.add_word_to_index("roots")
```
DeNeutoy: Instead of doing this indexing of your test sentence twice(in each test where you use it), you could do this in the `setUp` method of the test and assign everything to self. This would make the actual tests smaller and easier to read.

tests/data/instances/sequence_tagging/verb_semantics_instance_test.py
```diff
@@ -0,0 +1,121 @@
...
+        data_indexer.add_word_to_index("B-ARG2", namespace="tags")
+
+        indexed_instance = instance.to_indexed_instance(data_indexer)
+        indexed_instance.pad(padding_lengths)
+
+        expected_sentence_indices = [0, 0, roots_index, absorb_index, water_index]
+        expected_verb_indices = [0, 0, 0, 1, 0]
+        expected_entity_indices = [0, 0, 0, 0, 1]
+
+        b_array = numpy.asarray([0, 1], dtype='int32')
```
DeNeutoy: what are b_array/ o_array? perhaps use full names for the variables.

deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
```diff
@@ -0,0 +1,208 @@
...
+
+    @classmethod
+    @overrides
+    def empty_instance(cls):
+        return IndexedVerbSemanticsInstance([], [], [], label=([], [[]]), index=None)
+
+    @overrides
+    def get_padding_lengths(self) -> Dict[str, int]:
+        """
+        This simple IndexedInstance only has one padding dimension: word_indices.
```
DeNeutoy: Change `IndexedInstance` to `VerbSemanticsInstance`

deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
```diff
@@ -0,0 +1,208 @@
...
+            verb_indices[entity_id] = 1
+
+        entity_indices = numpy.zeros(len(sentence_indices))
+        for entity_id in range(self.entity[0], self.entity[1] + 1):
+            entity_indices[entity_id] = 1
+
+        # indexed_label = (state_change, tags)
+        # state_change = [1-hot vector of size #change_types]
+        # tags = [list of size sentence_length [1-hot vector of size #tag_types] ]
+
```
DeNeutoy: remove empty line.

tests/data/instances/sequence_tagging/verb_semantics_instance_test.py
matt-gardner: In this case, it actually is the whole name, because "b" and "o" are tags.  It's like my `a_index` variable in other tests, for the word "a".

deep_qa/data/instances/sequence_tagging/verb_semantics_instance.py
bhavanadalvi: arg1_str, arg2_str is never empty, if there is no argument, they will be populated with "-1, -1" int he training data itself.

PR #361, opened 2017-05-16
copy original dataset in data generator

DeNeutoy:


tests/data/data_generator_test.py
```diff
@@ -139,8 +159,12 @@ def __init__(self, index, a_length, b_length, c_length):
     def get_padding_lengths(self):
         return {'a': self.a_length, 'b': self.b_length, 'c': self.c_length}
 
-    def pad(self, lengths):
-        pass
+    def pad(self, lengths):  # pylint: disable=unused-argument
+        # Re-assign lengths so that we can check that the
+        # padding is applied to the original dataset.
+        self.a_length = self.b_length
```
matt-gardner: This needs to be `self.a_length = lengths['a']`, right?

deep_qa/data/data_generator.py
```diff
@@ -108,12 +109,13 @@ def create_generator(self, dataset: IndexedDataset):
         Main external API call: converts an ``IndexedDataset`` into a data generator suitable for
         use with Keras' ``fit_generator`` and related methods.
         """
+        unpadded_dataset = deepcopy(dataset)
```
matt-gardner: This line needs to be inside the `if self.sort_every_epoch` block.  To be sure of this, make a test that fails with the current code, to be sure that the change actually fixes the test.  I don't think your test fails with current code.

tests/data/data_generator_test.py
```diff
@@ -103,6 +103,26 @@ def test_sort_every_batch_actually_adds_noise_every_batch(self):
...
+                'padding_noise': 0.0,
+                'sort_every_epoch': True,
+                })
+        generator = DataGenerator(self.text_trainer, params)
+        # Initial padding lengths
+        dataset = IndexedDataset(self.instances)
+        initial_padding_lengths = [instance.get_padding_lengths() for instance in dataset.instances]
+        batches = generator.create_generator(dataset)
+        # Generate a bit.
+        _ = [next(batches) for _ in range(5)]
```
matt-gardner: I think a better test would be to generate several epochs worth of batches, and make sure that the lengths of the instances are constant in each batch.

PR #364, opened 2017-05-16
Fixes #330, unpinning the Keras version

### LINKED ISSUE #330: InputShape as a list is broken in latest keras (2.0.4)
AttributeError: 'list' object has no attribute 'shape'
Exception ignored in: <bound method BaseSession.__del__ of <tensorflow.python.client.session.Session object at 0x1211e8c18>>
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/tensorflow/python/client/session.py", line 582, in __del__
AttributeError: 'NoneType' object has no attribute 'TF_DeleteStatus'

matt-gardner:
I _think_ this was the only issue with using the latest Keras.  I guess CI will tell us if that's true or not.


deep_qa/contrib/layers/tree_composition_lstm.py
```diff
@@ -160,141 +160,146 @@ def build(self, input_shape):
...
                                       name='{}_U3'.format(self.name), initializer=self.initializer)
             # Input dimensionality for all V3s is output_dim, and there are 6 V3s: i, fp, fa1, fa2, u, o
-            self.V3 = self.add_weight((self.output_dim, 6 * self.output_dim),
+            self.V3 = self.add_weight(shape=(self.output_dim, 6 * self.output_dim),
                                       name='{}_V3'.format(self.name), initializer=self.initializer)
 
             self.b2 = K.variable(np.hstack((np.zeros(self.output_dim),
-                                            K.get_value(self.add_weight(self.output_dim,
+                                            K.get_value(self.add_weight(shape=self.output_dim,
+                                                                        name='FIXME',
```
DeNeutoy: What do we need to fix about this? Perhaps add a comment/open an issue.

matt-gardner: I was just trying to fix the pylint error, and didn't want to think about this code.  I since thought about it a bit, and gave a kind-of-reasonable name.

PR #365, opened 2017-05-17
Tensorboard logging

DeNeutoy:
This PR allows us to have tensorboard summaries from any tensor within a model by using:

`tf.summary.scalar("name", scalar_tensor)`
`tf.summary.histogram("name", tensor)`
etc

rather than the previous slightly restrictive inbuilt keras ones which only allowed loss summaries or summarising all weights in the network, which isn't very helpful. By default, it records the total loss from the model at each iteration. It's also a step in the right direction for having a completely separate training loop, eventually.


DeNeutoy:
screen shot 2017-05-18 at 5 07 57 pm

This is what the default loss summary looks like. 

At the moment, because we aren't using tensorflow to save weights, we don't have embedding visualisations, but that will be the next thing i'll try to switch over (possibly as part of getting mattp's multi-gpu model integrated) so then we'll be able to do the tsne plots in the browser and stuff.

deep_qa/training/step.py
```diff
@@ -0,0 +1,81 @@
...
+
+import tensorflow
+import numpy
+import keras.backend as K
+
+
+class Step:
+    """Runs a computation graph.
+
+    # Arguments
```
matt-gardner: Formatting here should be numpydoc.

Or we need to just switch to markdown...  But for now, everything's rST / numpydoc, so we should be consistent.

DeNeutoy: Ah - sorry. I keep doing this!

PR #369, opened 2017-05-24
Clean up embedding API

### LINKED ISSUE #307: Clean up the embeddings API and parameter passing
Currently, there are keys around pretrained embeddings, projecting the embeddings, dropout, and so on, that are flat parameters to `TextTrainer`.  There's also an `embedding_dim` parameter, which is a _dict_, with arbitrary allowed keys.  We should make the flat parameters also a part of this dictionary, so the parameters look something like this:

```json
"embeddings": {
  "words": {
    "dim": 100,
    "pretrained_file": "/path/to/glove",
    "fine_tune": false
  },
  "characters": {
    "dim": 16
  }
}
```

matt-gardner:
Fixes #307.  You can see the gist of the change by looking at the changes to the example experiment files.  The actual code changes are just making the experiment spec file changes possible.


deep_qa/data/tokenizers/word_and_character_tokenizer.py
```diff
@@ -81,10 +81,10 @@ def embed_input(self,
         # mask_split_axis argument to VectorMatrixSplit.
         words, characters = VectorMatrixSplit(split_axis=-1)(input_layer)
         word_embedding = embed_function(words,
-                                        embedding_name='word_' + embedding_name,
```
DeNeutoy: Personally I think having the default embedding suffix being "embedding" is more clear, but it's not a big issue.

matt-gardner: This is changing the meaning of that parameter: previously `embedding_name` was used as the layer name, and the key in a dictionary that stored actual embedding layers.  Now, `embedding_name` is used as the key to the _parameters_ that you pass.  If by default you have `_embeddings` as a suffix to the embedding name, that means that in your parameter file you need to have something like this:

```json
"embeddings": {
  "word_embeddings": {
    "dimension": 100
  },
  "character_embeddings": {
    "dimension": 8
  }
}
```

That felt a bit redundant to me.  Instead, it looks like this:
```json
"embeddings": {
  "words": {
    "dimension": 100
  },
  "characters": {
    "dimension": 8
  }
}
```

and if for some reason you want to have multiple embeddings of words, you can specify them like this:
```json
"embeddings": {
  "words_A0": {
    "dimension": 100
  },
  "words_A1": {
    "dimension": 100
  }
}
```

The layer itself still has `_embedding` appended to it, so you still see something like `words_embedding` in the model summary.

DeNeutoy: ah ok.

PR #370, opened 2017-05-25
implemented baseline verb semantics model

bhavanadalvi:


deep_qa/models/sequence_tagging/verb_semantics_model.py
```diff
@@ -0,0 +1,96 @@
+from keras.layers import Dense, Input, TimeDistributed
+
+from deep_qa.layers.attention import WeightedSum
+from deep_qa.layers.encoders.bag_of_words import BOWEncoder
+from overrides import overrides
```
DeNeutoy: This external import should come before the ones from DeepQA.

deep_qa/models/sequence_tagging/verb_semantics_model.py
```diff
@@ -0,0 +1,96 @@
...
+        # Output: (state-change-type, argument-tags)
+
+        # shape: (batch_size, text_length)
+        sentence_input = Input(shape=self._get_sentence_shape(), dtype='int32', name='word_array_input')
+        verb_input = Input(shape=self._get_sentence_shape(), dtype='int32', name='verb_array_input')
+        entity_input = Input(shape=self._get_sentence_shape(), dtype='int32', name='entity_array_input')
+
+        # shape: (batch_size, text_length, embedding_dim)
+        sentence_embedding = self._embed_input(sentence_input)
+        verb_embedding = self._embed_input(verb_input)
```
DeNeutoy: Remember here that your verb and entity embeddings are actually indices of the sentence positions where the verb/entity appear in the sentence - so at the moment, you are embedding either a 0 or a 1, which I think is probably not your intention. We can go through this tomorrow or something.

tests/common/test_case.py
```diff
@@ -95,8 +98,11 @@ def ensure_model_trains_and_loads(self, model_class, args: Params):
             # As above, we'll just pass on this.
             pass
         else:
-            assert_allclose(model.model.predict(model.validation_arrays[0]),
-                            loaded_model.model.predict(indexed_validation_arrays[0]))
+            model_predictions = model.model.predict(model.validation_arrays[0])
```
DeNeutoy: Thanks for making this change - it wouldn't have been a bug we would have caught any time soon!

tests/models/sequence_tagging/verb_semantics_model_test.py
```diff
@@ -0,0 +1,24 @@
...
+                'save_models': True,
+                'show_summary_with_masking_info': True,
+                'instance_type': 'VerbSemanticsInstance',
+                'tokenizer': {'processor': {'word_splitter': 'no_op'}},
+                # Since VerbSemanticsModel has 2 outputs, validation metric cannot be "validation accuracy",
+                # hence setting it to "validation loss".
+                'validation_metric': 'val_loss',
+                })
+
+        # Serialization and deserialization of the model currently results in following error:
```
DeNeutoy: Is this error still happening? I think this is todo with a bug in Keras which Matt recently fixed - can you re-install the requirements by running `pip install -r requirements.txt`? Once you've done this, remove this comment.

bhavanadalvi: yes, this error is still happening. Matt had installed the latest version of Keras on my machine, don't know why that did not help.

deep_qa/models/sequence_tagging/verb_semantics_model.py
bhavanadalvi: Oh yes, good catch!
sure, will talk to you tomorrow :-)

tests/models/sequence_tagging/verb_semantics_model_test.py
matt-gardner: I installed keras==2.0.4, not the git commit in `requirements.txt`.  Installing that one does indeed look like it will fix this issue.

bhavanadalvi:
I am done with fixing the pylon errors and all checks pass now. Can I merge the PR?

.gitignore
```diff
@@ -1,7 +1,19 @@
 *.pyc
+*.egg
```
matt-gardner: These changes shouldn't be in the PR, they should be in your global `.gitignore` file in `~/.gitignore`.  We don't want to pollute the repository's gitignore file with things that are specific to your development environment.

bhavanadalvi: agree, I reverted .gitingore to allenai/master version

deep_qa/models/sequence_tagging/verb_semantics_model.py
```diff
@@ -0,0 +1,102 @@
...
+        verb_indices = multiply_layer([sentence_input, verb_input])
+        verb_embedding = self._embed_input(verb_indices)
+        entity_indices = multiply_layer([sentence_input, entity_input])
+        entity_embedding = self._embed_input(entity_indices)
+
+        # For state-change-type prediction: We first convert verb and entity into bag-of-words(BOW) representation
+        # and then apply a dense layer with soft-max activation to predict state change type.
+        bow_features = BOWEncoder()
+        weighted_sum_layer = WeightedSum(name="verb_entity_vectors", use_masking=False)
+        verb_entity_vectors = weighted_sum_layer([bow_features(verb_embedding), bow_features(entity_embedding)])
```
DeNeutoy: This is not correct - WeightedSum takes a matrix and a vector of row weights with which to weight the matrix by, whereas here you are passing 2 vectors. You should just use the `Average` Keras layer instead.

DeNeutoy: Also can you change this variable name to `verb_entity_vector`, as there's only one vector.

bhavanadalvi: This was a major bug, thanks for catching 👍 
I switched to Average layer.
Since I am using Average and Multiple directly from Keras, would there be any padding issues?

bhavanadalvi: Done

DeNeutoy: No - generally Keras handles masking correctly, but there are a few cases with more complex layers which they don't handle. A good rule is that if there is a layer in DeepQA with similar functionality to one that you find in Keras, you should use the DeepQA one. We normally say in the docs as well whether we've overridden it to make it do masking correctly.

PR #372, opened 2017-05-25
Clean multi-gpu training

DeNeutoy:


DeNeutoy:
FYI @matt-gardner  - this gpu stuff works(as in 2 gpus is nearly twice as fast over the epoch) but there is some small indexing bug which i'll find tomorrow. But nearly there!

DeNeutoy:
Actually there wasn't a bug, I was just being dumb and running one of the experiments with dynamic padding by accident. 

1 GPU: 1126 sec/epoch
2 GPU: 592 sec/epoch
3 GPU 415 sec/epoch

All with nearly full GPU utilisation (~90%). I think it should be almost trivial to get it working with dynamic batching as well, so i'll give that a crack tomorrow.

deep_qa/training/models.py
```diff
@@ -146,6 +150,175 @@ def _make_predict_function(self):
...
+                except TypeError:
+                    raise TypeError('TypeError while preparing batch. '
+                                    'If using HDF5 input data, '
+                                    'pass shuffle="batch".')
+
+                # Here is the main difference between a single gpu model and one split
+                # across multiple gpus. In our multiple gpu model, all of the inputs
+                # are replicated num_gpus times, so we need to split our large batch
+                # into the corresponding sets of smaller batches for each model.
+                if hasattr(self, "num_gpus"):
```
DeNeutoy: Checking this is a bit of a hack - do you think it would be better to add `self.num_gpus` as a variable which gets passed to compile?

deep_qa/training/trainer.py
```diff
@@ -399,6 +403,90 @@ def evaluate_model(self, data_files: List[str], max_instances: int=None):
...
+        updates += [train_operation]
+        # Gets loss and metrics. Updates weights at each call.
+        primary_model = tower_models[0]
+        primary_model.train_function = Step(inputs,
+                                            [train_loss] + merged_metrics,
+                                            global_step,
+                                            summary_writer=train_summary_writer,
+                                            summary_frequency=self.tensorboard_frequency,
+                                            updates=updates)
+        # TODO(Mark): This is a bit hacky. Add to compile instead?
```
DeNeutoy: This is how the model.num_gpus attribute get's set, i'm thinking now this is obviously something to go in compile.

deep_qa/training/train_utils.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def average_gradients(tower_gradients: List[List[Tuple]]):
+    """
+    Given a list of (gradient, variable) pairs from the result of
+    a gradient calculation from multiple GPUs, calculate their
+    average.
+    """
+    # Make a map from variables -> [gradients that are not none].
+    gradient_map = {}
```
matt-gardner: `gradient_map = defaultdict(list)` lets you remove the `if variable not in gradient_map` check below.

deep_qa/training/train_utils.py
```diff
@@ -0,0 +1,113 @@
...
+    return average_gradient_list
+
+
+def _deduplicate_indexed_slices(values, indices):
+    """
+    Sums `values` associated with any non-unique `indices`.
+
+    Parameters
+    ----------
+    values: A `Tensor` with rank >= 1.
```
matt-gardner: Formatting here as I mentioned earlier.

deep_qa/training/train_utils.py
```diff
@@ -0,0 +1,113 @@
...
+            mean_grad = tensorflow.reduce_mean(grad, 0)
+            average_gradient_list.append((mean_grad, variable))
+
+    assert len(average_gradient_list) == len(gradient_map)
+    return average_gradient_list
+
+
+def _deduplicate_indexed_slices(values, indices):
+    """
+    Sums `values` associated with any non-unique `indices`.
```
matt-gardner: You need double backticks to get monospace font in RST.  Single backticks is italics.

tests/training/multi_gpu_test.py
```diff
@@ -4,7 +4,7 @@
 import keras.backend as K
```
matt-gardner: You should probably rename this test.  And can you add tests for the methods you added to `train_utils`?

deep_qa/training/train_utils.py
```diff
@@ -0,0 +1,113 @@
...
+    def _assign(graph_op):
+        node_def = graph_op if isinstance(graph_op, tensorflow.NodeDef) else graph_op.node_def
+        if node_def.op in ['Variable', 'VariableV2']:
+            return variable_device
+        else:
+            return device
+    return _assign
+
+
+def average_gradients(tower_gradients: List[List[Tuple]]):
```
matt-gardner: This type seems off to me.  At the least, `Tuple` should be parameterized somehow.

deep_qa/training/train_utils.py
```diff
@@ -0,0 +1,113 @@
...
+            # IndexedSlices object.
+            indices = []
+            values = []
+            for grad in gradients:
+                indices.append(grad.indices)
+                values.append(grad.values)
+            all_indices = tensorflow.concat(indices, 0)
+            avg_values = tensorflow.concat(values, 0) / len(gradients)
+            # Deduplicate across indices.
+            values, unique_indicies = _deduplicate_indexed_slices(avg_values, all_indices)
```
matt-gardner: I might call this something like `_collapse_sparse_vector`, or `_deduplicate_sparse_vector`.  Maybe `IndexedSlices` is tensorflow's name for a sparse vector, so this is equivalent, but `sparse_vector` is the more immediately recognizable name to someone who's not familiar with tensorflow.

matt-gardner: Or, another option, just split this whole block into a `_average_sparse_gradients` method, and split the `else` block into a `_average_dense_gradients` method, which would give a simpler method, anyway.  Then the `_deduplicate_indexed_slices` method could just be in-lined.  If you go that route, you might think a bit more about names - there's a lot of overlap between `average_gradients` and the names I suggested.

deep_qa/training/trainer.py
```diff
@@ -399,6 +403,90 @@ def evaluate_model(self, data_files: List[str], max_instances: int=None):
         for idx, metric in enumerate(self.model.metrics_names):
             print("{}: {}".format(metric, scores[idx]))
 
+    def compile_parallel_model(self) -> DeepQaModel:
```
matt-gardner: Method should be private, I think.

deep_qa/training/trainer.py
```diff
@@ -399,6 +403,90 @@ def evaluate_model(self, data_files: List[str], max_instances: int=None):
...
+                                             trainable=False)
+
+        # Place a copy of the model on each GPU, each getting a slice of the batch.
+        for gpu_index in range(self.num_gpus):
+            with tensorflow.device(pin_variable_device_scope('/gpu:%d' % gpu_index)):
+                with tensorflow.name_scope('tower_%d' % gpu_index):
+                    # This is a new model object every time.
+                    model = self._build_model()
+                    # We are using the optimizer directly here, so we don't clutter
+                    # the graph creation by not creating optimizers we don't use.
```
matt-gardner: I think you've got an extra "not" in there.

deep_qa/training/trainer.py
matt-gardner: Oh, I see what you mean.  Instead of checking for `hasattr` in `model._fit_loop`, you just always add an attribute to `DeepQaModel` that will be `1` for the tower models, but `>1` for the primary model?  Seems reasonable.

deep_qa/training/trainer.py
```diff
@@ -534,11 +622,9 @@ def _get_callbacks(self):
         # Some witchcraft is happening here - we don't specify the epoch replacement variable
         # checkpointing string, because Keras does that within the callback if we specify it here.
         if self.save_models:
-
-            checkpoint_callback = ReplicaModelCheckpoint if self.num_gpus > 1 else ModelCheckpoint
-            checkpointing = checkpoint_callback(self.model_prefix + "_weights_epoch={epoch:d}.h5",
-                                                save_best_only=True, save_weights_only=True,
-                                                monitor=self.validation_metric)
+            checkpointing = ModelCheckpoint(self.model_prefix + "_weights_epoch={epoch:d}.h5",
```
matt-gardner: Does this mean that a model trained on 4 GPUs will be loaded into 4 GPUs?

deep_qa/training/models.py
```diff
@@ -146,6 +151,183 @@ def _make_predict_function(self):
...
+        """
+        Abstract fit function which preprocesses and batches
+        data before training a model. We override this keras backend
+        function to support multi-gpu training via splitting a large
+        batch size across multiple gpus. This function is broadly the
+        same as the Keras backend version aside from this - changed elements
+        have corresponding comments attached.
+
+        Note that this should not be called directly - it is used by calling
+        model.fit()
```
DeNeutoy: No - this is the next step, which would get it working with `DataGenerators`.  I think it should be do-able, possibly requiring overriding large methods in keras' backend only to change small things, but for that reason I was going to leave it for a separate PR.

deep_qa/training/train_utils.py
DeNeutoy: Ah thanks - it is a tuple of the gradient and variable pairs.

deep_qa/training/trainer.py
DeNeutoy: No, this is the nice part about only using this for training - the reference to the model which is saved is  _one_ of the towers.

matt-gardner: Oh, I get it.  I missed the `primary_model = tower_models[0]` line, and was thinking the `primary_model` was some global thing that contained the tower models.  But that was how the previous code works, and this one just hacks things a bit, overwriting the first tower model's training function with the losses and gradients of all of the tower models.

And this only happens at training time, and not a validation time, because we're only overwriting the train function?

In any event, the information in this discussion should be summarized in the docstring to this method, I think.

deep_qa/training/models.py
```diff
@@ -146,6 +151,183 @@ def _make_predict_function(self):
...
+                for label, output in zip(out_labels, outs):
+                    batch_logs[label] = output
+
+                callbacks.on_batch_end(batch_index, batch_logs)
+
+                if batch_index == len(batches) - 1:  # Last batch.
+                    if do_validation:
+                        # If we are using multiple gpus, our batch size will be
+                        # scaled up accordingly. However, validation will run
+                        # on a single gpu, so we divide by the number of gpus
```
DeNeutoy: This requires that we also override the `make_test_function` to provide multi-gpu functionality. I think it would be better to keep multi-gpu stuff as training only for the time being, as this is the time intensive part - we can add it later if required.

tests/training/multi_gpu_test.py
DeNeutoy: To what?

matt-peters:
One high level design question/observation:  You have split the multiple GPU stuff across several different classes with fairly different main objectives.  The `DeepQAModel` class has low level details about how to generate batches in the multiple GPU regime in `_fit_loop`.  However, the `Trainer` class also has low level details related to the specific multiple GPU implementation in `_compile_parallel_model`.  This tightly bounds together the implementations and will make them much harder to separate or maintain.  With the goal of building modular, reusable components I'd suggest you encapsulate the details of the multiple GPU implementation in one place, that could be used independent of the details of deep QA's `Trainer` class.  One use case for this is where someone wants to use just use some pieces of deep_qa, but not all of it:  e.g., the pieces that interact with Keras/tensorflow like the layers, models and multiple GPU implementation.

matt-gardner:
The `DeepQaModel` class exists and needs to be modified because we're using Keras' training code, and this is our subclass of Keras' `Model`, so that we can modify Keras' loop.  I'm totally fine with ditching Keras' training code entirely, if we can do it better because we're just focused on tensorflow.  That would essentially make `DeepQaModel` go away, instead just returning input and output tensors in `_build_model`.

I would push back a bit on the "make this library re-usable in pieces" part, though.  Yes, we should definitely make the library as modular as possible, as that's just good software design.  But I don't think it should be our explicit goal to let someone just use parts of this library.  That's not really the point, and trying to support that kind of use could lead us to poor design choices, just like Keras makes suboptimal decisions if all you want to do is use tensorflow.  If it happens as a byproduct, like for the layers, and some of the data processing code, great.  But I don't think it should be a top-level design goal.  Instead, we should make the library good enough to be used directly, instead of just pulling out parts of it.

matt-peters:
We should take this offline for discussion and out of this PR -- but in brief I disagree.  We should have an explicit goal to make subcomponents useful on their own as I think this will naturally lead to more uptake in usage and a much more flexible / extendable framework.  It also makes the code easier to maintain.  Most researchers already have major components of their workflow and infrastructure established and will be more likely to try out deep_qa in bits and pieces if they can easily integrate them into their existing work, vs replacing and starting from scratch.  Of course we can still provide the full functionality for those who want it.

deep_qa/training/models.py
```diff
@@ -146,6 +152,217 @@ def _make_predict_function(self):
...
+                                    'If using HDF5 input data, '
+                                    'pass shuffle="batch".')
+
+                # Here is the main difference between a single gpu model and one split
+                # across multiple gpus. In our multiple gpu model, all of the inputs
+                # are replicated num_gpus times, so we need to split our large batch
+                # into the corresponding sets of smaller batches for each model.
+                if self.num_gpus > 1:
+
+                    def multi_gpu_batch(variable_list):
```
matt-peters: You could pull this function definition out of the loop since it doesn't use any closures.  I'm not sure, but maybe there is some overhead to including the function definition in a loop?

deep_qa/training/trainer.py
```diff
@@ -509,6 +515,95 @@ def _dataset_indexing_kwargs(self) -> Dict[str, Any]:
...
+                    compile_kwargs['optimizer'] = None
+                    model.compile(self.__compile_kwargs())
+                    loss = model.total_loss
+                    tower_models.append(model)
+                    grads = self.optimizer.compute_gradients(loss)
+                    tower_gradients.append(grads)
+                    train_loss += loss
+
+        grads = average_gradients(tower_gradients)
+        train_operation = self.optimizer.apply_gradients(grads, global_step=global_step)
```
matt-peters: Do you plan to support gradient clipping?  I've found it essential for training certain kinds of RNNs.  This would be the place to include it.

deep_qa/training/trainer.py
```diff
@@ -509,6 +515,95 @@ def _dataset_indexing_kwargs(self) -> Dict[str, Any]:
...
+        train_loss = tensorflow.get_variable('train_loss', [],
+                                             initializer=tensorflow.constant_initializer(0.0),
+                                             trainable=False)
+
+        # Place a copy of the model on each GPU, each getting a slice of the batch.
+        for gpu_index in range(self.num_gpus):
+            with tensorflow.device(pin_variable_device_scope('/gpu:%d' % gpu_index)):
+                with tensorflow.name_scope('tower_%d' % gpu_index):
+                    # This is a new model object every time.
+                    model = self._build_model()
```
matt-peters: Does `_build_model` ever have any side effects?  Like creating objects and stashing them away for re-use?

deep_qa/training/trainer.py
```diff
@@ -509,6 +515,95 @@ def _dataset_indexing_kwargs(self) -> Dict[str, Any]:
...
+        """
+        tower_models = []
+        tower_gradients = []
+        global_step = tensorflow.train.get_or_create_global_step()
+        train_loss = tensorflow.get_variable('train_loss', [],
+                                             initializer=tensorflow.constant_initializer(0.0),
+                                             trainable=False)
+
+        # Place a copy of the model on each GPU, each getting a slice of the batch.
+        for gpu_index in range(self.num_gpus):
```
matt-peters: I don't see anywhere in this loop to reuse variables on the GPU copies.  Is that hidden somewhere in the magic of `pin_variable_device_scope`?

deep_qa/training/trainer.py
matt-gardner: Yes, it does.  That's a good thing to think about.  But it _should_ be ok, because the only objects that get created are `Layer` objects that are stored in keyed dictionaries, and reused if they're called again.  But we should definitely document somewhere that when you use data parallelism, this method will get called several times, so you need to be careful about what you do in it.

deep_qa/training/trainer.py
DeNeutoy: This [line in this commit](https://github.com/fchollet/keras/commit/82f9e2358c35c5bbff8714280174617ba29cf92b#diff-fe3ca734e8421a17729d0b0ea4c800baR445) added name scoping into building keras layers(although not to address this problem, just to make better layer scopes for tensorboard).

tf.name_scope scopes all ops, with the exception of calls to `tf.get_variable`. This includes `tf.Variable` which is what keras uses in all of it's backend tensorflow variable creation methods. 
example:
```
with tf.name_scope("test"):
    var1 = tf.Variable([], name="var1")

assert var1.name == 'test/test_var:0' # True

with tf.name_scope("test"):
    var2 = tf.get_variable("var2", [])

assert var2.name == "var2:0"
```
As we keep all of the variables on the cpu with the `pin_variable_device_scope` and always use the `compile` method of keras to create the graphs, the variables are reused.

tests/training/multi_gpu_test.py
matt-gardner: Oh, I just meant that you're testing the `train_utils.py` functions in a file called `multi_gpu_test.py`.  These should match, generally.

deep_qa/training/multi_gpu.py
```diff
@@ -1,88 +1,128 @@
...
-    """
-    def _assign(graph_op):
-        node_def = graph_op if isinstance(graph_op, tensorflow.NodeDef) else graph_op.node_def
-        if node_def.op in ['Variable', 'VariableV2']:
-            return variable_device
-        else:
-            return device
-    return _assign
+    This function compiles a multi-gpu version of your model. This is done using data
+    parallelism, by making N copies of the model on the different GPU, all of which
```
matt-gardner: `s/GPU/GPUs/`

deep_qa/training/multi_gpu.py
```diff
@@ -1,88 +1,128 @@
...
+    share parameters. Gradients are updated synchronously, using the average gradient
+    from all of the outputs of the various models. This effectively allows you to scale
+    a model up to batch_sizes which cannot fit on a single GPU.
 
+    This method returns a "primary" copy of the model, which has had it's training
```
matt-gardner: `s/it's/its/`

deep_qa/training/multi_gpu.py
```diff
@@ -1,88 +1,128 @@
...
+    function which is run by Keras overridden to be a training function which trains
+    all of the towers of the model. The other towers never have their training functions
+    initialised or used and are completely hidden from the user. The returned model
+    can be serialised in the same way as any other model and has no dependency on
+    multiple gpus being available when it is loaded.
+
+    Note that in the multiple gpu case, the model_builder function will be called multiple times
```
matt-gardner: Calling this function _is_ the "multiple GPU case", so you can probably simplify this.

PR #374, opened 2017-05-26
pin awscli version

DeNeutoy:
See email from Jesse - we need the more recent version to support this `--no-include-email` flag, which we use to login to ECR for our Docker images.


requirements.txt
```diff
@@ -48,6 +48,9 @@ spacy
 numpy
 matplotlib
 
+# aws commandline tools for running on Docker remotely.
+awscli>=1.1
```
matt-gardner: Dshouldnt this be 1.11.91, matching the comment in the other file?

Issue #314, opened 2017-04-20
Make docs incorporate READMEs


matt-gardner:
We have a few places under `doc/` where we've written explanations of things.  Those explanations really should be in READMEs in the code, and just _read_ from the README and put into `doc/`, instead of duplicated in both places.

I don't know if this is easy or hard, because I don't know enough about how sphinx works, or how ReadTheDocs actually builds our docs for each commit.  It's probably hard.

nelson-liu:
This is probably not too hard. Here's an example of how scikit-learn does it:

- There's an [`about.rst`](https://github.com/scikit-learn/scikit-learn/blob/master/doc/about.rst) in the `doc/` directory
  - If you look at the [raw source of that rst](https://raw.githubusercontent.com/scikit-learn/scikit-learn/master/doc/about.rst), note the `.. include:: ../AUTHORS.rst`. This basically tells sphinx "ok, go to that path, take all the stuff in [that RST file](https://github.com/scikit-learn/scikit-learn/blob/master/AUTHORS.rst) and put it in here"
- Then, the final generated doc [looks like this](http://scikit-learn.org/stable/about.html). Note that the contents of `AUTHORS.rst` has been inserted

matt-gardner:
Nice, thanks Nelson!  It probably means that we'd need to move to RST instead of markdown everywhere, though...  I was hoping to go the other way, but I could live with it if it gives us nicer docs.

matt-gardner:
There might be a way to combine the two: https://github.com/sphinx-doc/sphinx/issues/2840#issuecomment-284863143.

PR #376, opened 2017-05-30
possible way to get READMEs into docs

DeNeutoy:
Here is a possible way to get the .md files in the docs - not ideal to have to convert them and it also means we can't use relative github paths in the .md files, but it seems to work.

I tried doing the way you suggested in #314, but I couldn't get it to recognise the imported .md file - I tried several ways of doing this, including switching index.rst over to a md file. 

I wanted to see what you thought initially with this approach before I clean up the links and include more of the readmes in the other directories.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,18 @@ the details of your model based on the data that you're working with.
 code for these systems is typically 50 lines or less.
 
 
+This library has several main components:
+
+* A ``training`` module, which has a bunch of helper code for training Keras models of various
+  kinds.
+* A ``models`` module, containing implementations of actual Keras models grouped around various
+  prediction tasks.
+* A ``layers`` module, which contains code for custom Keras Layers that we have written.
+* A ``data`` module, containing code for reading in data from files and converting it into numpy
+  arrays suitable for use with Keras.
+* A ``common`` module, which has a few random things dealing with reading parameters and a few
+  other things.
+
 ## Working in a clone of DeepQA
 
 To train or evaluate a model using a clone of the DeepQA repository, the recommended entry point is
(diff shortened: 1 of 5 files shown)


DeNeutoy:
@nelson-liu any idea how I can stop this error happening whilst still using `html-strict`  as the build command? Does it even need to be `strict`?

nelson-liu:
I think the original thinking behind using `html-strict` was to make sure our docstrings were parseable.
looks like this isnt a new problem though, and this stackoverflow question seems to have some possible solutions? https://stackoverflow.com/a/28778969/2544124

nelson-liu:
to be clear, this looks like the relevant thing to add to `conf.py`
```
suppress_warnings = ['image.nonlocal_uri']
```

see:  https://github.com/sphinx-doc/sphinx/issues/2466#issuecomment-213211534

DeNeutoy:
Yeah, that doesn't seem to be working: https://github.com/allenai/deep_qa/pull/376/files#diff-684a0ceb2878d48b19612ca360e277efR62. 

I'll try that monkey-patch fix, but maybe that's not preferable to just using the non-strict version... Thanks for looking!

doc/conf.py
```diff
@@ -54,6 +57,10 @@
 # The master toctree document.
 master_doc = 'index'
 
+# Suppress warnings for images with external uri's so that we get the benefits of
+# the other checks using html-strict.
+supress_warnings = ["image.nonlocal_uri"]
```
nelson-liu: `s/supress_warnings/suppress_warnings`?

DeNeutoy: 😭

matt-gardner:
For the longer term question, I think it'd be preferable to switch everything (including docstrings) to markdown, instead of RST, we'd just have to figure out how to build the docs correctly in that setting.  There are some complicated issues there, though, and it's not super high priority.

For what you've done in this PR, I think this is a great way to get us part of the way there, and at least not duplicate the README files that we have.  So, yes, I think this approach is a good idea.

matt-gardner:
Changed the name of this so it's easier to keep track of what needs to be looked at.  When it's ready to be looked at again (or if it already is), change it back (and/or add the "Please Review" label).

PR #383, opened 2017-06-02
data instance for frame cloze problem

nikett:
This is similar to tuple instance.

### DIFF
--- a/deep_qa/data/instances/text_classification/frame_instance.py
+++ b/deep_qa/data/instances/text_classification/frame_instance.py
@@ -1,0 +1,174 @@
+from typing import Dict, List
+
+import numpy
+from overrides import overrides
+
+from ..instance import TextInstance, IndexedInstance
+from ...data_indexer import DataIndexer
+
+
+# Functionality supported in this class.
+#
+#    raw input line:
+#            "event:plant absorb water###participant:water###agent:plant###finalloc:soil" TAB "agent:plant"
+#
+#    formatted to:
+#           ["plant", "unk", "unk", "unk", "unk", "plant absorb water",
+#            "soil", "unk", "unk", "unk", "unk", "unk",
+#            "unk", "unk", "unk", "unk", "unk", "unk",
+#            "unk", "unk", "unk", "unk", "unk", "unk",
+#            "unk", "unk", "water"]
+#
+#    padded to a fixed length of say, 5 words per phrase.
+#           ["plant P P P P", "unk P P P P", "unk P P P P", "unk P P P P", "unk P P P P", "plant absorb water P P ",
+#             ...
+
+# TODO PR request for having these in the json as an application specific content
+# the slotnames can vary according to different end applications, e.g., a HowTo tuple, OpenIE tuple ...
+SLOTNAMES_ORDERED =["agent", "beneficiary", "causer", "context", "definition", "event",
+     "finalloc", "headverb", "initloc", "input", "output", "manner",
+     "patient", "resultant", "timebegin", "timeend", "temporal", "hierarchical",
+     "similar", "contemporary", "enables", "mechanism", "condition", "purpose",
+     "cause", "openrel", "participant"]
+UNKNOWN_SLOTVAL = "unk"  # making an open world assumption, we do not observe all the values
+QUES_SLOTVAL = "ques"  # this slot in the frame must be queried/completed.
+MAX_PHRASE_LEN = 6
+
+class FrameInstance(TextInstance):
+
+    """
+    A FrameInstance is a kind of TextInstance that has text in multiple slots. This generalizes a FrameInstance.
+    """
+    def __init__(self,
+                 # the input can have only few slots and not all and in any random order.
+                 # e.g., "event:plant absorb water###participant:water###agent:plant###finalloc:soil" TAB "agent:plant"
+                 text: List[str],
+                 # output is a phrase
+                 label: str=None):
+        super(FrameInstance, self).__init__(label)
+        self.text = text  # "event:plant absorb water###participant:water###agent:plant" TAB "agent:plant"
+
+    def __str__(self):
+        return 'FrameInstance( [' + ',\n'.join(self.text) + '] , ' + str(self.label) + ')'
+
+    @overrides
+    # accumulate words from each slot's phrase
+    def words(self) -> Dict[str, List[str]]:
+        words = []
+        for phrase in self.text:  # phrases
+            phrase_words = self._words_from_text(phrase)
+            words.extend(phrase_words['words'])
+        label_words = self._words_from_text(self.label)
+        words.extend(label_words['words'])
+        return {"words": words}
+
+    @staticmethod
+    # s: "participant:water" => name=participant, val=water
+    def slot_from(s: str, kv_separator: str=":"):
+        kv = s.split(kv_separator)
+        return {'name': kv[0], 'val': kv[1]}
+
+    @staticmethod
+    # s: "event:plant absorb water###participant:water" TAB "participant:water"
+    def unpack_input(s: str, kv_separator: str="\t"):
+        # no information is lost in lowercasing, and simplifies matching.
+        kv = s.lower().split(kv_separator)
+        if len(kv) != 2:
+            raise RuntimeError("Unexpected number (not 2) of fields in frame: " + s)
+        return {'content': kv[0], 'query': kv[1]}
+
+    @staticmethod
+    def given_slots_from(s: str, values_separator: str= "###", kv_separator: str= ":"):
+        return dict(map(lambda x: x.split(kv_separator), s.split(values_separator)))
+
+    # Performs two types of padding:
+    # i) unobserved slots are filled with self.unknown_slotval
+    # ii) query slot is masked with self.unknown_queryval
+    @staticmethod
+    def padded_slots_from(sparse_frame: Dict[str, str], query_slotname: str):
+        slots = []
+        for slotname in SLOTNAMES_ORDERED:
+            if slotname == query_slotname:  # query hence masked
+                slots.append(QUES_SLOTVAL)
+            elif slotname in sparse_frame:  # observed hence as-is
+                slots.append(sparse_frame[slotname])
+            else:  # unobserved hence padded
+                slots.append(UNKNOWN_SLOTVAL)
+        return slots
+
+    @classmethod
+    @overrides
+    def read_from_line(cls, line: str):
+        """
+        Reads a FrameInstance from a line.  The format is:
+        
+        frame represented as list of <role:role value phrase of maxlen 5> TAB <label> 
+        e.g., from
+        event:plant absorb water###participant:water###agent:plant###finalloc:soil
+
+        to
+        ["plant", "unk", "unk", "unk", "unk", "plant absorb water",
+            "soil", "unk", "unk", "unk", "unk", "unk",
+            "unk", "unk", "unk", "unk", "unk", "unk",
+           "unk", "unk", "unk", "unk", "unk", "unk",
+            "unk", "unk", "water"]
+            
+        Provides ordering and sparseness flexibility to the input text. 
+        """
+        # Extract the query slot name and expected value
+        # e.g. from, participant:water, extract the expected slot value "water"
+        unpacked_input = cls.unpack_input(line)
+        query_slot = cls.slot_from(unpacked_input['query'])
+        given_slots = cls.given_slots_from(unpacked_input['content'])
+        padded_slots = cls.padded_slots_from(given_slots, query_slot['name'])
+        return cls(padded_slots, label=query_slot['val'])
+
+    @overrides
+    def to_indexed_instance(self, data_indexer: DataIndexer):
+        # a phrase in a slot, is converted from [list of words] to [list of wordids]
+        # this is repeated for every slot, hence a list of list of wordids/integers
+        indices_slotvals = [self._index_text(phrase, data_indexer) for phrase in self.text]
+        # the label is a phrase, and is converted from [list of words] to [list of wordids]
+        indices_label = self._index_text(self.label, data_indexer)
+        return IndexedFrameInstance(indices_slotvals, indices_label)
+
+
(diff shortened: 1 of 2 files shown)


deep_qa/data/instances/text_classification/frame_instance.py
```diff
@@ -0,0 +1,177 @@
...
+    def unpack_input(frame_as_string: str, kv_separator: str="\t"):
+        # no information is lost in lowercasing, and simplifies matching.
+        partialframe_query = frame_as_string.lower().split(kv_separator)
+        if len(partialframe_query) != 2:
+            raise RuntimeError("Unexpected number (not 2) of fields in frame: " + frame_as_string)
+        return {'content': partialframe_query[0], 'query': partialframe_query[1]}
+
+    @staticmethod
+    def given_slots_from(slots_csv: str, values_separator: str="###", kv_separator: str=":"):
+        return dict(map(lambda x: x.split(kv_separator), slots_csv.split(values_separator)))
```
DeNeutoy: Could you add a comment describing this function? It's not clear immediately what it does. You could consider re-writing this like:
```
key_value_list = slots_csv.split(values_separator)
key_value_tuples = [pair.split(key_value_separator) for x in key_value_list]
key_value_dict = {key: value for key, value in key_value_tuples}

```

deep_qa/data/instances/text_classification/frame_instance.py
```diff
@@ -0,0 +1,174 @@
...
+                 # output is a phrase
+                 label: str=None):
+        super(FrameInstance, self).__init__(label)
+        self.text = text  # "event:plant absorb water###participant:water###agent:plant" TAB "agent:plant"
+
+    def __str__(self):
+        return 'FrameInstance( [' + ',\n'.join(self.text) + '] , ' + str(self.label) + ')'
+
+    @overrides
+    # accumulate words from each slot's phrase
```
nikett: Done. I did it the java style, incorrectly.

deep_qa/data/instances/text_classification/frame_instance.py
```diff
@@ -0,0 +1,174 @@
...
+QUES_SLOTVAL = "ques"  # this slot in the frame must be queried/completed.
+MAX_PHRASE_LEN = 6
+
+class FrameInstance(TextInstance):
+
+    """
+    A FrameInstance is a kind of TextInstance that has text in multiple slots. This generalizes a FrameInstance.
+    """
+    def __init__(self,
+                 # the input can have only few slots and not all and in any random order.
```
nikett: Done.

deep_qa/data/instances/text_classification/frame_instance.py
```diff
@@ -0,0 +1,174 @@
+from typing import Dict, List
+
+import numpy
+from overrides import overrides
+
+from ..instance import TextInstance, IndexedInstance
+from ...data_indexer import DataIndexer
+
+
+# Functionality supported in this class.
```
nikett: Done.

deep_qa/data/instances/text_classification/frame_instance.py
```diff
@@ -0,0 +1,177 @@
...
+        # e.g. from, participant:water, extract the expected slot value "water"
+        unpacked_input = cls.unpack_input(line)
+        query_slot = cls.slot_from(unpacked_input['query'])
+        given_slots = cls.given_slots_from(unpacked_input['content'])
+        padded_slots = cls.padded_slots_from(given_slots, query_slot['name'])
+        return cls(padded_slots, label=query_slot['val'])
+
+    @overrides
+    def to_indexed_instance(self, data_indexer: DataIndexer):
+        # a phrase in a slot, is converted from [list of words] to [list of wordids]
```
nikett: Done.

deep_qa/data/instances/text_classification/frame_instance.py
```diff
@@ -0,0 +1,174 @@
...
+
+    @staticmethod
+    # s: "participant:water" => name=participant, val=water
+    def slot_from(s: str, kv_separator: str=":"):
+        kv = s.split(kv_separator)
+        return {'name': kv[0], 'val': kv[1]}
+
+    @staticmethod
+    # s: "event:plant absorb water###participant:water" TAB "participant:water"
+    def unpack_input(s: str, kv_separator: str="\t"):
```
nikett: Data processing in between the logic renders it unreadable. Small functions make the functionality easier to read, and debug.

tests/data/instances/text_classification/frame_instance_test.py
```diff
@@ -1,47 +1,67 @@
...
+        self.padded_slots = ['plant', 'unk', 'unk', 'unk', 'unk',
+                             'plant absorb water', 'ques', 'unk', 'unk',
+                             'unk', 'unk', 'unk', 'unk', 'unk', 'unk',
+                             'unk', 'unk', 'unk', 'unk', 'unk', 'unk', 'unk',
+                             'unk', 'unk', 'unk', 'unk', 'water']
+        self.data_indexer = DataIndexer()
+        for word in ['plant', 'unk', 'absorb', 'ques', 'water', 'soil']:
+            self.data_indexer.add_word_to_index(word)
+
+    def tearDown(self):
```
DeNeutoy: This is not required as you do nothing but call the super class.

tests/data/instances/text_classification/frame_instance_test.py
```diff
@@ -1,47 +1,67 @@
...
+        super(TestFrameInstance, self).tearDown()
+
+    def test_convert_instance_to_indexed_instance(self):
+        instance = FrameInstance.read_from_line(self.line)
+        indexed_instance = instance.to_indexed_instance(self.data_indexer)
+        print(instance.label + "\t padded to: " + indexed_instance.label.__str__())
```
DeNeutoy: Can you remove these print statements?

tests/data/instances/text_classification/frame_instance_test.py
```diff
@@ -1,47 +1,67 @@
...
-    # TODO more test with boundary conditions on imperfect input.
     def test_slots_unwrap_correctly(self):
-        instance = FrameInstance.read_from_line(line)
+        instance = FrameInstance.read_from_line(self.line)
         # what we get
-        machine_label = "" + instance.label
-        machine_slotvals = ""+instance.text.__str__()
+        # TODO: consider when label is OOV or finalloc instead of finalloc:soil
+        machine_label = instance.label
+        machine_slotvals = instance.text
```
DeNeutoy: No need to shorten variable names here - just use `machine_slot_values`.

tests/data/instances/text_classification/frame_instance_test.py
```diff
@@ -1,47 +1,67 @@
...
+
+    def tearDown(self):
+        super(TestFrameInstance, self).tearDown()
+
+    def test_convert_instance_to_indexed_instance(self):
+        instance = FrameInstance.read_from_line(self.line)
+        indexed_instance = instance.to_indexed_instance(self.data_indexer)
+        print(instance.label + "\t padded to: " + indexed_instance.label.__str__())
+        print(indexed_instance.word_indices.__str__())
+        assert indexed_instance.label == [self.data_indexer.get_word_index('soil')]  # TODO
```
DeNeutoy: remove this TODO.

tests/data/instances/text_classification/frame_instance_test.py
nikett: Done

tests/data/instances/text_classification/frame_instance_test.py
nikett: Done

tests/data/instances/text_classification/frame_instance_test.py
nikett: Done

tests/data/instances/text_classification/frame_instance_test.py
```diff
@@ -0,0 +1,86 @@
...
+    def test_slots_unwrap_correctly(self):
+        instance = FrameInstance.read_from_line(self.line)
+        # what we construct
+        machine_label = instance.label
+        machine_slot_values = instance.text
+        # what we expect
+        expected_label = "soil"
+        # do they match?
+        assert machine_label == expected_label
+        assert machine_slot_values == self.padded_slots
```
matt-gardner: The label here is what you're trying to predict.  You want to have the model output "soil".  But it looks like there's nothing in this instance that tells the model what slot you're trying to query.  How would a model know which slot to fill, when it's trying to predict "soil"?

deep_qa/data/instances/text_classification/frame_instance.py
```diff
@@ -0,0 +1,198 @@
...
+    @classmethod
+    @overrides
+    def empty_instance(cls):
+        return IndexedFrameInstance([], label=None)
+
+    @overrides
+    def get_padding_lengths(self) -> Dict[str, int]:
+        # The parent class somehow understands the semantics of these dictionary keys.
+        # would be better if these keys were more explicit, e.g., as enumerations.
+        return {'num_sentence_words': MAX_PHRASE_LEN}
```
matt-gardner: Yeah, there's no need to hard-code a length here, and that parameter is better specified in the model you build, anyway.  This will actually break the code we have that tries to be smart about letting you swap out word representations.  It looks like the tuple instance code that Mark linked to also will break the word representation stuff that we have.  Instead, what you should do is something like [this](https://github.com/allenai/deep_qa/blob/dd6f20cd0fde902cc54f58baff29f355a356c3cd/deep_qa/data/instances/entailment/sentence_pair_instance.py#L66-L71).

@DeNeutoy, I think we need way better documentation around how the whole data pipeline works.  I think it does some really nice things, but people have a hard time understanding it and how to use it correctly.

deep_qa/data/instances/text_classification/frame_instance.py
```diff
@@ -0,0 +1,198 @@
...
+
+import numpy
+from overrides import overrides
+
+from ..instance import TextInstance, IndexedInstance
+from ...data_indexer import DataIndexer
+
+# TODO PR request for having these in the json as an application specific content
+# the slotnames can vary according to different end applications, e.g., a HowTo tuple, OpenIE tuple ...
+SLOTNAMES_ORDERED = ["agent", "beneficiary", "causer", "context", "definition", "event",
```
matt-gardner: If you want to always have a fixed set of slots that every dataset uses, where your input includes a one-hot vector for which slot you're trying to fill, or something like that, this should work.  The model's weights will be tied to particular slots, so a model trained on one slot configuration will not possibly generalize to any other slot configuration.

If you want to make this more modular, you could consider using a separate namespace in the data indexer to keep track of this for you, instead of hard-coding it here.  Something similar is done with tag labels [here](https://github.com/allenai/deep_qa/blob/dd6f20cd0fde902cc54f58baff29f355a356c3cd/deep_qa/data/instances/sequence_tagging/tagging_instance.py#L30).

deep_qa/data/instances/text_classification/frame_instance.py
```diff
@@ -0,0 +1,198 @@
...
+MAX_PHRASE_LEN = 6
+
+
+class FrameInstance(TextInstance):
+
+    """
+    A FrameInstance is a kind of TextInstance that has text in multiple slots. This generalizes a FrameInstance.
+    """
+    def __init__(self,
+                 padded_slots: List[str],
```
nikett: Fixed to sparse and dense frame, instead of unpadded and padded slots.

deep_qa/data/instances/text_classification/frame_instance.py
nikett: I now understand. So something like this?

```  lengths = {}
for key in slot_word_lengths:
            lengths[key] = len(key)
return lengths```

File: deep_qa/data/instances/text_classification/frame_instance.py (hunk 13 above)
Comment (nikett) [inline deep_qa/data/instances/text_classification/frame_instance.py]: The reason to have this configurable was to be able to train any variant of the input with the same code, by passing a list with the config file upfront. This configuration allows ordering the input on the fly for any types of slots. 

It appears that there is no possibility to specify these list of slotnames in the config file, so, perhaps having this as a constant may be better than passing through a function. The data indexer does not really do any book keeping because the frame_instance takes care of it.

File: tests/data/instances/text_classification/frame_instance_test.py (hunk 11 above)
Comment (nikett) [inline tests/data/instances/text_classification/frame_instance_test.py]: Very good point. I was hoping to include a one-hot vector depicting the slot to fill. I thought that by specifying a special tag "ques", I could inform the model eventually. You are probably right, that in the data instance, my input could explicitly include a one-hot vector. I don't know how to do that so I would ask Mark.

File: deep_qa/data/instances/text_classification/frame_instance.py (hunk 13 above)
Comment (matt-gardner) [inline deep_qa/data/instances/text_classification/frame_instance.py]: Oh, yeah, your preprocess scripts putting everything into this format will take care of that, I think.

Another consideration is that you'll be essentially wasting computation and modeling capacity on all of the un-filled slots by filling them with "unk".  You could either: (1) leave them empty, and let them be masked in whatever model computation you do later.  This will keep the model from wasting its capacity trying to figure out that "unk" means "unfilled".  Or (2) instead of having a fixed-sized input here, you could have a key-value input, where you give a list of (slot, value) pairs as input to the model.  In this setting, it's also a lot easier to think about how to pass in the query - it's just another (slot, value) pair, just with a `"QUERY"` value, or something similar.  Then instead of learning what is essentially an embedding of each slot type in the weights of a softmax layer, you can explicitly learn an embedding for them, and have that be part of your model.  Happy to talk about this more in person if it's hard to follow what I'm suggesting here.

File: deep_qa/data/instances/text_classification/frame_instance.py (hunk 12 above)
Comment (matt-gardner) [inline deep_qa/data/instances/text_classification/frame_instance.py]: Yeah, similar.  I would do something like this:

```python
all_slot_lengths = [self._get_word_sequence_lengths(slot_indices) for slot_indices in self.word_indices]
lengths = {}
for key in slot_lengths[0]:
    lengths[key] = max(slot_lengths[key] for slot_lengths in all_slot_lengths)
return lengths
```

PR #384, opened 2017-06-05
WIP: DO NOT MERGE!

schmmd:
This PR is to test the build in Travis--DO NOT MERGE.

### DIFF
--- a/build_tools/travis/test_script.sh
+++ b/build_tools/travis/test_script.sh
@@ -10,8 +10,6 @@ set -e
 
 python --version
 
-export PYTHONHASHSEED=2157
-
 run_tests() {
     KERAS_BACKEND=tensorflow py.test -v --cov=deep_qa --durations=20
 }


matt-gardner:
https://stackoverflow.com/a/39162893/1467533

Try adding
```
env = PYTHONHASHSEED=2157
```
to `pytest.ini`.  That _should_ make the tests pass even with the change you made in your previous commit (and would be an improvement!).

schmmd:
@matt-gardner no dice.  I think it sets the environment variable after python has been loaded, so PYTHONHASHCODE does not take effect.  Yikes, what a setup!

matt-peters:
FWIW, you can specify environmental variables in .travis.yml:  https://docs.travis-ci.com/user/environment-variables/

matt-gardner:
@matt-peters, that's a good idea.

schmmd:
Yes, that's probably better than https://github.com/allenai/deep_qa/blob/master/build_tools/travis/test_script.sh#L13.  I'll try it out today if I have time.

matt-gardner:
Oh, wait, yeah, we already do that, and that's why it works.  Not sure that setting it in .travis.yml actually buys us anything.  CI passes; the issue is that it's opaque and you can easily miss things if you're setting up the environment for yourself, not in CI.  Having a failing test is a good idea here, but I'm not sure that moving it from where it is in the script into .travis.yml helps with clarity.

PR #385, opened 2017-06-05
Add a "Setting up a development environment" section to the README.

schmmd:
Here are some modest changes to the README.

1.  Add a "Setting up a development environment" section to the root README.
2.  Remove most of the `deep_qa/` README as it was redundant.
3.  A few small tweaks.

Feel free to give me all the feedback you've got--I think these are improvements but that can be pretty subjective.

I experimented for a bit to see if I could remove PYTHONHASHCODE from the Getting Started instructions--but it looks like that's not possible.  I also learned that if you run tests without PYTHONHASHCODE set (or set to another value) the tests fail for cryptic reasons as the checks are not run.

The Getting Started section largely copies the Dockerfile.  I think it would be great to have a Dockerfile that creates a development environment--the current one doesn't work because it installs `tensorflow-gpu` rather than `tensorflow` so when you run it on your Mac you get a CUDA exception.


README.md
```diff
@@ -14,32 +14,77 @@ point of having a supported release yet.
...
+    python -m spacy.en.download all
+    ```
+
+5.  Set the `PYTHONHASHSEED` for repeatable experiments.
+
+    ```
+    export PYTHONHASHSEED=2157
+    ```
+
+You should now be able to test your installation with `pytest -v`.  Congradulations!
```
DeNeutoy: Congradulations -> Congratulations

README.md
```diff
@@ -14,32 +14,77 @@ point of having a supported release yet.
...
+  tasks, as well).  The actual model code for these systems is typically 50
+  lines or less.
+
+## Running DeepQA
+
+### Setting up a development environment
+
+DeepQA is built using Python 3.  The easiest way to set up a compatible
+environment is to use [Conda](https://conda.io/).  This will set up a virtual
+environment the exact version of Python used for development along with all the
```
matt-gardner: "This will set up a virtual environment [with?] the exact version of..."

matt-gardner:
And I agree that if we can get `PYTHONHASHSEED` set in pytest, that'd be lovely.  Is there a way to set environment variables in `pytest.ini`? (and do they load before the python interpreter?)

I think it'd be fine to have a parallel dockerfile that's for CPU usage instead of GPU usage, though I don't know enough about docker to know if it's even allowed to have two dockerfiles in the same directory, or how that works if it's possible.

schmmd:
@matt-gardner I'll follow up with Jesse to learn about Dockerfile best practices here.  I'm also not sure what's the best approach.

DeNeutoy:
Could we use a `--local` `ARG` in the Dockerfile which would then install the correct tensorflow package?

matt-gardner:
I think the argument is probably the easiest idea, though I'd call the option `--no-gpu` instead of `--local`.

PR #386, opened 2017-06-06
Add a test case to ensure PYTHONHASHSEED is set.

schmmd:
Previously, tests would fail when PYTHONHASHSEED was not set, but the error messages were not informative to determine the underlying problem.


tests/common/pythonhashseed_test.py
```diff
@@ -0,0 +1,4 @@
+from deep_qa.common.checks import ensure_pythonhashseed_set
+
+def test_pythonhashseed():
+    ensure_pythonhashseed_set()
```
matt-gardner: I wonder if it's worth doing some "assert fails" thing here, and giving an error message specifically about other tests possibly failing because of this one.

matt-gardner: I guess assert fails isn't the right thing; I just mean changing the error message inside the test, e.g., by catching the assertion failure and raising with a new message about how you should probably ignore the other failures if this one is failing.

PR #388, opened 2017-06-06
Simplify installation instructions

schmmd:
* Move Python environment initialization scripts to a script (`scripts/install_requirements.sh`).

* Add an argument to the Dockerfile which controls whether a CPU or a GPU environment is created.


schmmd:
Unfortunately the Dockerfile doesn't create a great environment, since it doesn't contain all the files in the CWD (although we could change that).  Also, the `ENTRYPOINT` is a bit annoying and needs to be overridden on the CLI (`docker run -it deep_qa_cpu --entrypoint=/bin/bash`).

scripts/set_processor.sh
```diff
@@ -0,0 +1,22 @@
...
+
+USAGE="usage: ./set-processor [gpu|cpu]"
+
+if [ $# != 1 ]; then
+    echo "$USAGE"
+    exit 1
+fi
+
+PROCESSOR=$1
+if [ "$PROCESSOR" == "gpu" ]; then
```
schmmd: This is oh-so-nasty.  LMK if you have any better suggestions.

Dockerfile
```diff
@@ -7,18 +7,17 @@ WORKDIR /stage
...
-RUN sed -ie 's/^tensorflow/tensorflow-gpu/g' requirements.txt
-RUN /bin/bash -c "source activate runenv && pip install -r requirements.txt"
-RUN /bin/bash -c "source activate runenv && python -m nltk.downloader punkt"
-RUN /bin/bash -c "source activate runenv && python -m spacy.en.download all"
-
-# Copying the source code.
+# Copying the source code and pip requirements.
 COPY scripts/ scripts/
 COPY deep_qa/ deep_qa/
+COPY requirements.txt requirements.txt
```
matt-gardner: This line should be earlier in the dockerfile than `COPY` lines for the scripts and the code.  This is because of how docker images are built on top of each other.  This line is the most expensive image to build, because it's grabbing and installing a bunch of packages, and the requirements change less frequently than the code does.  So you don't want to have to rebuild the requirements every time you change your code.  I would move the current lines 11 and 12 down below line 20, to maintain the nice build behavior we currently have.

scripts/set_processor.sh
matt-gardner: Yeah, I wondered if there was a better way, then realized this was how we already had it in the dockerfile... As ugly as it looks, I can't think of anything better.

matt-gardner:
I don't understand the issue about overriding the entry point - what's going on there?

Dockerfile
schmmd: Good point.  I added a dependency on `scripts` for installing pip requirements however.  Which of the following would you prefer:

1.  Move `COPY script/ script/` earlier in the Dockerfile
2.  Copy the `set_processor.sh` script separately and move it to the root folder.

schmmd:
@matt-gardner there's no issue overriding the entrypoint--it works fine--you just need to know to do it otherwise it will try and execute the job that's specified.

Dockerfile
matt-gardner: The thing that changes most frequently is the code itself, so as long as that's as late as possible, we should be fine.  A bit annoying that if you change a script you have to reinstall requirements, but not the end of the world.

matt-gardner: I don't think copying `set_processor.sh` separately helps, because you also need `scripts/install_requirements.sh`.  So just put `COPY script/ script/` earlier, I'd say.

PR #389, opened 2017-06-07
Remove memory networks

DeNeutoy:


DeNeutoy:
There are a few more things that I could remove - e.g there are some fairly niche layers which could probably be removed, such as the `memory_updaters` and `knowledge_selectors` etc.

matt-gardner:
I would also remove the niche layers - there's no reason to keep them around, and a lot of their functionality has now been implemented in better ways (e.g., `VectorMatrixSplit`, `Attention`, etc.).  It'd be pretty trivial to re-implement a memory network (or any of the variants deleted here) in our current code, and it would be much better designed that what you're deleting.

DeNeutoy:
@matt-gardner  I kept the `AttentiveGRU` because it's fairly general and not actually connected to the memory network stuff, but apart from that this is ready.

deep_qa/layers/attentive_gru.py
```diff
@@ -80,38 +23,38 @@ class AttentiveGRUKnowledgeCombiner(GRU):
...
         it instead.
-
         Note that there is some redundancy here - for instance, in the GPU mode, we do a
         larger matrix multiplication than required, as we don't use one part of it. However, for
         readability and similarity to the original GRU code in Keras, it has not been changed. In each section,
         there are commented out lines which contain code. If you were to uncomment these, remove the differences
         in the input size and replace the attention with the z gate at the output, you would have a standard
         GRU back again. We literally copied the Keras GRU code here, making some small modifications.
-        '''
+
```
matt-gardner: Extra blank line.

deep_qa/layers/attentive_gru.py
```diff
@@ -168,27 +111,26 @@ def build(self, input_shape):
         This is used by Keras to verify things, but also to build the weights.
         The only differences from the Keras GRU (which we copied exactly
         other than the below) are:
+        We generate weights with dimension input_dim[2] - 1, rather than
+        dimension input_dim[2].
+        There are a few variables which are created in non-'gpu' modes which
+        are not required. These are commented out but left in for clarity below.
```
matt-gardner: Extra blank line.

deep_qa/layers/attentive_gru.py
```diff
@@ -168,27 +111,26 @@ def build(self, input_shape):
...
     def preprocess_input(self, inputs, training=None):
-        '''
+        """
         We have to override this preprocessing step, because if we are using the cpu,
-        we do the weight - input multiplications in the internals of the GRU as seperate,
+        we do the weight - input multiplications in the internals of the GRU as separate,
         smaller matrix multiplications and concatenate them after. Therefore, before this
         happens, we split off the attention and then add it back afterwards.
-        '''
+
```
matt-gardner: Extra blank line.

deep_qa/layers/attentive_gru.py
```diff
@@ -1,76 +1,19 @@
...
-        base_config = super(WeightedAverageKnowledgeCombiner, self).get_config()
-        config.update(base_config)
-        return config
-
-
-class AttentiveGRUKnowledgeCombiner(GRU):
-    '''
+class AttentiveGRU(GRU):
+    """
     GRUs typically operate over sequences of words. Here we are are operating over the background knowledge
```
matt-gardner: I'd update this to remove reference to "background knowledge" - as you said, this is now just a general GRU that takes an additional input (e.g., this would still work if you ran it over a single sentence).  I'd also move it into the `encoders/` module.

deep_qa/layers/attentive_gru.py
```diff
@@ -80,38 +23,38 @@ class AttentiveGRUKnowledgeCombiner(GRU):
...
     When it is initialised, the Keras backend will call the build method. It uses this to check that inputs being
     passed to this function are the correct size, so we allow this to be the actual input size as normal.
-
     However, for the internal implementation, everywhere where this global shape is used, we override it to be one
     less, as we are passing in a tensor of shape (batch, knowledge_length, 1 + encoding_dim) as we are including
     the attention mask. Therefore, we need all of the weights to have shape (*, encoding_dim),
     NOT (*, 1 + encoding_dim). All of the below methods which are overridden use some
     form of this dimension, so we correct them.
-    '''
+
```
matt-gardner: Extra blank line.

deep_qa/layers/attentive_gru.py
```diff
@@ -1,76 +1,19 @@
...
-                "input_length": -1
-                }
-        base_config = super(WeightedAverageKnowledgeCombiner, self).get_config()
-        config.update(base_config)
-        return config
-
-
-class AttentiveGRUKnowledgeCombiner(GRU):
-    '''
+class AttentiveGRU(GRU):
```
matt-gardner: This should be `AttentiveGru`, I think: https://google.github.io/styleguide/cppguide.html#Type_Names, https://google.github.io/styleguide/javaguide.html#s5.3-camel-case.  It's just easier to read the name if the acronym isn't all uppercase.

deep_qa/layers/attentive_gru.py
DeNeutoy: Hmm, i'm not so sure this is a good idea - this layer _requires_ that the `embedding_dim + 1`'th dimension is a distribution, which means that if you specified this as your encoder in a model where you weren't appending the distribution onto your input first, it would do something utterly crazy.

matt-gardner: I guess, it's conceptually an encoder, so to me it goes in the `encoders/` module, but you're right that it doesn't fit the normal encoder API.  So I just wouldn't put it as one of the options in the encoder dictionaries.  Does that make sense?

matt-gardner: And it really should take the attention as a separate input, so it's not possible to make the mistake you're talking about, but let's not worry about making that change until we actually use this somewhere.

DeNeutoy: Yeah agreed, that is best.

DeNeutoy: Also it has to be implemented this way to use the `Recurrent` API in Keras - the input at timestep `t` needs to have the attention from timestep `t` appended, rather than being passed as 2 inputs, as Keras calls `K.rnn` on this function, so you have to have a function which you could `foldRight` over.

PR #390, opened 2017-06-12
Change README to reflect our growing team and goals.

schmmd:


README.md
```diff
@@ -196,19 +199,14 @@ Experiments](https://github.com/allenai/deep_qa_experiments), however.
...
 If you use this code and think something could be improved, pull requests are very welcome. Opening
-an issue is ok, too, but we're a lot more likely to respond to a PR. The primary maintainer of this
-code is [Matt Gardner](https://matt-gardner.github.io/), with a lot of help from [Pradeep
-Dasigi](http://www.cs.cmu.edu/~pdasigi/) (who was the initial author of this codebase), [Mark
-Neumann](http://markneumann.xyz/) and [Nelson Liu](http://nelsonliu.me/).
-
-A note on issues: we are a very small team, and our focus is on getting research done, not on
-building this library.  We do not have anyone dedicated full-time to maintaining and improving
-this code.  As such, we generally do not have bandwidth to solve your problems.  Sorry.  The code
-is well tested and works on our continuous integration server, so if the tests do not pass in your
```
matt-gardner: You didn't like this... =).  Yeah, the note at the top (which I added after I added this) is probably enough.

schmmd: It isn't particularly friendly ;-)

I think it made sense before, but not now that we're actually investing in deep_qa as a library.

Regardless, I'm sure we'll get plenty of feedback that we won't prioritize following up on (I get plenty for Open IE).

PR #391, opened 2017-06-12
Multi gpu generator compatibility

DeNeutoy:
Threading the batch size through `Trainer` and `TextTrainer` might not be the best way to do this, because it is probably a bit confusing as to why we aren't using the `self.batch_size` attribute. Aside from that, this appears to work smoothly.


deep_qa/training/models.py
```diff
@@ -71,6 +71,75 @@ def compile(self, params: Params):  # pylint: disable=arguments-differ
...
+        or list of scalars (if the model has multiple outputs
+        and/or metrics). The attribute `model.metrics_names` will give you
+        the display labels for the scalar outputs.
+        """
+        inputs, targets, sample_weights = self._standardize_user_data(x, y,
+                                                                      sample_weight=sample_weight,
+                                                                      class_weight=class_weight,
+                                                                      check_batch_axis=True)
+        if self.uses_learning_phase and not isinstance(K.learning_phase(), int):
+            ins_batch = inputs + targets + sample_weights + [1.]
```
matt-gardner: `ins_batch` means `inputs_batch`, right?  This would be a whole lot more readable as either `batch_inputs` or `inputs_batch` (I'd prefer the former).

DeNeutoy: Yep, this is a Keras hangover, i'll change it.

PR #392, opened 2017-06-12
Make VectorMatrixMerge handle more general cases

matt-gardner:
This is for @bhavanadalvi - this PR makes it possible to use `VectorMatrixMerge` to concatenate hand-computed features onto word embeddings, basically doing the expand and the concatenate in a single layer.


deep_qa/layers/vector_matrix_merge.py
```diff
@@ -67,17 +67,32 @@ def compute_output_shape(self, input_shapes):
...
-        # it.  We'll ignore the actual values in this.
-        vector_mask_template = K.sum(matrix_mask, axis=self.mask_concat_axis)
-        vector_mask = K.expand_dims(K.ones_like(vector_mask_template), axis=self.mask_concat_axis)
+        matrix_mask = mask[-1]
         result_mask = matrix_mask
-        for _ in range(num_vectors):
-            result_mask = K.concatenate([vector_mask, result_mask], axis=self.mask_concat_axis)
-        return K.cast(result_mask, 'bool')
+        if mask_concat_axis >= K.ndim(matrix_mask):
+            # This means we're concatenating along an axis that is lower than the mask.  E.g.,
```
DeNeutoy: You mean greater here right? (In the sense of further to the right in terms of number of dimensions of the tensor)

deep_qa/layers/vector_matrix_merge.py
```diff
@@ -67,17 +67,32 @@ def compute_output_shape(self, input_shapes):
...
+            # we're adding dimensions to a word embedding.  In this case, we take all of the masks,
+            # assume they have the same shape, and compute K.any() with them.
+            masks = [matrix_mask] + [m for m in mask[:-1] if m is not None]
+            shapes = set([K.int_shape(m) for m in masks])
+            assert len(shapes) == 1, "Can't compute mask with uneven shapes..."
```
DeNeutoy: Maybe add the shapes to the error message.

deep_qa/layers/vector_matrix_merge.py
matt-gardner: Hmm...  I wasn't sure what to say here, and when I said "lower" I was thinking "lower level", not "lower numerical value".  But yeah, it's not clear.  Should I just say "greater"?  Is it clear enough why it works this way in that case?

deep_qa/layers/vector_matrix_merge.py
matt-gardner: Will do.

deep_qa/layers/vector_matrix_merge.py
DeNeutoy: "This means we're concatenating along an axis in the tensor that is greater than the number of dimensions in the mask."

matt-gardner:
Something got messed up in github's connection to travis, I think because I merged master immediately after pushing the commit.  Travis is actually passing.  I'm going to merge this now.

PR #393, opened 2017-06-12
Refactor tokenizers

matt-gardner:
NOTE: This is a PR to the `new_data_api` branch, not to master.  This will not pass tests as it is, I am sure.  But to avoid having one massive PR once I've finished passing all of the tests, I'm going to submit pieces to the `new_data_api` branch as I go along, recognizing that tests will be broken as I'm pulling things apart and rebuilding them.

So, this PR is just to review the basic structure of how I've set this up.  So far I've just got the notion of `Fields`, where a `TextField` can be tokenized and represented as arrays in various ways.  `Instances` will be built up of multiple `Fields`, and `Datasets` will have multiple `Instances`.  Once all of that is done, I'll worry about embedding stuff for each of the token representations.


deep_qa/data/token_indexers/token_indexer.py
```diff
@@ -0,0 +1,77 @@
+from typing import Dict, List, Union
+
+from ...common.params import Params
+from ..vocabulary import Vocabulary
+
+
+TokenType = Union[int, List[int]]  # pylint: disable=invalid-name
```
DeNeutoy: I don't think this is complex or long enough to warrant it's own type.

deep_qa/data/vocabulary.py
```diff
@@ -0,0 +1,119 @@
...
+                    self.add_token_to_index(token, namespace)
+
+    def increment_token_count(self, token: str, namespace: str='tokens', count: int=1):
+        """
+        Increments the count of ``token`` in ``namespace`` by ``count``.  This is used for
+        determining which tokens are included in the vocabulary.
+        """
+        self._token_counts[namespace][token] += count
+
+    def add_token_to_index(self, token: str, namespace: str='tokens') -> int:
```
DeNeutoy: Can we call this and the method below `add_token_to_namespace` rather than `index`. Everyone will know what the index of a `Vocabulary` object does, but this is specifically adding a token to a particular namespace.

deep_qa/data/vocabulary.py
```diff
@@ -0,0 +1,119 @@
...
+            return self._token_indices[namespace].get(token, -1)
+        if token not in self._token_indices[namespace]:
+            index = len(self._token_indices[namespace])
+            self._token_indices[namespace][token] = index
+            self._reverse_token_indices[namespace][index] = token
+            return index
+        else:
+            return self._token_indices[namespace][token]
+
+    def tokens_in_index(self, namespace: str='tokens'):
```
DeNeutoy: `tokens_in_namespace`/ `all_tokens_in_namespace`

deep_qa/data/vocabulary.py
```diff
@@ -0,0 +1,119 @@
...
+    Vocabularies also allow for several different namespaces, so you can have separate indices for
+    'a' as a word, and 'a' as a character, for instance.  Most of the methods on this class allow
+    you to pass in a namespace; by default we use the 'tokens' namespace, and you can omit the
+    namespace argument everywhere and just use the default.
+    """
+    def __init__(self):
+        self._padding_token = "@@PADDING@@"
+        self._oov_token = "@@UNKOWN@@"
+        self._token_counts = defaultdict(lambda: defaultdict(int))
+        self._token_indices = defaultdict(lambda: {self._padding_token: 0, self._oov_token: 1})
```
DeNeutoy: can these be called `token_to_index` and `index_to_token`?

deep_qa/data/tokenizers/word_splitter.py
```diff
@@ -128,6 +139,7 @@ def split_words(self, sentence: str) -> List[str]:
 
 word_splitters = OrderedDict()  # pylint: disable=invalid-name
 word_splitters['simple'] = SimpleWordSplitter
+word_splitters['spaces'] = SpaceWordSplitter
 word_splitters['nltk'] = NltkWordSplitter
 word_splitters['spacy'] = SpacyWordSplitter
 word_splitters['no_op'] = NoOpWordSplitter
```
DeNeutoy: I think there might be a way we could remove this completely if we expose the `TokenisedTextField` as a `Field` you are allowed to specify in a `Dataset`, but this can happen after this mega re-write.

deep_qa/data/vocabulary.py
```diff
@@ -0,0 +1,119 @@
...
+            namespace_dict = instance.words()
+            for namespace in namespace_dict:
+                for token in namespace_dict[namespace]:
+                    namespace_token_counts[namespace][token] += 1
+        for namespace in tqdm.tqdm(namespace_token_counts):
+            for token, count in namespace_token_counts[namespace].items():
+                if count >= min_count:
+                    self.add_token_to_index(token, namespace)
+
+    def increment_token_count(self, token: str, namespace: str='tokens', count: int=1):
```
DeNeutoy: I feel strongly that this is not how we should cap the vocabulary size, by calling this method in the TokenIndexer. Instead, we should create a vocabulary per namespace by token frequency, using a sorted `Counter`, which then becomes the index for a given word. This way, we can just pass a maximum vocabulary size to `fit_token_dictionary` and remove the `count_vocab_items` methods from the `TokenIndexers`.

deep_qa/data/fields/text_field.py
```diff
@@ -0,0 +1,72 @@
...
+        self._token_arrays = token_arrays
+        self._token_indexers = token_indexers
+
+    @overrides
+    def get_padding_lengths(self) -> Dict[str, int]:
+        lengths = []
+        for indexer, array in zip(self._token_indexers, self._token_arrays):
+            indexer_lengths = {}
+            token_lengths = [indexer.get_padding_lengths(token) for token in array]
+            for key in token_lengths[0]:
```
DeNeutoy: Calling this `key` confused me because token_lengths isn't a dict.

deep_qa/data/fields/field.py
```diff
@@ -0,0 +1,117 @@
...
+        """
+        Given a set of specified padding lengths, actually pad the data in this field and return a
+        numpy array of the correct shape.  This actually returns a list instead of a single array,
+        in case there are several related arrays for this field (e.g., a ``TextField`` might have a
+        word array and a characters-per-word array).
+        """
+        raise NotImplementedError
+
+
+class TokenizedField:
```
DeNeutoy: Are users going to be able to specify a `TokenisedField` as one of the fields in dataset directly?

deep_qa/common/util.py
```diff
@@ -18,6 +18,41 @@ def group_by_count(iterable: List[Any], count: int, default_value: Any) -> List[
...
+        shorter ones are padded to it.
+
+    default_value: Callable, default=lambda: 0
+        Callable that outputs a default value (of any type) to use as padding values.
+
+    padding_on_right : bool, default=True
+        When we add padding tokens (or truncate the sequence), should we do it on the right or
+        the left?
+    """
+    if padding_on_right:
```
matt-gardner: Then this does nothing.  I tested that in an interpreter, and it will be tested by all of the instance tests (once I get those to actually pass).

DeNeutoy: Hmm sorry. I tested the `del x[20]`, without the colon and that crashes.

deep_qa/data/fields/field.py
matt-gardner: Maybe?  I debated about that.  It probably makes sense, but it also allows for easy errors.  E.g., if you're reading in an image, you can probably go straight to `IndexedField`, but then the dataset / instance class has to be sure that it's not calling `count_vocab_items` when it's iterating over the fields.

So, how about this:

The instances can take any of `RawField`, `TokenizedField`, or `IndexedField` as their input.  The first step will be to iterate over all of the fields, and if they're `Raw` (or maybe just `Untokenized`), we tokenize them.  Then we iterate over all of the fields again, and count vocab items on all of the `TokenizedFields`.  Then we iterate over all of them again, and convert `TokenizedFields` to `IndexedFields`.  It's a bunch of type checks on the fields, but it's not too bad, just checking the field's super class.

This way, when we have a `LabelField`, or a `TagField`, they start out as `TokenizedFields`, and have no `RawField` version - the `Dataset` that's constructing the instances doesn't have to think too hard about what level the field is getting constructed at, because there will only be one public class for each field type.  Does this make sense?

deep_qa/data/fields/field.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def index(self, vocab: Vocabulary) -> IndexedField:
+        """
+        Given a :class:`Vocabulary`, converts all tokens in this field into (typically) integer
+        arrays, returning an ``IndexedField`` as a result.
+        """
+        raise NotImplementedError
+
+
+class RawField:
```
matt-gardner: If we're going to allow `Fields` to be created at any level of this pipeline, I think we should stick with `Raw` here (or maybe `Untokenized`, or similar).

deep_qa/data/token_indexers/token_indexer.py
matt-gardner: I think it makes the code below more readable.  Instead of wondering why the return type or input type of a method below is `Union[int, List[int]]`, you just know that the method takes or returns a token.  It's also like this because if you add a new token representation later, (e.g., List[List[int]], or Tuple[int, List[int]], or whatever, for whatever reason), and you want the types to remain accurate, you just modify this type definition.

Basically, the types are only there for readability - we don't actually use a type checker anywhere.  I think `TokenType` aids readability, and the complex `Union[]` type actually harms readability if it's there.  Does that make sense?

deep_qa/data/vocabulary.py
matt-gardner: Yes.

deep_qa/data/vocabulary.py
matt-gardner: I'm not sure I understand what you're suggesting here - we should probably talk about this one in person.

deep_qa/data/vocabulary.py
matt-gardner: Sure.

deep_qa/data/vocabulary.py
matt-gardner: Ok.

deep_qa/data/tokenizers/word_splitter.py
matt-gardner: Yes, I agree with this.  See the comment above.

deep_qa/data/fields/text_field.py
matt-gardner: But it's a list of dicts, so `token_lengths[0]` is a dict.  Any suggestions for how to make this more clear?  Does it need a comment?

DeNeutoy: No sorry this is just me not knowing that if you iterate over a dictionary you just get the keys by default. I thought you always had to explicitly take the keys using `dict.keys()`. This is fine.

deep_qa/data/token_indexers/token_indexer.py
DeNeutoy: I would agree with you if it wasn't the union of two types, but you da boss

deep_qa/data/fields/field.py
matt-gardner: Or we could rename `TokenizedField` to be `UnindexedField`, and `RawField` to be `UntokenizedField`.  That would give us three field types, in this order: `UntokenizedField`, `UnindexedField`, `IndexedField`, with naming around what still needs to be done to convert them into arrays.

matt-gardner:
Ok, this is ready for another look, I think.

matt-peters:
Nice to see the model / embedding specific stuff removed from the tokenizers 👍  

Since I'm not familiar with some of the details of how the Instances work and this is the first time I've looked at most of this code, I'm having a hard time seeing how these pieces all fit together, to e.g. take a list of sentences and return some padded arrays.  Hopefully that will become apparent when tests are added.

Otherwise, I have two general comments:

(1) These classes seem to have a large public API footprint.  If these methods are necessary then keep them -- but otherwise it's better to remove them or make private sooner rather then later.

(2) I appreciate the use of the `Params` for specifying options, but I'd strongly consider moving the `Params` up a layer to the `Trainer` or elsewhere and using more traditional arguments for these lower level components.  For someone not familiar with this library it makes it difficult to determine how to construct these objects (and even adds an additional hurdle to just importing and using these classes, since it also requires understanding how `Params` works too).  It also means in most cases one needs to inspect the code to see what arguments are required and what the defaults are as it is essentially using a `def __init__(self, **kwargs):`  signature for the constructors.

matt-gardner:
On `Params` - that's how we're handling dependency injection, so you can just specify a list of parameters in a config file for easy experimentation.  An alternative would be to have each class that currently takes `Params` instead have a `@classmethod` called `from_params()`, that lets the experiment framework still construct the experiment, while not making the data code itself _require_ the use of the experiment framework (there's still a code dependency, through the `from_params()` method, but it's no longer required for using it).  Would that resolve your concerns with it?

matt-gardner:
Also, the part where the pieces fit together aren't part of this PR, so it's not surprising that you can't see it yet =).  I'm almost done with those.  There's an `Instance` class that contains a dictionary of named `Fields`, and a `Dataset` that batches `Instances` together.  The `Dataset` can read itself from a file (or list of files), constructing `Instances` with specific `Fields`.  This `Dataset` can then either be converted into one large set of arrays (keyed by field name), for `model.fit()`, or split up into batches (which are also just `Dataset` objects) by a data generator, for `model.fit_generator()`.

matt-peters:
RE: Params, a `from_params` method would resolve it.  I'm not sure how much additional overhead that adds.  I was just looking for a way to decouple the experiment framework from the data generator piece.  The code dependency is fine -- by using this library someone has already introduced that dependency into their stack.

deep_qa/data/vocabulary.py
```diff
@@ -0,0 +1,111 @@
...
+    namespace argument everywhere and just use the default.
+    """
+    def __init__(self):
+        self._padding_token = "@@PADDING@@"
+        self._oov_token = "@@UNKOWN@@"
+        self._token_to_index = defaultdict(lambda: {self._padding_token: 0, self._oov_token: 1})
+        self._index_to_token = defaultdict(lambda: {0: self._padding_token, 1: self._oov_token})
+        self._finalized = False
+
+    def set_from_file(self, filename: str, oov_token: str="@@UNKNOWN@@", namespace: str="tokens"):
```
matt-gardner: There are a couple of considerations:

1. I am planning on changing the `fit_word_dictionary` method to use a counter object, removing the `increment_count` method.  Then this could be a `@classmethod` like `from_dataset()`, like you're suggesting.
2. The `set_from_file` only exists to support reading in a vocabulary from someone else's trained model (I added this recently when I was getting your language model stuff into DeepQA).  You're right that I could similarly make this a `@classmethod`, and I will.
3. You actually might want the vocabulary to be mutable after construction in some cases.  Roy is looking at modifying a loaded model, changing the embedding layer and possibly adding words to it, which would require changing a loaded vocabulary.  So I'm not as convinced about removing the methods that allow modifying the vocabulary (there's only one, right?).

deep_qa/data/vocabulary.py
```diff
@@ -0,0 +1,111 @@
...
+        self._oov_token = oov_token
+        self._token_to_index[namespace] = {self._padding_token: 0}
+        self._index_to_token[namespace] = {0: self._padding_token}
+        with codecs.open(filename, 'r', 'utf-8') as input_file:
+            for i, line in enumerate(input_file.readlines()):
+                token = line[:-1]  # remove the newline
```
matt-gardner: Good catch; I normally do that too.  Not sure why I did it this way this time.

deep_qa/data/vocabulary.py
```diff
@@ -0,0 +1,111 @@
...
+        if token not in self._token_to_index[namespace]:
+            index = len(self._token_to_index[namespace])
+            self._token_to_index[namespace][token] = index
+            self._index_to_token[namespace][index] = token
+            return index
+        else:
+            return self._token_to_index[namespace][token]
+
+    def tokens_in_namespace(self, namespace: str='tokens'):
+        return self._token_to_index[namespace].keys()
```
matt-gardner: Hmm, looks like I need to switch `_index_to_token` to be a list, not a dictionary.

deep_qa/data/tokenizers/character_tokenizer.py
```diff
@@ -1,66 +1,39 @@
...
+        embedding will decide that they are at least similar - while the token for the accent is
+        indeed shared across different accented vowels.
+    lowercase_characters : ``bool``, optional (default=``False``)
+        If ``True``, we will lowercase all of the characters in the text before doing any other
+        operation.  You probably do not want to do this, as character vocabularies are generally
+        not very large to begin with, but it's an option if you really want it.
     """
     def __init__(self, params: Params):
+        self.byte_encoding = params.pop('byte_encoding', None)
+        self.lowercase_characters = params.pop('lowercase_characters', False)
```
matt-gardner: That also means we would need a way to specify that `from_params`, which is possible, but more work.  I vote for switching that one later, when there's a compelling use for it.

matt-peters: Got it -- agree, it's low priority and I wouldn't bother changing it now either.

deep_qa/data/vocabulary.py
matt-peters: Sounds good, and agree on the usefulness of adding some additional words to a loaded model.  I wasn't sure if you were anticipating supporting that or not.

matt-gardner: What I'm currently thinking is to have the vocabulary's constructor look like this:

```python
def __init__(self,                                   
             counter: Dict[str, Dict[str, int]]=None,
             non_padded_namespaces: List[str]=None,  
             min_count: int=1                        
             max_vocab_size: int=None):              
```

(The non-padded namespaces bit is so that we can use the same vocabulary object for handling tag sequences and other labels that need to be mapped to integers, but don't need padding or OOV token ids.)

This way, the vocabulary is given an initial set of counts, and constructs itself that way, and we can have a `from_dataset()` method that computes counts from the dataset, then passes those counts into the constructor.  Does this make sense?  Any better recommendations?

matt-gardner:
I think separating the experiment framework from the data code is a reasonable thing to do, and I've been conflicted about exactly how `Params` works ever since we started using it - any method for doing dependency injection has issues.  I think making a `from_params` method is a reasonable compromise, so you can just ignore the whole thing if you want to use some other method.

matt-gardner:
Ok, I think I address all of the outstanding issues with this - unless I get a more comments, I'm going to merge this tomorrow morning sometime, and submit a new PR with the instance and dataset classes.

PR #394, opened 2017-06-13
Split repo

DeNeutoy:


DeNeutoy:
@matt-gardner You may find it easier to review this by looking actually at my branch and seeing what is left and if there is anything else you want removed, rather than looking at all this deleted stuff 🗑

PR #398, opened 2017-06-15
 add tensorflow to setup as keras doesn't install it

DeNeutoy:


matt-gardner:
There should maybe be a note somewhere, though, about installing `tensorflow-gpu` if you want to use the GPU.  That's probably why keras doesn't install it.

PR #399, opened 2017-06-15
Delete RELEASE.md

schmmd:
I propose removing this file and instead using GitHub's build-in support.

https://github.com/allenai/deep_qa/releases


matt-gardner:
Maybe this was added because of pip?  @DeNeutoy might have more info than me here.

DeNeutoy:
I added this so we can have release information in the docs, up to you whether we delete it.

allenai-dev-role:
I'd still just link to the release pages in GitHub--it's pretty, informative, and useful.

PR #401, opened 2017-06-16
split run_model into two functions

DeNeutoy:
This is the easiest way to be able to access the `run_model` functionality from outside deep_qa at the moment. This way, we can load a json file seperately, get the model class from a `aristo-nlp-models` `concrete_classes` dict and then call `run_model`.


deep_qa/run.py
```diff
@@ -83,7 +96,9 @@ def run_model(param_path: str, model_class=None):
         handler.setLevel(logging.INFO)
         handler.setFormatter(logging.Formatter('%(asctime)s - %(levelname)s - %(name)s - %(message)s'))
         logging.getLogger().addHandler(handler)
-        shutil.copyfile(param_path, log_dir + "_model_params.json")
+        serialisation_params = deepcopy(params).as_dict()
```
matt-gardner: This is going to log all of the parameters, won't it?  Maybe adding a parameter to `as_dict()` that tells it to be quiet would be a good idea.

PR #402, opened 2017-06-16
Refactor instances and datasets

matt-gardner:
I redid the way `Fields` work a bit, making it simpler and easier to make one field depend on another.  I also moved over the `DatasetReaders` to return instances using this new API, and removed all of the old `Instance` classes.  This is a mostly final implementation of how you go from text files to numpy arrays.  I _think_ all of the pieces for that part are in place.  I'll update the tests as soon as you think this design is reasonable.

Actually committing this to master has to wait until we've implemented the embedding stuff, though, and changed how experiments run...  It'll be a while still before we can actually switch over.


deep_qa/data/vocabulary.py
```diff
@@ -7,35 +7,38 @@
...
+    def __missing__(self, key: str):
+        value = None
+        for namespace_str in self._non_padded_namespaces:
+            if namespace_str[0] == '*' and key.endswith(namespace_str[1:]):
+                value = self._non_padded_function()
+            elif namespace_str == key:
+                value = self._non_padded_function()
+        if value is None:
+            value = self._padded_function
+        dict.__setitem__(self, key, value)
```
DeNeutoy: Is there an important difference between this and `self[key] = value`?

deep_qa/data/vocabulary.py
```diff
@@ -7,35 +7,38 @@
 
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
+class _NamespaceDependentDefaultDict(defaultdict):
```
DeNeutoy: A short comment here explaining this would have helped me: "A defaultdict which initialises different default dictionaries depending on whether or not the name_space requires default padding and unknown tokens."

deep_qa/data/token_indexers/single_id_token_indexer.py
```diff
@@ -52,8 +52,8 @@ def get_padding_lengths(self, token: TokenType) -> Dict[str, int]:
     def pad_token_sequence(self,
                            tokens: List[TokenType],
                            desired_num_tokens: int,
-                           padding_lengths: Dict[str, int]):
-        pad_sequence_to_length(tokens, desired_num_tokens, default_value=0)
+                           padding_lengths: Dict[str, int]) -> List[TokenType]:
+        return pad_sequence_to_length(tokens, desired_num_tokens, default_value=0)
```
DeNeutoy: `default_value=lambda: 0`

deep_qa/data/token_indexers/token_characters_indexer.py
```diff
@@ -59,22 +59,22 @@ def get_padding_token(self) -> TokenType:
     def pad_token_sequence(self,
                            tokens: List[TokenType],
                            desired_num_tokens: int,
-                           padding_lengths: Dict[str, int]):
-        pad_sequence_to_length(tokens, desired_num_tokens, default_value=0)
+                           padding_lengths: Dict[str, int]) -> List[TokenType]:
+        padded_tokens = pad_sequence_to_length(tokens, desired_num_tokens, default_value=0)
```
DeNeutoy: `default_value=lambda:0`

deep_qa/common/util.py
```diff
@@ -41,16 +41,21 @@ def pad_sequence_to_length(sequence: List,
...
         When we add padding tokens (or truncate the sequence), should we do it on the right or
         the left?
+
+    Returns
+    -------
+    padded_sequence : List
     """
     if padding_on_right:
-        del sequence[desired_length:]
+        padded_sequence = sequence[:desired_length]
```
matt-gardner: I think it's cleaner to not modify the input list - that way we don't get subtle bugs like we had with the decomposable attention padding issue that you ran into.  I switch this to return a new list instead of modifying the input list.

deep_qa/data/__init__.py
```diff
@@ -1,5 +1,5 @@
-from .datasets.dataset import Dataset, IndexedDataset, TextDataset
-
+from .datasets.dataset import Dataset
```
matt-gardner: Thanks.  Though at this point I'm mostly just looking for API-level feedback - I can find all of these issues with tests and pylint later, I don't want to waste your time doing this manually.

deep_qa/data/dataset.py
```diff
@@ -0,0 +1,157 @@
...
+    def __init__(self, instances: List[Instance]):
+        """
+        A Dataset just takes a list of instances in its constructor.  It's important that all
+        subclasses have an identical constructor to this (though possibly with different Instance
+        types).  If you change the constructor, you also have to override all methods in this base
+        class that call the constructor, such as `merge()` and `truncate()`.
+        """
+        self.instances = instances
+
+    def truncate(self, max_instances: int):
```
matt-gardner: Yeah, that's a fair point.  It's my bias towards making things immutable, but `Dataset` is already very much not immutable, so I might as well make this one consistent.  I'll change it.

deep_qa/data/datasets/__init__.py
```diff
@@ -1,7 +1,7 @@
 from collections import OrderedDict
 
-from .entailment.snli_dataset import SnliDataset
-from .language_modeling.language_modeling_dataset import LanguageModelingDataset
+from .snli_dataset import SnliDataset
+from .language_modeling_dataset import LanguageModelingDataset
```
matt-gardner: Oh, they're the readers.  There's no additional class - we only need one `Dataset` class, and we just have `DatasetReaders` to construct datasets with particular instances.

This was from an earlier iteration of the design, before I settled on what you see in this PR.  I'll fix this.

deep_qa/data/dataset_readers/language_modeling.py
```diff
@@ -0,0 +1,111 @@
...
+        instances = []
+        for tokenized_string in tokenized_string:
+            input_field = TokenizedTextField(tokenized_string[:-1], self._token_indexers)
+            output_field = TokenizedTextField(tokenized_string[1:], [output_indexer])
+            instances.append(Instance({'input_tokens': input_field,
+                                       'output_tokens': output_field}))
+        return Dataset(instances)
+
+    @classmethod
+    def from_params(cls, params: Params):
```
matt-gardner: You should avoid having heavy constructors in general.  When we get to the experiment framework, it will construct all of the objects it needs from a parameter file.  If those constructors take a long time to load, that means you will be waiting for them before you get to validation of the parameters, for example.

I think the difference between what you're thinking and what I have here is this:

Yours:
```python
dataset_params = params.pop('dataset')
dataset = Dataset.from_params(dataset_params)  # expensive operation
```

Mine:
```python
dataset_reader_params = params.pop('dataset')
dataset_reader = DatasetReader.from_params(dataset_reader_params)
dataset = dataset_reader.read()  # expensive operation
```

That's the only real difference in usage, I think.  Some reasons I prefer mine:

- You can delay the expensive operation, so that, e.g., parameter validation can happen without waiting for an expensive operation.
- You don't have to subclass `Dataset`.  The notions of reading stuff from disk and actually manipulating in-memory `Instances` are totally separate, and this keeps them separate in the code.  Each class does one thing: the `DatasetReader` reads stuff from disk, and the `Dataset` deals with manipulating `Instances`.

deep_qa/data/fields/field.py
```diff
@@ -1,43 +1,62 @@
...
+
+
+    Fields go through up to two steps of processing: (1) tokenized fields are converted into token ids,
+    (2) fields containing token ids (or any other numeric data) are padded (if necessary) and converted
+    into data arrays.  The ``Field`` object has methods to say which state the field is in, and to get
+    it from one state to the other.  If ``Field.needs_indexing()``, we will compute a vocabulary and
+    pass it to the field to use for indexing.  Once all fields are indexed, we will determine padding
+    lengths, then intelligently batch together instances and pad them into actual arrays.
     """
+    def needs_indexing(self):
```
matt-gardner: I do think it's reasonable for a user to implement a new `Field`, it just won't be typical, because what we have should cover most NLP uses.

I'm not sure what you're suggesting here.  You just have implementations for `count_vocab_items()` and `index()` that pass in `IndexField`?  Yeah, I could probably be convinced to do it that way.

Here are three alternatives:

1. As shown in this PR, we have a `needs_indexing` method, and `Fields` that do not need any indexing don't have to implement indexing-specific methods at all.

2. As you suggest, we remove `needs_indexing`, and just require all `Fields` to implement indexing-specific methods with a `pass`, or similar.  Because there are only two of them, this isn't that big a deal.

3. Like `SequenceField`, we introduce a new abstract `Field` type that adds the indexing-specific methods (e.g., `StringField`), and leave them out of the `Field` base class.  Then things like `IndexField` don't need to implement the indexing-specific methods, just the `Fields` that have strings, and we use an `isinstance(field, StringField)` check instead of `needs_indexing()` when iterating over `Fields` in `Instance`.  The trouble with this is it makes something like `ListField` more complicated - should it be a `StringField`?  That depends on what's inside it, really.  Also, we'd need to use multiple inheritance in order to have `TextField` based on both `SequenceField` and `StringField` (not to mention the potential for naming confusion, with too many abstract base types).

I don't have a really strong opinion between (1) and (2), though (3) seems like not a very good idea.  (1) makes the `Field` implementations slightly cleaner, with less boilerplate for stuff unrelated to a non-string field.  (2) makes the calling code cleaner, without having to check if a `Field` needs indexing.  We could even do (4) remove `needs_indexing`, and change the default `RuntimeErrors` raised in `Field` to `pass`, so that non-string fields don't need to implement anything, and the calling code is still clean...  Maybe that's the right thing to do.

What do you think?

deep_qa/data/fields/field.py
```diff
@@ -1,43 +1,62 @@
...
+        """
+        If there are strings in this ``Field`` that need to be indexed, you must return ``True``
+        here; else return ``False``.
+        """
+        raise NotImplementedError
+
+    # pylint: disable=no-self-use,unused-argument
+    def count_vocab_items(self, counter: Dict[str, Dict[str, int]]):
```
matt-gardner: Good point.

deep_qa/data/dataset_readers/language_modeling.py
DeNeutoy: Ok fine on the not doing the expensive thing in the init call, but still why doesn't the signature look like this:
```
dataset_reader_params = params.pop('dataset')
dataset_type = concrete_dataset_readers[dataset_reader_params.pop("type")]

dataset_reader = dataset_type(dataset_reader_params)
dataset = dataset_reader.read()  # expensive operation
```
The classmethod on each DatasetReader is different so we still need to know which dataset_reader we are using beforehand - it seems like having this parameter unpacking inside the `DatasetReader.__init__` is effectively equivalent?

deep_qa/data/vocabulary.py
matt-gardner: I don't know.  The examples that I saw online suggested this way.  I'll see if I can figure out a better reason.

deep_qa/data/vocabulary.py
```diff
@@ -7,35 +7,38 @@
...
+                value = self._non_padded_function()
+            elif namespace_str == key:
+                value = self._non_padded_function()
+        if value is None:
+            value = self._padded_function
+        dict.__setitem__(self, key, value)
+
+
 
 class _TokenToIndexDefaultDict(defaultdict):
```
matt-gardner: Oops, yes, that was the intent.

deep_qa/data/fields/field.py
DeNeutoy: Yeah I was suggesting option 2 - we could even have the `Field` implement these methods as `pass` so that only fields which need to do something have to implement them. Let's see what @matt-peters thinks.

deep_qa/data/dataset_readers/language_modeling.py
matt-gardner: There will be a `from_params` method on the abstract base class, like for [`TokenIndexer`](https://github.com/allenai/deep_qa/pull/402/files#diff-ea26f15c9581a597a609b713eaf11965R79) and [`Tokenizer`](https://github.com/allenai/deep_qa/pull/402/files#diff-ca9750b4d8885ba66736abff077b9722R32), that does what you put in your code block - I just forgot to implement it for `DatasetReader`.  So the code that constructs a `DatasetReader` from parameters actually does just look like the block I showed:

```python
dataset_reader_params = params.pop('dataset')
dataset_reader = DatasetReader.from_params(dataset_reader_params)
```

The difference between separating out the use of `Params` from the constructor is two-fold:

1. It makes it more obvious what the actual objects are that you're taking as a dependency.  When you're looking at the constructor, you can see the right object types for, e.g., the tokenizer, instead of just getting a string description of the tokenizer.  This also helps someone new to the codebase who is familiar with python get started easier.

2. It separates the experiment framework code (`Params`) from the data generator code.  If you want to totally ignore our experiment framework, but still use our data processing and other code, this makes it a lot easier.

deep_qa/data/fields/field.py
matt-gardner: Yeah, the "we could even" that you said is my option (4), and that's what I'm leaning towards.

DeNeutoy: Sorry I didn't read that far 💡

deep_qa/data/dataset_readers/language_modeling.py
DeNeutoy: Ok fair enough, sounds good.

deep_qa/data/vocabulary.py
DeNeutoy: No worries, I was just wondering. I was generally under the impression that avoiding calling magic methods directly was good practice, but maybe that's wrong when we get a bit deeper into Python's internals.

matt-gardner: Yes, it is definitely good practice to avoid these.  I think I know the reason to avoid it here, though - you'll run into a recursive loop.  `self[key]` calls `__missing__` if `key not in self`.

deep_qa/data/instance.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def pad(self, padding_lengths: Dict[str, Dict[str, int]]) -> Dict[str, List[numpy.array]]:
+        """
+        Pads each ``Field`` in this instance to the lengths given in ``padding_lengths`` (which is
+        keyed by field name, then by padding key, the same as the return value in
+        :func:`get_padding_lengths`), returning a list of numpy arrays for each field.
+        """
+        arrays = {}
+        for key, field in self._fields.items():
+            field_lengths = padding_lengths.get(key, {})
```
matt-peters: In what case would `padding_lengths` not contain the key?  It's always set in `get_padding_lengths`.  If not possible better to fail with an exception here then silently.

deep_qa/data/instance.py
```diff
@@ -0,0 +1,62 @@
...
+        for _, field in self._fields.items():
+            if field.needs_indexing():
+                field.count_vocab_items(counter)
+
+    def index_fields(self, vocab: Vocabulary):
+        """
+        Converts all ``UnindexedFields`` in this ``Instance`` to ``IndexedFields``, given the
+        ``Vocabulary``.  This `mutates` the current object, it does not return a new ``Instance``.
+        """
+        for _, field in self._fields.items():
```
matt-peters: FYI, if you don't need the key you can just do `for field in self._fields.values():`

deep_qa/data/fields/field.py
```diff
@@ -1,43 +1,53 @@
...
+        """
+        If there are strings in this field that need to be converted into integers through a
+        :class:`Vocabulary`, here is where we count them, to determine which tokens are in or out
+        of the vocabulary.
+
+        If your ``Field`` does not have any strings that need to be converted into indices, you do
+        not need to implement this method.
+        """
+        pass
```
matt-peters: Will you run into problems if different Fields decide to use the same namespace / key of the `counter` dict?

deep_qa/data/fields/field.py
```diff
@@ -1,43 +1,53 @@
...
+    ids, (2) fields containing token ids (or any other numeric data) are padded (if necessary) and
+    converted into data arrays.  The ``Field`` API has methods around both of these steps,
+    though they may not be needed for some concrete ``Field`` classes - if your field doesn't have
+    any strings that need indexing, you don't need to implement ``count_vocab_items`` or ``index``.
+    These methods ``pass`` by default.
+
+    Once a vocabulary is computed and all fields are indexed, we will determine padding lengths,
+    then intelligently batch together instances and pad them into actual arrays.
     """
+    def count_vocab_items(self, counter: Dict[str, Dict[str, int]]):
```
matt-peters: Is this method always used upstream?  Or is there a hook if someone already has a pre-computed vocabulary then this is bypassed?

deep_qa/data/fields/index_field.py
```diff
@@ -0,0 +1,34 @@
...
+
+    @overrides
+    def pad(self, padding_lengths: Dict[str, int]) -> List[numpy.array]:
+        one_hot_index = numpy.zeros(padding_lengths['num_options'])
+        one_hot_index[self._index] = 1
+        return one_hot_index
+
+    @overrides
+    def empty_field(self):
+        return IndexField(0, None)
```
matt-peters: This seems odd that in this case you are returning an instance of this class with only partially working methods (e.g. `get_padding_lengths` won't work).  There must be a different way to deal with the `empty_field` that doesn't require this.

deep_qa/data/instance.py
```diff
@@ -0,0 +1,62 @@
...
+            if field.needs_indexing():
```
matt-peters: I don't see `needs_indexing` in the Field class?

matt-peters: An alternate is to just remove `needs_indexing` from the API and make `index` a no-op for those that don't need indexing.

deep_qa/data/fields/field.py
matt-peters: Will separate fields get different vocabularies via this mechanism?  I think you want many fields to share a single vocabulary, e.g. both sentences in SNLI should have the some token ids.

matt-peters: Regardless, I don't think it's desirable to set the vocabulary in this manner.  One use case is pre-trained word vectors.  You typically want to include all tokens in with pre-trained vectors in your task vocabulary, since even if a token isn't in the training data set but is in the test set you'll be able to take advantage of the pre-trained vector.

deep_qa/data/fields/label_field.py
```diff
@@ -0,0 +1,55 @@
+from typing import Dict, List, Union
+
+from overrides import overrides
+import numpy
```
matt-gardner: I prefer to avoid abbreviations wherever possible, so that someone doesn't have to know my import conventions in order to read my code.  The first time I saw `pd` in a piece of code I had no idea what I was looking at, and it still hurts readability, just to save 4 characters on a line that's already short, typically.

Yes, there are some conventions that are widespread enough that it's ok, and maybe `numpy as np` is one of them, but it's just three characters - I prefer to use the full name.

deep_qa/data/fields/label_field.py
```diff
@@ -0,0 +1,55 @@
...
+    def count_vocab_items(self, counter: Dict[str, Dict[str, int]]):
+        counter[self._label_namespace][self._label] += 1
+
+    @overrides
+    def index(self, vocab: Vocabulary):
+        self._label_id = vocab.get_token_index(self._label, self._label_namespace)
+
+    @overrides
+    def get_padding_lengths(self) -> Dict[str, int]:
+        return {'num_labels': self._label_id + 1}
```
matt-gardner: Oops, you're right.  I was originally going to use the vocab to get the number of labels (as the previous version did), but thought I could get away without it.  You're right, you can't.  I'll fix this.

deep_qa/data/fields/label_field.py
```diff
@@ -0,0 +1,55 @@
...
+
+    @overrides
+    def pad(self, padding_lengths: Dict[str, int]) -> List[numpy.array]:
+        label_array = numpy.zeros(padding_lengths['num_labels'])
+        label_array[self._label_id] = 1
+        return [label_array]
+
+    @overrides
+    def empty_field(self):
+        return LabelField(0, 'labels', False)
```
matt-gardner: It's never read, so it doesn't matter, but yeah, that'd be more consistent.

deep_qa/data/instance.py
matt-gardner: Thanks, fixed.

deep_qa/data/instance.py
matt-gardner: Yeah, you're right.  I think Pradeep ran into an issue where he needed to do this in a corner case in the old code, but it shouldn't be an issue here.

deep_qa/data/fields/field.py
matt-gardner: Different fields may or may not use the same namespace.  Text fields will by default all use the same namespace (so both sentences get the same token ids), but you could change that if you want, e.g., to give different ids for characters in a structured name field, if you really wanted to.

We also use this same mechanism to map tags and labels to integers (so, e.g., you don't need to do your own mapping from "B-PERS" to integers), and those will use a different namespace by default.

Regarding pre-trained vectors, if I have a word in my training data that appears often, but is not in my pre-trained vectors, I still want to assign it a non-OOV id, and learn a vector for it.  Or if I'm doing Turkish NLP, or using byte-level character representations, I might not have pre-trained vectors at all.  We definitely want to be able to count vocab items that occur in the data, and be able to map frequently occurring items in the data to unique ids.

I agree that we _also_ might want to give ids to any token I have a pre-trained vector for.  That's not incompatible with this - we're just computing counts here, not constructing the vocabulary, and the code in `Trainer` that constructs the vocabulary could add all tokens to the vocab that have a pre-trained vector.  That's not hard.

deep_qa/data/fields/field.py
matt-gardner: See comment above.  `TextFields` _should_ be using the same namespace.  If you have two different kinds of `LabelFields`, with mutually exclusive labels, you want to be sure you use different namespaces for them, and this is noted in the docstring for `LabelField`.

deep_qa/data/fields/index_field.py
matt-gardner: I agree that it's not ideal.  But we need to get the shapes right, and `TextField` could have arbitrary shape.  Any ideas on how to deal with this better?

This is an issue in, e.g., answer sentence selection, or memory networks, where each instance has a non-fixed number of sentences.  We need to know whether the sentence we're padding is being represented as word ids, character ids, two different word ids, or some other combination.

deep_qa/data/instance.py
matt-gardner: Yeah, I removed it, but forgot to update this class.  Thanks for catching it.

matt-gardner:
@DeNeutoy, on your top-level comment - isn't the whole point of importing things in the module `__init__.py` so that we can then import them from that module?  So you avoid things like `from data.vocabulary import Vocabulary`, and can just do `from data import Vocabulary`?  I'm not sure I understand your concern here.

Other than that, I think I've addressed the outstanding issues, and if you're happy with it, feel free to merge and work on the next stuff.

DeNeutoy:
That should be the objective for outside users of the library - however, within the library, it causes problems, because it makes the order of things which are imported in `__init__` important. It also makes it easy for circular imports to happen (such as here, where you import the `token_indexers` into the abstract class from init, but within init you import the abstract class). At least we should try not to do this within modules.

PR #407, opened 2017-06-20
Tests and fixes to new data api

DeNeutoy:


deep_qa/data/token_indexers/token_indexer.py
```diff
@@ -75,7 +74,6 @@ def pad_token_sequence(self,
         """
         raise NotImplementedError
 
-    @staticmethod
-    def from_params(params: Params):
-        choice = params.pop_choice('type', list(token_indexers.keys()), default_to_first_choice=True)
-        return token_indexers[choice].from_params(params)
```
matt-gardner: Where are you putting this code? This was intentional, actually, and makes constructing objects from params a lot easier. If it's not here, it has to be duplicated every time you want to construct an object.

deep_qa/data/token_indexers/single_id_token_indexer.py
```diff
@@ -2,10 +2,10 @@
 
 from overrides import overrides
 
-from ...common.util import pad_sequence_to_length
-from ...common import Params
-from ..vocabulary import Vocabulary
-from .token_indexer import TokenIndexer, TokenType
+from deep_qa.common.util import pad_sequence_to_length
+from deep_qa.common import Params
```
matt-gardner: These should be relative, right?

deep_qa/data/token_indexers/token_indexer.py
DeNeutoy: We can't have this here because it creates circular imports - `token_indexers` are imported from `__init__` but haven't been defined yet. We also can't import the classes directly because they inherit from this class.

deep_qa/data/token_indexers/single_id_token_indexer.py
DeNeutoy: thanks

deep_qa/data/token_indexers/token_indexer.py
DeNeutoy: One idea to fix this would be to have parameter checking only at the Model level. This would remove the need for the `from_params` on everything and we could just log parameters inside classes. This would have the benefit that the log would show where the parameter was defined, which it currently doesn't because all the logging happens in params. The replicability is nice, but we now achieve that with Dockerised experiments, so maybe it is less useful.

matt-gardner: Just move the import inside the method to remove the circular import.  We really do want everything to be _constructed_ `from_params`, so that we can have a single configuration file that runs a whole experiment.  Having this here makes it so that things like [this](https://github.com/allenai/deep_qa/blob/86d280c7eab491aedbc7fb38a4e1311ee4c31c1d/deep_qa/data/dataset_readers/snli.py#L69) work.  Otherwise it's more complicated to do the construction.

DeNeutoy: ok, will do

DeNeutoy:
In summary this PR does the following few things:

- Moves `sort_by_padding` from the `Dataset` to the `DataGenerator`
- Adds tests for : `Vocabulary`, `Dataset`, all concrete `Fields`, `Embeddings`, `TokenIndexers`, `DatasetReaders`(These could be more comprehensive).
- Roughly fixes the `DataGenerator` to work with the new `Dataset` class; I haven't fixed the tests for this, because MattG wrote them and they rely on some `FakeInstances`, which I'm not exactly sure how to translate to the new api.
- Adds getter methods to the `Field`s, `Instance` and `Dataset` classes, called things like `fields()` for the `Instance` class, and `tokens()` for the `TextField` etc. I think it's important to be able to interact with the different classes so that writing functions for mapping predictions back to text etc is easy.

matt-peters:
This looks great to me, from my somewhat quick look 👍 .  I didn't see any big API changes in this, just testing and cleanup, right?  It's nice to have usage examples in the test suite, thank you for adding them!

